PR #42, opened 2020-10-01
Added Chapter 1 Solutions

premkiran7:
### Solutions for Chapter 1

Felt happy after completing these tasks. The writeup was well written and interactive. 

Made some extra commits to handle edge cases involving -ve numbers for some tasks as mentioned below. The -ve number case was not obvious, so felt that it could be explicitly mentioned in the task info
```
firstDigit (-30)
sumLast2 (-10)
```
Apart from this, faced difficulties running the test-suite locally. Have mentioned it below,
```
Failed to build base-compat-0.11.2.
```

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -88,7 +88,7 @@ expressions unless really necessary. You can see the evaluation results either
 by running a Haskell program or by playing with some functions in the
 interactive interpreter (explained later).
 
-Haskell is a __statically-typed__ language, which means that each expression has
+Haskell is a __strongly-typed__ language, which means that each expression has
 a type. Each value and function is associated with some type. You can't change
 the value type. You can only pass a value to some function that will do its
 work, and maybe produce a value of a different type.
@@ -99,6 +99,11 @@ through the variables – begin with the lowercase letter. The concept of
 polymorphism is more sophisticated than working with concrete types, thus we
 won't dive too much into it in this chapter and will work with the concrete
 types for now.
+
+Furthermore, Haskell is a __statically-typed__ language, which means that each
+expression has the type known at compile-time, rather than run-time. It allows
+the compiler to catch some kinds of bugs in your program early; before you
+even run it.
 
 Additionally to static typing, Haskell has __type inference__. This means that
 you _don't need_ to specify the type of each expression as it is going to be
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -640,6 +640,11 @@ firstDigit n = error "firstDigit: Not implemented!"
 
 
 {-
+You did it! Now it is time to the open pull request with your changes
+and summon @vrom911 and @chshersh for the review!
+}
+
+{-
 =📜= Additional resources
 
 Modules: http://learnyouahaskell.com/modules
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -204,31 +204,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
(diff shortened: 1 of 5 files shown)


src/Chapter1.hs
```diff
@@ -614,9 +623,11 @@ Implement a function that returns the sum of the last two digits of a number.
...
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n = lastDigit (abs n)  + lastDigit (div (abs n) 10)
+  where
+    lastDigit :: Int -> Int
+    lastDigit n = let remainder = mod n 10 in remainder
```
vrom911, line 630: Great one 👌🏼 

Some additional info to this one: in the standard library, there is the `divMod` function, that is doing what the `lastDigit` here 🙂

src/Chapter1.hs
```diff
@@ -635,8 +646,8 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n = if div (abs n) 10 == 0 then n else firstDigit (div (abs n) 10)
```
vrom911, line 650: 💪🏼 Nice one!

A few things in here to pay attention to. You can see that `div (abs n) 10` is used twice in this function. In Haskell, the function with the same input will always produce the same result. That means, that these two expressions are always the same. So you can move in into the separate variable through the `let-in` or `where` constructions 🙂

vrom911: Also, I can see that there is a small nuance for numbers from -10 to 0 (negative one-digits). We apply the `div` function to the absolute value (which is _absolute_ correct), however, when we return the value, we return `n` itself. That would include minus as well. So this could be improved a bit 🙂

vrom911 (approved):
Amazing job 💯  Congrats with the successful Chapter 1 finish 🥳 
And great one with the advanced task 💪🏼

vrom911:
Regarding the local testing. Let's try to find out why it is happening 🤔 
Could you please provide a bit more information: more error text, the versions of GHC and Cabal and your operation system.

You can run

```shell
ghc --version
cabal --version
```

To get the GHC and Cabal versions.

premkiran7:
> Regarding the local testing. Let's try to find out why it is happening 
> Could you please provide a bit more information: more error text, the versions of GHC and Cabal and your operation system.
> 
> You can run
> 
> ```shell
> ghc --version
> cabal --version
> ```
> 
> To get the GHC and Cabal versions.

**OS -> Pop!_OS 20.04 LTS**
```
ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.10.2
```
```
cabal --version
cabal-install version 3.2.0.0
compiled using version 3.2.0.0 of the Cabal library 
```
**Error Text**

Have added a small snippet, since the error log is huge
```
collect2: error: ld returned 1 exit status
`gcc' failed in phase `Linker'. (Exit code: 1)
)
Failed to build exe:hsc2hs from hsc2hs-0.68.7 (which is required by
test:doctest-chapter1 from learn4haskell-0.0.0.0). See the build log above for
details.
Failed to build syb-0.7.1 (which is required by test:doctest-chapter1 from
learn4haskell-0.0.0.0). See the build log above for details.

make: *** [Makefile:11: test-chapter1] Error 1
```

vrom911:
Thanks for providing this information!
I have a suspicion that installing `libtinfo-dev` could solve the problem.

I am not sure about your OS, but for Ubuntu, you would need to:

```shell
sudo apt install libtinfo-dev
```

Let me know if it solves the issue.

You can also make

```shell
make clean
make build
```

To clean up cache and build again.

premkiran7:
> Thanks for providing this information!
> I have a suspicion that installing `libtinfo-dev` could solve the problem

Tried the solution, but the error still persists.

vrom911:
Hmm, interesting. Then it looks like a Cabal bug.
The best way would be to ask in their issue tracker, maybe they could advice what can we do with this one: https://github.com/haskell/cabal/issues
Unfortunately, I am not the Pop!_OS user, but I will try to see if I can find any solutions somewhere. Sorry that you experience this 😞

chshersh:
@premkiran7 We've been researching and discussing together with @vrom911 possible solutions to this problem. A few more things you can try to do:

* Copy commands from Makefile and add the `--enable-executable-static` flag
* Add GHC options to build with the `gold` linker to the `learn4haskell.cabal` file in the `common-options` stanza (below other `ghc-options`. You need to add the following lines with the proper indentation:
    ```haskell
      ghc-options: -optl-fuse-ld=gold
      ld-options:  -fuse-ld=gold
    ```
    (source: https://stackoverflow.com/questions/43243322/how-to-link-with-the-gnu-gold-linker-instead-of-ld-in-haskell/43243323#43243323)
* Download a different version of the `cabal-install` tool — a statically linked Cabal binary, and use it instead of your current one
    * https://downloads.haskell.org/~cabal/cabal-install-3.2.0.0/cabal-install-3.2.0.0-x86_64-alpine-linux-musl.tar.xz

Of course, you can always open a bug report directly to Cabal straightaway. These are just possible solutions.

Again, we are so sorry that you're experiencing such problems :disappointed: I wish it was easy and smooth for everyone to install Haskell and start using it without any problems!

premkiran7:
@vrom911 @chshersh I think I should create a separate branch and push chapter 2 solutions to it and then create a PR. Is that the expected way?

chshersh:
@premkiran7 Yes, you are correct. We've approved this PR and added a label, so it should count towards Hacktoberfest :shirt: :deciduous_tree: 
Ideally, you should open PRs for your fork. This approach gives you more flexibility, and you will be able to merge your PRs and have all solutions in a single place and see that CI is passing for each chapter. So, I'm going to close this PR as no further action is required in her, but feel free to open new PRs :slightly_smiling_face:

PR #45, opened 2020-10-01
Changed the relative path in the doctest files

poormate:
I changed the relative path in the doctest file for each chapter, because they were not pointing to the actual chapter files.

### DIFF
--- a/test/DoctestChapter1.hs
+++ b/test/DoctestChapter1.hs
@@ -5,5 +5,5 @@ import Test.DocTest (doctest)
 
 main :: IO ()
 main = doctest
-    [ "src/Chapter1.hs"
+    [ "../src/Chapter1.hs"
     ]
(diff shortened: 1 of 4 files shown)


chshersh:
@poormate Currently, with this patch, the CI for doctests is failing due to the following reason:

```
Running 1 test suites...
Test suite doctest-chapter1: RUNNING...

<no location info>: error: can't find file: ../src/Chapter1.hs
Test suite doctest-chapter1: FAIL
```

I guess you have errors running doctest tests locally, because you run them when you are inside the `src/` directory. Can you try running doctests from the root of this project and see whether they still fail? They work for me locally with `make` commands, and they also work on CI. So maybe a different solution is required rather than making paths relatives.

poormate:
@chshersh I tried it and they work from the root, sorry for bothering you.

chshersh:
@poormate No worries at all! :slightly_smiling_face:

PR #46, opened 2020-10-01
quickCheck test for two tasks in chapter one

dariodsa:
Implemented QuickCheck test for tasks ``lastDigit`` and ``firstDigit`` from Chapter 1.

cc @vrom911 @chshersh


test/Test/Chapter1QC.hs
```diff
@@ -0,0 +1,13 @@
...
+module Test.Chapter1QC
+    ( firstDigitRevLastDigit
+    ) where
+
+import Chapter1
+
+firstDigitRevLastDigit :: Int -> Bool
+firstDigitRevLastDigit x = firstDig == lastDig
+    where
+      lastDig = lastDigit (read  (reverse $ show $ abs x) :: Int)
```
vrom911, line 12: I think in here the sign should be preserved as well for the more fair property 🙂

vrom911 (commented):
Nice property! 👍🏼 

I would just suggest using `hedgehog`.
Specifically, there is `hspec-hedgehog` which would nicely integrate into our existing test suite, no need to add a new stanza.
Also, you could add it into the Chapter1 tests immediately, so it would be checked on CI during the PR review 🙂  Let me know what you think!

dariodsa:
> Nice property! 👍🏼
> 
> I would just suggest using `hedgehog`.
> Specifically, there is `hspec-hedgehog` which would nicely integrate into our existing test suite, no need to add a new stanza.
> Also, you could add it into the Chapter1 tests immediately, so it would be checked on CI during the PR review slightly_smiling_face Let me know what you think!

I agree that is really a great suggestion. I rewrite the code so I think it should be ok right now.

vrom911 (approved):
This is an awesome tests property! Thank you very much 🤗

test/Test/Chapter1.hs
```diff
@@ -56,6 +64,10 @@ chapter1normal = describe "Chapter1Normal" $ do
         it "sumLast2 0 > -10" $ sumLast2 (-9) `shouldBe` 9
         it "sumLast2 -10 > -100" $ sumLast2 (-56) `shouldBe` 11
         it "sumLast2 -100 > -1000" $ sumLast2 (-987) `shouldBe` 15
+    describe "Task 4 & 5 : first and last digit" $ do
+        it "last digit is the first digit of the reversed number" $ hedgehog $ do
+            x <- forAll $ Gen.int (Range.linear (-200) 200)
+            firstDigit x === lastDigit (reverseInt x)
```
vrom911, line 70: We don't have type annotation in the src as it is a part of the task, that is why we have a strange error on CI. This should help 🙂 
```suggestion
            (firstDigit x :: Int) === (lastDigit (reverseInt x) :: Int)
```

PR #47, opened 2020-10-01
Chapter One

drupol:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -204,31 +204,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -503,7 +504,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if abs x == min (abs x) (abs y) then x else y
```
chshersh, line 507: That's a valid implementation, but I think you can slightly improve it. Note, that you only need to compare `abs x` and `abs y` only once to get the result.

src/Chapter1.hs
```diff
@@ -614,8 +628,12 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
+sumLast2 :: Int -> Int
+sumLast2 n =
+  let units = mod n 10
+      dozen = mod (div (n - units) 10) 10
+  in units + dozen
```
chshersh, line 635: Nice solution! Though, you can notice that you can reuse the `lastDigit` function, you've already implemented, instead of `mod x 10`. This is not critical, just a suggestion :slightly_smiling_face: 
Alternatively, you can have a look into the `divMod` function, that returns the result of `div` and `mod` at the same time.

And one more suggestion: I think you don't need to do `n - units`, and `div n 10` will work jus fine. Maybe I'm missing some corner case with negative numbers, and you thought about this more than I :slightly_smiling_face:

chshersh (approved):
Great work! :crossed_swords: :shield: 
You solutions are great. If you want to improve them even further, you can think more about different corner cases and adjust some functions. But nevertheless, you got it!

chshersh:
@drupol The solutions look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to open PRs with solutions to other chapters!

PR #48, opened 2020-10-01
Characterize strong and static typing separately

mjcz:
I think it's worth to differentiate between strong and static typing here


vrom911 (approved):
This is a great suggestion! 👍🏼

chshersh (approved):
@mjcz Thanks, that's a nice improvement!

PR #49, opened 2020-10-01
Complete Chapter 1

danielboros:
### Solutions for Chapter 1

Thanks for the effort in putting this together! I'm liking it so far.

I'm running on Windows on my work laptop so I just copied and pasted the `cabal` test commands from the `Makefile` directly into my Powershell as a stop gap from not having `make` and they all passed. Then I realized I could install `make` with Chocolatey (same as the official way of installing Haskell on Windows) and that worked OK for running the various targets. Just FYI since there weren't any Windows instructions.

I had to do [this](https://www.stackage.org/blog/2020/08/ghc-8-10-2-windows-workaround) to deal with a GHC 8.10.2 issue on Windows which prevented the test suite from building correctly. Probably should have just used Linux...

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -204,31 +204,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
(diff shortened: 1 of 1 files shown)


danielboros:
Closed because I opened it against the main repo. Yikes.

PR #53, opened 2020-10-02
chapter1 completed

err0r500:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -417,6 +429,7 @@ task is to specify the type of this function.
 49
 -}
 
+squareSum :: Num a => a -> a -> a
```
chshersh, line 432: Nice work on specifying the most general type for this function! In this task, we don't require to specify the polymorphic type, as we haven't explained it yet. But you are already one step ahead! :top:

src/Chapter1.hs
```diff
@@ -437,7 +450,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next = (+) 1
```
chshersh, line 453: Nice work on partial application and eta-reduction! :pie:

src/Chapter1.hs
```diff
@@ -477,8 +490,8 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Integral a => a -> a
```
chshersh, line 493: Again, nice work on polymorphism! :clap:

src/Chapter1.hs
```diff
@@ -508,7 +521,8 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y =
+  let absn = min (abs x) (abs y) in if absn == abs x then x else y
```
chshersh, line 525: The implementation of this function is valid. But I see that it can be slightly improved. In your code, you use `abs` several time and introduce a helper variable, but in this task, it's enough to compare two absolute values only once to get the final result.

chshersh (approved):
Excellent work on Chapter One! And congrats with solving the Advanced Task! :trophy: 

You did a great job on polymorphic functions and writing clean code, I enjoyed reviewing your solutions :hugs:

src/Chapter1.hs
err0r500: oh yeah, sorry ! ```if (abs x) < (abs y) then x else y```  :facepalm: 

I must say, I wasn't expecting anyone to actually review the PR, thanks a lot !

PR #57, opened 2020-10-02
Solution to Chapter 1

Couraxe:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -204,31 +204,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -204,31 +204,31 @@ So, the output in this example means that 'False' has type 'Bool'.
...
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+Num p => p
```
chshersh, line 211: We haven't covered this in this Chapter as it is a bit more advanced concept, but in reality it will be 
```suggestion
42 :: Num a => a
```
We will explain what it exactly means in Haskell later, but for now, you can think of it as a `Num`eric object. If you look at it this way, it indeed makes sense, as 42 could represent `Int`, `Natural`, `Integer` and even `Double`! So it is a generalised method to say the type of any numeric 🙂 

Btw, we check this example with some test, that is why it requires the exact output of the GHCI on that. Let us know if we could help with running GHCI. If you installed everything that was in the instructions, you should be able to:

```
$ ghci
```

src/Chapter1.hs
```diff
@@ -432,7 +433,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next = (+) 1
```
chshersh, line 436: Nice work on using partial application and eta-reduction!

src/Chapter1.hs
```diff
@@ -473,8 +474,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
-
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
```
chshersh, line 478: The implementation in here is almost correct. Unfortunately, it will fail on negative numbers because of how mod works for them. Maybe you can think of a way to fix this?


src/Chapter1.hs
```diff
@@ -503,7 +504,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = min (abs x) (abs y)
```
chshersh, line 507: This implementation is almost correct. Unfortunately, for a negative number, it returns its absolute value instead of the negative number itself. So, `closestToZero (-7) 3` will return `7` instead of `-7`.

src/Chapter1.hs
```diff
@@ -615,7 +627,10 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = let d0 = mod n 10
+                 d1 = (mod (div n 10) 10)
+             in d0 + d1
```
chshersh, line 633: On a side note, there is a standard function as well that combines both `div` and `mod` at once: `divMod`.

```haskell
```



```haskell
```

src/Chapter1.hs
```diff
@@ -636,7 +651,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+    | n > 0 && n < 10 = n
+    | otherwise = firstDigit (div n 10)
```
chshersh, line 657: The implementation is correct, but it works only for positive numbers. You probably can use one of the familiar functions to handle negative numbers :slightly_smiling_face:

chshersh (approved):
Great solutions, and well done on the advanced task :medal_sports: 

I left a few comments with some suggestions :slightly_smiling_face:

src/Chapter1.hs
Couraxe: Thanks for the remarks, I directly typed the answers from my head and didn't even run `ghci`, maybe I should have done that :)

src/Chapter1.hs
Couraxe: Damn it, I kind of overread that subtlety :)

src/Chapter1.hs
vrom911: That's great that you checked yourself first, this is actually a very healthy thing to do, great one 💪🏼

chshersh:

PR #58, opened 2020-10-02
Solutions for Chapter 1

Volence:

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -204,31 +204,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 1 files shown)


Volence:
Saw some issues from hintman and touched up a few things, just committed some new junk

src/Chapter1.hs
```diff
@@ -615,7 +625,11 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = lastDigit n + secondLast n
+      where
+        secondLast :: Int -> Int
+        secondLast n = (`div` 10) $ (abs n) `mod` 100 - lastDigit n
```
chshersh, line 632: Nice work on reusing already defined `lastDigit`! Though, I see that the implementation for `secondLast` can be improved a bit. How about, if you do `div n 10` and then just `mod result 10`?

src/Chapter1.hs
```diff
@@ -636,7 +650,8 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n = read [head $ show n]
```
chshersh, line 654: Very creative solution! :smile: :art:

chshersh (approved):
Nice solutions! And well done on the Advanced Task :tada: 

I left some comments and suggestions with possible improvements to the existing code :slightly_smiling_face:

src/Chapter1.hs
```diff
@@ -537,7 +539,14 @@ value after "=" where the condition is true.
...
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | x > y && x < z = x
+    | y > x && y < z = y
+    | z > x && z < y = z
+    | x > z && x < y = x
+    | y > z && y < x = y
+    | z > y && z < x = z
```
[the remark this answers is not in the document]
Volence: That makes sense! I was looking at other solutions after this, completely forgot about the otherwise

src/Chapter1.hs
```diff
@@ -551,7 +560,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel = (`elem` "aeiou")
```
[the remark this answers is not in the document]
Volence: I didn't know this was called eta-reductions, thanks! And that's interesting on the speed. It's still performant enough in most cases to keep this solution right? Only if I were looking to get as optimal speed as possible would I want to change it?

src/Chapter1.hs
Volence: That makes a lot more sense!

src/Chapter1.hs
Volence: Thank you for all the feedback so far! Haskell is very... daunting, but I love learning what I can about it!

src/Chapter1.hs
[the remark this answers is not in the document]
chshersh: We explain eta-reduction in Chapter Two, and you are just one step ahead :foot: 

Sure, you can keep your solution! No need to change anything :slightly_smiling_face: It is just an interesting fact. And, fortunately, the next version of GHC will implement proper optimizations around `elem` in such cases, so no need to worry at all!

chshersh:
@Volence The solutions look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to open PRs with solutions to other chapters!

PR #60, opened 2020-10-02
chapter 2

err0r500:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -500,7 +515,9 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird42 :: (Num a, Eq a) => [a] -> Bool
+isThird42 (_ : _ : x : _) = x == 42
+isThird42 _               = False
```
vrom911, line 520: Great! You made it work with the wider number of types 👍🏼 

Also, note, that you could pattern match on the number as well, so you can write:

```haskell
isThird42 (_ : _ : 42 : _) = True
```

src/Chapter2.hs
```diff
@@ -620,10 +638,16 @@ Write a function that takes elements of a list only on even positions.
...
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven = go 1
+ where
+  go :: Int -> [a] -> [a]
+  go _   []       = []
+  go acc (x : xs) = if mod acc 2 == 0 then rest else x : rest
+    where rest = go (acc + 1) xs
```
vrom911, line 647: Nice idea on the helper functions :+1:

src/Chapter2.hs
```diff
@@ -727,7 +751,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate = concatMap (\x -> replicate x x)
```
vrom911, line 754: Nice usage of one function + eta reduction 😎

src/Chapter2.hs
```diff
@@ -740,11 +764,14 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: (Eq a) => a -> [[a]] -> [[a]]
+contains _ []       = []
+contains c (x : xs) = if elem c x then x : rest else rest
+  where rest = contains c xs
```
vrom911, line 770: Great!

You could also use `filter` function, that does actually the same that you wrote in here 🙂 But it is great that you can easily implement standard functions, great job 👍🏼

src/Chapter2.hs
```diff
@@ -780,13 +807,13 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
--- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Ord a => a -> [a] -> [a]
+listElementsLessThan x = filter (< x)
 
--- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: Num c => [c] -> [c] -> [c]
+pairMul = zipWith (*)
```
vrom911, line 816: Eta-reduction reward 🏆

src/Chapter2.hs
```diff
@@ -857,7 +885,20 @@ and reverses it.
...
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+
+-- I guess I miss something...
+rewind' :: [a] -> [a]
+rewind' []       = []
+rewind' (x : xs) = rewind xs ++ [x]
+
+-- this version prepends to target list so it may be more efficient
```
vrom911, line 894: You are absolutely right!

In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

You can read a bit more about the `go` pattern in here if interested: https://kowainik.github.io/posts/haskell-mini-patterns#recursive-go

vrom911 (approved):
Perfect 👌🏼  Chapter Two is done ✔️ 

Also great investigation on the advanced task! You are correctly wrote everything, great stuff! 🎉

chshersh:
@err0r500 The solutions look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to open PRs with solutions to other chapters!

PR #62, opened 2020-10-02
Update Test of Chapter2.hs

pro2s:
Fix typo in the Test description for task 12*

cc @vrom911 @chshersh

### DIFF
--- a/test/Test/Chapter2.hs
+++ b/test/Test/Chapter2.hs
@@ -69,7 +69,7 @@ chapter2normal = describe "Chapter2Normal" $ do
         it "empty on negative" $ rotate (-5) [1..5] `shouldBe` emptyInts
 
 chapter2advanced :: Spec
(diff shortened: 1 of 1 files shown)


vrom911 (commented):
Hey @pro2s ! Nice catch!

Though, we won't be able to accept this, as this has the answers for Chapter 1 😄 
Could you Amend this commit with only test file change?

pro2s:
> Hey @pro2s ! Nice catch!
> 
> Though, we won't be able to accept this, as this has the answers for Chapter 1 😄
> Could you Amend this commit with only test file change?

My bad, I will create a branch and update the PR

pro2s:
@vrom911, Fixed, I reverted my `master` to correct state.

vrom911 (approved):
Great! Thank you so much 🤗

PR #63, opened 2020-10-02
First chapter completed

dreamsmasher:
### Solutions for Chapter 1.

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -668,7 +668,7 @@ firstDigit n
 
    -}
 firstDigit :: Integral a => a -> a
-firstDigit = last . takeWhile (> 0) . iterate (flip div 10) 
+firstDigit = last . takeWhile (> 0) . iterate (`div` 10)
 
 
 {-
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -481,7 +481,7 @@ Implement a function that returns the last digit of a given number.
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit :: (Integral a) => a -> a
-lastDigit = flip rem 10
+lastDigit = abs . flip rem 10
 
 
 {- |
@@ -512,7 +512,8 @@ branches because it is an expression and it must always return some value.
 -}
 
 closestToZero :: Int -> Int -> Int
-closestToZero a b = min (abs a) (abs b)
+closestToZero a b = if (abs a < abs b) then a else b
+ 
 
 
 {- |
@@ -546,12 +547,15 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
+qsort :: Ord a => [a] -> [a]
+qsort [] = []
+qsort [x] = [x]
+qsort (x:xs) = smol ++ (x : lorg)
+    where smol = qsort [z | z <- xs, z <= x]
+          lorg = qsort [z | z <- xs, z > x]
 
 mid :: Ord a => a -> a -> a -> a
-mid x y z
-  | x <= y && y <= z = y
-  | y <= x && x <= z = x
-  | otherwise = z
+mid x y z = (!! 1) (qsort [x,y,z])
 
 
 {- |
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -416,7 +418,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: (Num a) => a -> a -> a
```
chshersh, line 421: We explain `Num` in the following chapters, so you are one step ahead 😸

src/Chapter1.hs
```diff
@@ -437,7 +439,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next = succ
```
chshersh, line 442: Nice usage of standard functions! :books:

src/Chapter1.hs
```diff
@@ -478,7 +480,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: (Integral a) => a -> a
+lastDigit = abs . flip rem 10
```
chshersh, line 484: Clever solution! :brain: And nice work on using function composition :+1:

src/Chapter1.hs
```diff
@@ -542,7 +547,16 @@ value after "=" where the condition is true.
...
-mid x y z = error "mid: not implemented!"
+qsort :: Ord a => [a] -> [a]
+qsort [] = []
+qsort [x] = [x]
+qsort (x:xs) = smol ++ (x : lorg)
+    where smol = qsort [z | z <- xs, z <= x]
+          lorg = qsort [z | z <- xs, z > x]
+
+mid :: Ord a => a -> a -> a -> a
+mid x y z = (!! 1) (qsort [x,y,z])
```
chshersh, line 558: Very creative solution! :smile: Looks like you already mastered the List topic in Chapter one! Though, of course, reusing the standard `sort` function will result in a shorter code :slightly_smiling_face:

src/Chapter1.hs
```diff
@@ -556,7 +570,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel = flip elem "AEIOUaeiou"
```
chshersh, line 574: Absolutely correct! And you use eta-reduction 👍🏼 

src/Chapter1.hs
```diff
@@ -641,7 +660,15 @@ You need to use recursion in this task. Feel free to return to it later, if you
...
-firstDigit n = error "firstDigit: Not implemented!"
+{-
+-- naive approach:
+firstDigit n
+ | n < 10 && n > (-10) = n
+ | otherwise = firstDigit (div n 10)
+
+   -}
+firstDigit :: Integral a => a -> a
+firstDigit = last . takeWhile (> 0) . iterate (`div` 10)
```
chshersh, line 671: Very clever solution with the usage of standard function! :exploding_head:

chshersh (approved):
Amazing work on Chapter One! And bonus points for solving an Advanced Task in a such nice way :trophy: 

I reviewed only the first Chapter, but I see that you already have commits for the second chapter in this PR. I would encourage you to open a separate PR with solutions for Chapter Two, so you can get 4 PRs of contributions for Hacktoberfest. And it is also easier to review for us :slightly_smiling_face:

dreamsmasher:
Thank you for the feedback! I already have some experience with Haskell so I feel bad using up your time reviewing my code so deeply.

src/Chapter1.hs
dreamsmasher: Interesting, it makes sense since guards and case expressions just compile down to jump tables iirc.

src/Chapter1.hs
dreamsmasher: Wanted to save a dependency ;)

chshersh:
@dreamsmasher Don't worry! I'm happy if you find my review useful :slightly_smiling_face: 

PR #64, opened 2020-10-02
Complete chapter 1

OxSon:
### Solutions for Chapter 1

This was harder than I had expected. Lots of edge cases that I missed at first for `firstDigit`. Took me a few times to realize what was happening. Thanks for putting this together!

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -72,6 +72,8 @@ the `.hs` extension.
 -}
 module Chapter1 where
 
+import Data.List ( sort )
+
 {- |
 In Haskell, we have __expressions__. Expressions can be represented by some
 primitive values (numbers: 1, 100; characters: 'a', 'z'; booleans: True, False;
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -437,7 +440,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next = succ
```
vrom911, line 443: Nice usage of standard functions 👍🏼 Great research work!

src/Chapter1.hs
```diff
@@ -478,8 +481,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
-
+lastDigit :: Integral a => a -> a
+lastDigit n = (abs n) `mod` 10
```
vrom911, line 485: Yeah, you can use functions as operators with backticks, isn't it cool?! 💯

src/Chapter1.hs
```diff
@@ -508,8 +511,8 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
-
+closestToZero x y = if (abs x) < (abs y) then x else y
+  --min (abs x) (abs y)
```
vrom911, line 515: That is a nice guess! 👍🏼 
However, this solution doesn't preserve sign, so for negative numbers, it still will return positive.

src/Chapter1.hs
```diff
@@ -542,7 +545,8 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Integral a => a -> a -> a -> a
+mid x y z = (sort [x, y, z]) !! 1
```
vrom911, line 549: Very creative solution 🎨 😃

src/Chapter1.hs
```diff
@@ -556,7 +560,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel = (`elem` ['a', 'e', 'i', 'o', 'u'])
```
vrom911, line 564: Absolutely correct! And you use eta-reduction 👍🏼 

src/Chapter1.hs
```diff
@@ -620,8 +625,12 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Integral a => a -> a
+sumLast2 n = lastD + secondToLastD
+  where
+    n' = abs n
+    lastD = n' `mod` 10
+    secondToLastD = (div n' 10) `mod` 10
```
vrom911, line 633: Nice solution! 🍬 

You could also reuse `lastDigit` from the previous task instead of `mod 10` 🙂

vrom911 (commented):
Congratulation with finishing Chapter 1 ⭐  Very elegant solutions! Keep up the great work!
And, btw, you nailed the advanced task 🥇

PR #65, opened 2020-10-03
Completed Chapter 1

pramalin:
### Solutions for Chapter {1}

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -479,7 +479,7 @@ Implement a function that returns the last digit of a given number.
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit :: Int -> Int
-lastDigit n = mod (abs n) (10 * div (abs(n)) 10)
+lastDigit n = mod (abs n) (10 * div (abs n ) 10)
 
 {- |
 =⚔️= Task 6
@@ -508,7 +508,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = if abs(x) < abs(y) then x else y
+closestToZero x y = if abs x  < abs y  then x else y
 
 
 {- |
@@ -543,8 +543,8 @@ Casual reminder about adding top-level type signatures for all functions :)
 -}
 mid :: Int -> Int -> Int -> Int
 mid x y z
-   | (x > y && x < z) = x
-   | (y > x && y < z) = y
+   | x > y && x < z = x
+   | y > x && y < z = y
    | otherwise = z
 {- |
 =⚔️= Task 8
@@ -629,10 +629,10 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 sumLast2 :: Int -> Int
-sumLast2 n = last(div n 10) + last(n)
+sumLast2 n = last(div n 10) + last n 
   where
      last :: Int -> Int
-     last n = mod (abs(n)) (10 * div (abs(n)) 10)
+     last n = mod (abs n) (10 * div (abs n) 10)
 
 {- |
 =💣= Task 10*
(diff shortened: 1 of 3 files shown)


src/Chapter1.hs
```diff
@@ -301,31 +301,31 @@ expressions in GHCi
...
+15
 >>> (3 + 5) < 10
-
+True
 >>> True && False
-
+False
 >>> 10 < 20 || 20 < 5
-
+False
```
vrom911, line 314: Here it would be `True`, as `||` (OR) operator would immediately return `True` when sees `10 < 20` 🙂

src/Chapter1.hs
```diff
@@ -478,8 +478,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
-
+lastDigit :: Int -> Int
+lastDigit n = mod (abs n) (10 * div (abs n ) 10)
```
vrom911, line 482: Creative solution 🎨 

However, it won't work on `-10 < n < 10`, as you will have `mod 5 (10 * div 5 10) = mod 5 (10 * 0) = mod 5 0`  – and this is division on 0 💥 

The solution is a bit simpler 🙂

src/Chapter1.hs
```diff
@@ -619,9 +628,11 @@ Implement a function that returns the sum of the last two digits of a number.
...
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n = last(div n 10) + last n 
+  where
+     last :: Int -> Int
+     last n = mod (abs n) (10 * div (abs n) 10)
```
vrom911, line 635: This function has the same issue as above `lastDigit` one. Once that is implemented, then you can just reuse it here instead of `last` and your solution would work perfectly! 👍🏼

src/Chapter1.hs
```diff
@@ -619,9 +628,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n = last(div n 10) + last n
```
vrom911, line 632: In Haskell, you usually add space between each argument, even those in brackets 🙂 
```suggestion
sumLast2 n = last( div n 10) + last n 
```

src/Chapter1.hs
```diff
@@ -541,9 +541,11 @@ value after "=" where the condition is true.
...
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
-
+mid :: Int -> Int -> Int -> Int
+mid x y z
+   | x > y && x < z = x
+   | y > x && y < z = y
+   | otherwise = z
```
vrom911, line 548: Elegant! 💃🏼

src/Chapter1.hs
```diff
@@ -640,8 +651,10 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+  | n < 10 = n
+  | otherwise = firstDigit (div n 10)
```
vrom911, line 657: Great recursion function! Well done 👏🏼 
Almost there! This function should take into consideration negative numbers as well.

vrom911 (approved):
Amazing work! Congrats on finishing Chapter One 🎉 
And great one with the Advanced task 🌟

vrom911:
Great! I think this PR is done!

Could you please open the new one for Chapter 2 separately? I then will close this PR, if you don't mind 🙂

PR #67, opened 2020-10-03
Solutions #1

jamesthesnake:
### Solutions for Chapter 1

cc @vrom911 @chshersh


src/Chapter1.hs
```diff
@@ -209,31 +209,29 @@ So, the output in this example means that 'False' has type 'Bool'.
...
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
-
+(True,'x') :: (Bool,Char)
+ 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
-
+not :: Bool->Bool
```
chshersh, line 224: I think, doctests will fail on this line, as the output is not exactly the same as GHCi produces.

src/Chapter1.hs
```diff
@@ -436,6 +434,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   every type ｡.☆.*｡. No need to worry much about "error" here, just replace the
   function body with the proper implementation.
 -}
+next x =x +1
 next :: Int -> Int
 next x = error "next: not implemented!"
```
chshersh, line 439: You need to replace the code after `next x =` on the second line to solve this task. In Haskell, you are not allowed to have multiple definitions of the same function. Though in this particular case something different is happening, so this code compiles but behaves differently.

To solve this task successfully, you need to write the following function definition:

```haskell
next :: Int -> Int
next x = x + 1
```

src/Chapter1.hs
```diff
@@ -466,7 +465,7 @@ Implement a function that returns the last digit of a given number.
 
 >>> lastDigit 42
 2
-
+lastDigit x = mod x 10
```
chshersh, line 468: The implementation in here is almost correct. Unfortunately, it will fail on negative numbers because of how mod works for them. Maybe you can think of a way to fix this?


chshersh: Also, I see that you've implemented the function inside the comment, but the actual function definition is below this comment after the closing `-}` for the description. You need to move the implementation out of the comment for the compiler to check it. And don't forget to write the top-level function signature!

src/Chapter1.hs
```diff
@@ -508,9 +507,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
-
-
+closestToZero x y = if abs x < (abs y) then x else y
```
chshersh, line 510: `()` are redundant in here, so you can go ahead and remove them for cleaner code :slightly_smiling_face: 
```suggestion
closestToZero x y = if abs x < abs y then x else y
```

src/Chapter1.hs
```diff
@@ -541,8 +538,8 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+import Data.List (sort)
```
chshersh, line 541: Imports in Haskell should go after the `module ... where` and before the first definition of a function. Unfortunately, you are not allowed to write `import` statements between function definitions. For this to work, you need to move this import up. We explain `imports` in more details in the next chapter.

src/Chapter1.hs
```diff
@@ -541,8 +538,8 @@ value after "=" where the condition is true.
...
+mid x y z =(sort [x,y,z]) !!1
```
chshersh, line 542: Nice work on using standard functions and lists for this task! :+1:

src/Chapter1.hs
```diff
@@ -556,7 +553,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel = (elem ['a','e','i','o','u'])
```
chshersh, line 557: Absolutely correct! And you use eta-reduction 👍🏼 

src/Chapter1.hs
```diff
@@ -640,8 +642,10 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Integral a=> a->a 
+firstDigit n = if n' <10 then  n' else firstDigit(div n' 10)
```
chshersh, line 646: One minor stylistic suggestion: operators and function arguments are usually separated by spaces for better visibility.
```suggestion
firstDigit n = if n' < 10 then  n' else firstDigit (div n' 10)
```

chshersh (approved):
Nice work on solving our challenges! And well done with the Advanced Task :1st_place_medal: 

I reviewed only the first Chapter, but I see that you already have commits for the second chapter in this PR. I would encourage you to open a separate PR with solutions for Chapter Two, so you can get 4 PRs of contributions for Hacktoberfest. And it is also easier to review for us :slightly_smiling_face:

jamesthesnake:
> Nice work on solving our challenges! And well done with the Advanced Task 1st_place_medal
> 
> I reviewed only the first Chapter, but I see that you already have commits for the second chapter in this PR. I would encourage you to open a separate PR with solutions for Chapter Two, so you can get 4 PRs of contributions for Hacktoberfest. And it is also easier to review for us slightly_smiling_face

yeah my bad messed up my git  :0

jamesthesnake:
I think this is all fixed! @chshersh

jamesthesnake:
@chshersh @vrom911 , I think chapter one is stuck in a loop for the CI check, any idea why?

src/Chapter1.hs
```diff
@@ -209,31 +210,29 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
```
chshersh, line 214: I see that doctests are failing on this line because of extra space. Other lines also have some extra spaces. Maybe you can try to remove them, run tests locally and see, where exactly tests are failing?

jamesthesnake:
Hey @vrom911  @chshersh  I think we are all good now, I think I need the octoberfest label and we are good to go with one more apporval :)

chshersh:
@jamesthesnake Well done on solutions! One approval should be enough, and I've added the Hacktoberfest label just in case :slightly_smiling_face: I'm closing this PR, as no further action is required.

PR #69, opened 2020-10-03
Solutions for Chapter 1

Alison-Li:
### Solutions for Chapter 1

cc @vrom911 @chshersh

Thanks for taking the time to review my solutions! I'm particularly seeking feedback for my solution to the final boss question, but a brief look over my other solutions would be appreciated as well! I feel like I should be using `let ... in ...` but I can't seem to get the hang of using it :(

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -480,7 +480,7 @@ Implement a function that returns the last digit of a given number.
 -}
 
 lastDigit :: Integer -> Integer
-lastDigit n = mod n 10
+lastDigit n = mod (abs n) 10
 
 
 {- |
@@ -549,8 +549,7 @@ mid :: Int -> Int -> Int -> Int
 mid x y z
     | x < y && x > z || x < z && x > y = x
     | y < x && y > z || y < z && y > x = y
-    | z < y && z > x || z < x && z > y = z
-    | otherwise = error "mid: no middle found!"
+    | otherwise = z
 
 {- |
 =⚔️= Task 8
@@ -638,8 +637,7 @@ specifying complex expressions.
 sumLast2 :: Int -> Int
 sumLast2 n = x + y
   where
-    x = mod (mod n 100) 10
-    y = div (mod n 100) 10
+    (x, y) = divMod (mod (abs n) 100) 10
 
 
 {- |
@@ -664,7 +662,9 @@ firstDigit :: Int -> Int
 firstDigit n = helper n
   where
     helper :: Int -> Int
-    helper n = if div n 10 == 0 then n else helper (div n 10)
+    helper n = 
+      let nextN = div n 10
+      in if nextN == 0 then n else helper nextN
 
 
 {-
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -477,8 +478,9 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+
+lastDigit :: Integer -> Integer
+lastDigit n = mod n 10
```
chshersh, line 483: The implementation in here is almost correct. Unfortunately, it will fail on negative numbers because of how mod works for them. Maybe you can think of a way to fix this?

src/Chapter1.hs
```diff
@@ -542,7 +545,12 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | x < y && x > z || x < z && x > y = x
+    | y < x && y > z || y < z && y > x = y
+    | z < y && z > x || z < x && z > y = z
+    | otherwise = error "mid: no middle found!"
```
chshersh, line 553: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

So, it looks like you can get rid of redundant `error` case and make the implementation shorter by returning `z` on the `otherwise` case and collapse last two guards into a single one.

src/Chapter1.hs
```diff
@@ -620,7 +635,11 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = x + y
+  where
+    x = mod (mod n 100) 10
+    y = div (mod n 100) 10
```
chshersh, line 642: I think that this function doesn't work properly on negative numbers, as `div` and `mod` behave differently for them.

On a side note, there is a standard function as well that combines both `div` and `mod` at once: `divMod`.

```haskell
```



```haskell
```

src/Chapter1.hs
```diff
@@ -641,7 +660,11 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n = helper n
```
chshersh, line 664: That's a very nice idea to use a _helper_ recursive function! If you pass `abs n` instead of `n` to this function in here, you will handle negative numbers automatically, and you will call `abs` only once in the beginning.

Usually, such helper recursive function are called `go` in Haskell. You can read more about such pattern in our blog post, but don't worry to much for now :slightly_smiling_face: We cover recursion and this pattern in more detail in the next Chapter.

* https://kowainik.github.io/posts/haskell-mini-patterns#recursive-go

src/Chapter1.hs
```diff
@@ -641,7 +660,11 @@ You need to use recursion in this task. Feel free to return to it later, if you
...
+  where
+    helper :: Int -> Int
+    helper n = if div n 10 == 0 then n else helper (div n 10)
```
chshersh, line 667: Your implementation is correct :+1: 
And it looks fine. Instead of `if ... then ... else` you can use guards, but that's a matter of taste :slightly_smiling_face: 

Here you can assign `div n 10` to a variable and reuse it.

```suggestion
    helper n =
        let nextN = div n 10
        in if nextN == 0 then n else helper nextN
```

I see that you were confused about the `let ... in` syntax, and you have all rights for that! Let-in can be confusing indeed. In general, nobody encourages you to use `let` more, we just want people to learn Haskell syntax basics in here. In some cases `where` works better, in others `let` are good as well. Most of the time it's a matter of taste. I personally prefer guards instead of `if` and `where` instead of `let` :slightly_smiling_face:

chshersh (approved):
Nice solutions for Chapter One task!
And very well done on the Advanced Task :trophy: 

I left some comments and tried to explain some things with more details. I hope my review clarifies better some of the topic for your :slightly_smiling_face:

src/Chapter1.hs
Alison-Li: Thank you! Something I noticed with using guards directly in a function is that `=` doesn't follow right after the function name and arguments? For example, the following doesn't compile due to the `=` in `mid x y z =`:
```haskell
mid :: Int -> Int -> Int -> Int
mid x y z =
    | x < y && x > z || x < z && x > y = x
    | y < x && y > z || y < z && y > x = y
    | otherwise = z
```
This was interesting to me since guards look similar to pattern matching in another language I'm familiar with, OCaml. I'm guessing it's because we're doing a pattern matching on the function arguments `x`, `y`, `z`? Is this syntax kind of like a shortcut for saying:
```
mid x y z =
    match (x, y, z)
        | ...
        | ...
        | ...
```
?

chshersh: @Alison-Li Yes, your intuition is correct. We are exactly doing pattern-matching on variables, but it's a special case of pattern matching — a pattern that captures anything and just assigns the variable name to it. The `=` sign goes after each guard because Haskell validates all checks from top to  bottom and returns the expression on a first guard that succeeds.

chshersh:

PR #70, opened 2020-10-03
first chapter

Gexeg:
### Solutions for Chapter 1
cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -478,7 +478,7 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = mod n 10
+lastDigit n = mod (abs n) 10
 lastDigit :: Int -> Int
 
 {- |
@@ -508,7 +508,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = if (abs x) > (abs y) then y else x
+closestToZero x y = if abs x > abs y then y else x
 
 
 {- |
@@ -546,7 +546,7 @@ mid x y z
     | (y < x) && (y > z) || (y > x) && (y < z) = y
     | (z < x) && (z > y) || (z > x) && (z < y) = z
     | otherwise = x
-mid :: Int -> Int -> Int -> Int 
+mid :: Int -> Int -> Int -> Int
 
 {- |
 =⚔️= Task 8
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -418,7 +418,7 @@ task is to specify the type of this function.
 -}
 
 squareSum x y = (x + y) * (x + y)
-
+squareSum :: Integer -> Integer -> Integer
```
vrom911, line 421: That's correct!

In Haskell, the type signature usually goes before the function body 🙂

src/Chapter1.hs
```diff
@@ -556,8 +560,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
-
+isVowel c = elem c  "aeiouy" || elem c "AEIOUY"
```
vrom911, line 563: One note in here, that sometimes, `elem` could be slower than the explicit pattern matching. I remember there were some benchmarks on one particular case, that showed how moving to pattern matching on each case separately drastically decrease time 🐎

src/Chapter1.hs
```diff
@@ -620,8 +624,15 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 n = last1 n + last2 n
+    where
+        last1 x = mod (abs x) 10
+        last1 :: Int -> Int
+
+        last2 x = div (mod (abs x) 100) 10
+        last2 :: Int -> Int
```
vrom911, line 633: Great solution!

You could also reuse the `lastDigit` function in here, instead of `last1` 🙂

src/Chapter1.hs
```diff
@@ -641,8 +652,11 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit n
+    | mod n 10 == n = n
+    | otherwise = firstDigit (div n 10)
```
vrom911, line 657: That is the right idea! 💡 
But, it won't work correctly on negative numbers. It would show the wrong output for ` -10 < x < 0` (in case of negatives, the result will be negative). And on the other negatives in will hang.

vrom911 (approved):
Awesome work! Congrats on finishing Chapter One 🍰

vrom911:
I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to open PRs with solutions to other chapters!

PR #71, opened 2020-10-03
Chapter two - solutions

drupol:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -338,6 +338,8 @@ ghci> :l src/Chapter2.hs
 subList :: Int -> Int -> [a] -> [a]
 subList from to x
   | to < from = []
+  | to < 0 = []
+  | from < 0 = []
   | otherwise =
     let size = to - from + 1
     in take size (drop from x)
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -865,7 +865,9 @@ list.
 🕯 HINT: Use the 'cycle' function
 -}
 rotate :: Int -> [Int] -> [Int]
-rotate i l = take (length l) (drop i (cycle l))
+rotate i l
+  | i < 0 = []
+  | otherwise = take (length l) (drop i (cycle l))
 
 {- |
 =💣= Task 12*
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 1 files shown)


drupol:
I submit the answer, but I have to say that Hint Man gave me already very nice hints for the next tasks.

src/Chapter2.hs
```diff
@@ -349,7 +355,11 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 -- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf l =
+  let pivot = div (length l) 2
+  in take pivot l
```
vrom911, line 361: Great usage of the `let-in` construction! 👍🏼

src/Chapter2.hs
```diff
@@ -500,7 +510,9 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird42 :: [Int] -> Bool
+isThird42 (_:_:42:xs) = True
```
vrom911, line 514: That's right!
Moreover, as you don't use `xs` in the body, you can also pattern match on it as on the wilcard:
```suggestion
isThird42 (_:_:42:_) = True
```

src/Chapter2.hs
```diff
@@ -533,7 +545,10 @@ some information during recursive calls.
 For example, we can patch the previous function to count the number of
 steps we need to take in order to reduce the number to zero.
 
-🤔 Blitz question: can you guess what this number represents?
+🤔 Blitz question: can you guess what this number (acc) represents?
+
+The amount of time a number can be divided by 2, plus one. In other words,
+2^[acc] is the first number greater than n, in base 2.
```
vrom911, line 551: Fantastic!
Thank you for sharing the answer on the non-code question as well ❤️

src/Chapter2.hs
```diff
@@ -605,7 +620,8 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate [] = []
+duplicate (x:xs) = replicate 2 x ++ duplicate xs
```
vrom911, line 624: This is the correct implementation!
However, it could be sped up a bit 🙂 

2 is not a big number, so it is manageable to right `x:x` manually.
Also, with such usage, it is possible to get rid of the list concatenation, which is slow, and use `:`.

```suggestion
duplicate (x:xs) = x : x : duplicate xs
```

src/Chapter2.hs
```diff
@@ -620,7 +636,9 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+takeEven [] = []
+takeEven (x:xs) = x : takeEven (drop 1 xs)
```
vrom911, line 641: That is a clever solution! 🧠 

Alternatively, you could pattern match on that as well, e.g. `(x: _ : xs) = x : takeEven xs`.
(Note that you would need to handle the case with one element as well in that case)>

src/Chapter2.hs
```diff
@@ -727,7 +745,9 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate (x:xs) = replicate x x ++ smartReplicate xs
```
vrom911, line 749: Nice one!

Note, that you also can use standard functions to do the same. You see that you actually map `replicate` over each element of the list and then concatenate all the results together. You can take a look at `concatMap` standard function, which behaves exactly like this! 🙂

src/Chapter2.hs
```diff
@@ -740,7 +760,11 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains a [] = []
+contains a (x:xs)
+  | elem a x = x : contains a xs
+  | otherwise = contains a xs
```
vrom911, line 767: Here you also wrote the function, that is similar to the standard `filter` function 👍🏼

src/Chapter2.hs
```diff
@@ -780,13 +804,14 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
+divideTenBy = div 10
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul = zipWith (*)
```
vrom911, line 814: Eta-reduction award 🏆

src/Chapter2.hs
```diff
@@ -857,7 +885,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [Int] -> [Int]
+rewind [] = []
+rewind l = last l : rewind (take (length l - 1) l)
```
vrom911, line 890: Interesting idea and realisation! 😎 

Though, note, that `last` and `length`are very slow, and this implementation would be quite inefficient.
So you could try to implement it without these functions only using `:`.
For that, you would need to use the `go` pattern with the helper function and accumulator. You can read a bit more about the `go` pattern in here: https://kowainik.github.io/posts/haskell-mini-patterns#recursive-go

vrom911 (approved):
Fantastic job on Chapter Two 🎉 
A lot of interesting and creative solutions 👏🏼

src/Chapter2.hs
drupol: Oh this is beautiful ! Thanks ! Amazing language :-)

src/Chapter2.hs
drupol: Indeed, very nice !

vrom911:

PR #76, opened 2020-10-03
chapter 1

Yu-zh:
### Solutions for Chapter 1

Thanks for the awesome Haskell tutorial! I'm a newbie in using GitHub so I don't quite understand how to open PR to my own fork instead of the original repo. It seems I am trying to merge my code into kowainik/learn4haskell. Please let me know if I did anything wrong. Thanks again!

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -416,10 +416,9 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Num a => a -> a -> a
```

src/Chapter1.hs
```diff
@@ -477,8 +476,8 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Integral a => a -> a
+lastDigit n = (abs n) `mod` 10
```
chshersh, line 480: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1: 

```suggestion
lastDigit n = abs n `mod` 10
```

src/Chapter1.hs
```diff
@@ -541,8 +540,8 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: (Num a, Ord a) => a -> a -> a -> a
+mid x y z = (x + y + z) - (min (min x y) z) - (max (max x y) z)
```
chshersh, line 544: That's a very clever implementation! :brain: 
Though, I wonder, will it still work around big numbers because of integer overflow :thinking: 
We don't test such cases in our course, but something to keep in mind when implementing it in real life :slightly_smiling_face:

src/Chapter1.hs
```diff
@@ -556,7 +555,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+  | c `elem` "aeiou" = True
+  | otherwise = False
```
chshersh, line 561: Absolutely correct! And you use eta-reduction 👍🏼 

Also, you can write it a bit shorter. Since `elem` already returns `Bool`, you can just:

```haskell
isVowel c = c `elem` "aeiou"
```

src/Chapter1.hs
```diff
@@ -619,8 +621,12 @@ Implement a function that returns the sum of the last two digits of a number.
...
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Integral a => a -> a
+sumLast2 n = last1 + last2
+  where
+    absN = abs n
+    last1 = absN `mod` 10
+    last2 = (absN `mod` 100) `div` 10
```
chshersh, line 629: Nice implementation!

On a side note, there is a standard function as well that combines both `div` and `mod` at once: `divMod`.

```haskell
```



```haskell
```

chshersh (approved):
Very nice solutions! Seems like you already get Haskell concepts, and even some advanced topics we haven't explained yet :smile: 
And well done on solving the advanced task! :trophy: 

As for your question about opening a PR to your fork. For this, you need to do commits in a different branch from `master`, and then you will be able to choose to open PR to your fork. I pasted a screenshot.

Fork-PR

PR #85, opened 2020-10-04
Chapter 1

Dj-Denis:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -438,7 +438,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
 -}
 next :: Int -> Int
 next x = error "next: not implemented!"
-
+next_x = x + 1
```
chshersh, line 441: Unfortunately, this implementation still has `error`. When writing `next_x`, you don't introduce a variable called `x`. You need to add space between function name and argument. Also, you need to replace the `error "..."` with your implementation.

src/Chapter1.hs
```diff
@@ -479,7 +479,7 @@ Implement a function that returns the last digit of a given number.
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit n = error "lastDigit: Not implemented!"
-
+lastDigit n = (abs n) `mod` 10
```
chshersh, line 482: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1: 

```suggestion
lastDigit n = abs n `mod` 10
```

src/Chapter1.hs
```diff
@@ -509,7 +509,7 @@ branches because it is an expression and it must always return some value.
 -}
 closestToZero :: Int -> Int -> Int
 closestToZero x y = error "closestToZero: not implemented!"
```
chshersh, line 511: As I mentioned before, you need to replace the part with `error "..."` with your implementation for tests to pass

src/Chapter1.hs
```diff
@@ -543,7 +543,7 @@ Casual reminder about adding top-level type signatures for all functions :)
 -}
 
 mid x y z = error "mid: not implemented!"
-
+mid x y z = (x + y + z) - (min (min x y) z) - (max (max x y) z)
```
chshersh, line 546: That's a very clever implementation! :brain: 
Though, I wonder, will it still work around big numbers because of integer overflow :thinking: 
We don't test such cases in our course, but something to keep in mind when implementing it in real life :slightly_smiling_face:

src/Chapter1.hs
```diff
@@ -557,7 +557,9 @@ True
 False
 -}
 isVowel c = error "isVowel: not implemented!"
-
+isVowel c
+  | c `elem` "aeiou" = True
```
chshersh, line 561: Absolutely correct! 👍🏼 

Also, since `elem` already returns `Bool`, you don't need to use guards. You can just do:

```haskell
isVowel c = c `elem` "aeiou"
```

src/Chapter1.hs
```diff
@@ -621,6 +623,11 @@ specifying complex expressions.
 -}
 
 sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 n = last1 + last2
+  where
+    absN = abs n
+    last1 = absN `mod` 10
+    last2 = (absN `mod` 100) `div` 10
```
chshersh, line 630: On a side note, there is a standard function as well that combines both `div` and `mod` at once: `divMod`.

```haskell
```



```haskell
```

chshersh (approved):
Nice solutions! And well with the advanced task :medal_sports: 

The only thing I want to draw attention is that in all implementations you need to replace `error` with your own implementation, otherwise functions won't work properly.

Also, I see that you have solutions to Chapter Two in this PR. I reviewed only Chapter One. I would encourage you to open solutions for Chapter Two in a separate PR, so you can get 4 contributions towards Hacktoberfest, and also it's more convenient for us to review in this way :slightly_smiling_face:

PR #88, opened 2020-10-04
docs(Readme): Avoid unknown ghcup command error

isthatcentered:
I spent the last 15 minutes trying to figure out why I got an `unknown ghcup command` when the installation went perfectly. Then I remembered I had to restart the terminal for the changes to be taken into account 😵. 

This might save a few minutes to someone else :)

cc @vrom911 @chshersh

(Note, I did not create an issue since this seemed way overkill for such a small change but I'll be happy to do so if required 🙂)


vrom911 (approved):
Nice hint! Thanks for sharing this! This is definitely could be useful for other people as well 💟

PR #89, opened 2020-10-04
mullakhmetov / Chapter #2

mullakhmetov:

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,42 +136,52 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
+[True, False] :: [Bool]
 
 
 String is a list of characters:
 >>> :t "some string"
+"some string" :: [Char]
 
 
 Empty list:
 >>> :t []
+[] :: [a]
 
 
 Append two lists:
 >>> :t (++)
+(++) :: [a] -> [a] -> [a]
 
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
+(:) :: a -> [a] -> [a]
 
 
 Reverse a list:
 >>> :t reverse
+reverse :: [a] -> [a]
 
 
 Take first N elements of a list:
 >>> :t take
+take :: Int -> [a] -> [a]
 
 
 Create list from N same elements:
 >>> :t replicate
+replicate :: Int -> a -> [a]
 
 
 Split a string by line breaks:
 >>> :t lines
+lines :: String -> [String]
 
 
 Join a list of strings with line breaks:
 >>> :t unlines
+unlines :: [String] -> String
 
 
 -}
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +359,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList l r list
+  | l < 0 || r < 0 = []
```
chshersh, line 363: You can cover extra case here when `l > r` to return an empty list earlier

src/Chapter2.hs
```diff
@@ -620,7 +651,11 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+takeEven [] = []
+takeEven [x] = [x]
+takeEven [x, _] = [x]
```
chshersh, line 657: This case is actually redundant, as it's covered by the next line.
Note, that `[x, _]` is syntax sugar for `(x : _ : []`).
Though, it is very helpful to cover smaller cases before generalizing :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -740,7 +775,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains x l = filter (elem x) l
```
chshersh, line 779: You can actually eta-reduce this function as well

src/Chapter2.hs
```diff
@@ -841,7 +879,13 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate _ [] = []
+rotate _ [x] = [x]
```
chshersh, line 884: It looks like this case doesn't return an empty list for negative numbers :thinking:

src/Chapter2.hs
```diff
@@ -841,7 +879,13 @@ list.
...
+rotate times l
+  | times < 0 = []
+  | times == 0 = l
+  | otherwise = subList times (times + length l - 1) (cycle l)
```
chshersh, line 888: Well done on reusing the `subList` function in `rotate`! So far, you are the first one :1st_place_medal:

src/Chapter2.hs
```diff
@@ -857,7 +901,10 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [Int] -> [Int]
+rewind [] = []
+rewind [x] = [x]
+rewind (x:xs) = rewind xs ++ [x]
```
chshersh, line 907: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
@mullakhmetov Nice work on solving tasks from Chapter Two!

I left my feedback, but I'm going to approve this PR anyways, as solutions look good :100: 
Feel free to close this PR after changes, or just now if you don't want to spend too much time on this :slightly_smiling_face:

src/Chapter2.hs
mullakhmetov: Thanks for your feedback and encouraging me to recall SICP lessons about tail recursion :)

PR #91, opened 2020-10-04
Completed chapter1

pennychase:
### Solutions for Chapter {1}

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -480,7 +480,7 @@ Implement a function that returns the last digit of a given number.
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit :: Integral a => a -> a
-lastDigit n = (abs n) `mod` 10
+lastDigit n = abs n `mod` 10
 
 
 {- |
@@ -511,10 +511,7 @@ branches because it is an expression and it must always return some value.
 -}
 closestToZero :: Int -> Int -> Int
 closestToZero x y =
-  let
-    closest = min (abs x) (abs y)
-  in if closest == abs x then x else y
-  
+  if abs x < abs y then x else y
 
 {- |
 =⚔️= Task 7
@@ -549,9 +546,12 @@ Casual reminder about adding top-level type signatures for all functions :)
 
 mid :: (Ord a) => a -> a -> a -> a
 mid x y z
-  | z >= max x y    = max x y
-  | z >= min x y    = z
-  | otherwise       = min x y
+  | z >= maxXY   = maxXY
+  | z >= minXY    = z
+  | otherwise       = minXY
+  where
+    maxXY = max x y
+    minXY = min x y
 
 {- |
 =⚔️= Task 8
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -412,6 +412,7 @@ task is to specify the type of this function.
 49
 -}
 
+squareSum :: Num a => a -> a -> a
```
vrom911, line 415: We explain `Num` in the following chapters, so you are one step ahead 😸

src/Chapter1.hs
```diff
@@ -473,7 +474,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Integral a => a -> a
+lastDigit n = (abs n) `mod` 10
```
vrom911, line 478: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1: 

```suggestion
lastDigit n = abs n `mod` 10
```

src/Chapter1.hs
```diff
@@ -503,8 +505,11 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
-
+closestToZero x y =
+  let
+    closest = min (abs x) (abs y)
+  in if closest == abs x then x else y
```
vrom911, line 511: Very interesting solution! 💡 

src/Chapter1.hs
```diff
@@ -537,7 +542,11 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: (Ord a) => a -> a -> a -> a
+mid x y z
+  | z >= max x y    = max x y
+  | z >= min x y    = z
+  | otherwise       = min x y
```
vrom911, line 549: Nice solution!
You see that you use `min` and `max` of x and y several times. You can move it into the variables under the `where` clause and reuse in each guard 👍🏼

src/Chapter1.hs
```diff
@@ -551,7 +560,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c 
+  | c `elem` ['a', 'e', 'i', 'o', 'u']    = True
+  | otherwise                             = False
```
vrom911, line 566: Nice one! 👍🏼 

```haskell
```

src/Chapter1.hs
```diff
@@ -615,7 +627,12 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Integral a => a -> a
+sumLast2 n = 
+  let
+    (q, r) = quotRem (abs n) 10
```
vrom911, line 633: Cool usage of standard functions 👍🏼  Excellent implementation!

vrom911 (approved):
Looks great! Congrats on finishing Chapter One 🎉 
And great job on the advanced task!

vrom911:

PR #97, opened 2020-10-04
Solution to chapter 2

Couraxe:
### Solutions for Chapter 2

cc @vrom911 @chshersh

I appreciate the review! :)

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -186,31 +186,43 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
+[10, 2, 3, 1, 5]
```
chshersh, line 189: Doctest expects strings in the exact same format as GHCi produces. And for lists, GHCi doesn't add spaces. So, to make the doctest pass, you need to remove extra spaces after commas.
```suggestion
[10,2,3,1,5]
```

src/Chapter2.hs
```diff
@@ -500,8 +518,12 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
 
+isThird42 :: [Int] -> Bool
+isThird42 [] = False
+isThird42 [x] = False
+isThird42 [x, y] = False
```
chshersh, line 525: If you put the first case on the first line, you can just pattern match on `_` in the last case, as you are interested only in capturing one specific pattern

src/Chapter2.hs
```diff
@@ -500,8 +518,12 @@ True
...
+isThird42 (x:y:z:_) = z == 42
```
chshersh, line 526: You can replace `x` and `y` variables with `_` as we don't care about their names. Also, you can pattern match on number 42 directly instead of comparing variable `z`

src/Chapter2.hs
```diff
@@ -604,9 +626,10 @@ Implement a function that duplicates each element of the list
 "aabbaacc"
 
 -}
-duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
 
+duplicate :: [a] -> [a]
+duplicate [] = []
+duplicate (x:xs) = concat [[x], [x], duplicate xs]
```
chshersh, line 632: It is more efficient to construct lists using the cons operator directly without introducing intermediate values.
```suggestion
duplicate (x:xs) = x : x duplicate xs
```

src/Chapter2.hs
```diff
@@ -779,14 +808,16 @@ nextInt = add 1
...
 
--- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Ord a => a -> [a] -> [a]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
chshersh, line 820: Eta-reduction award! :trophy:

src/Chapter2.hs
```diff
@@ -841,7 +872,12 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+
+rotate :: Int -> [a] -> [a]
+rotate i l
+	| i < 0 = []
+	| i == 0 = l
+	| otherwise = rotate (i - 1) (take (length l) $ tail (cycle l))
```
chshersh, line 880: Your implementation rotates a list by a single element N times. This is not so efficient. Taking elements from the end of the list and calculating list length is costly.

src/Chapter2.hs
```diff
@@ -857,8 +893,10 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
 
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = (rewind xs) ++ [x]
```
chshersh, line 899: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
Well done on solving our challenges and the Advanced Task :medal_sports: 

As usual, provided some feedback :slightly_smiling_face:

vrom911:

PR #98, opened 2020-10-04
update README

ScientificX:
spelling error


vrom911 (changes requested):
We actually meant [eta-reduction](https://wiki.haskell.org/Eta_conversion)

ScientificX:
Ok. My bad then. Thanks for the clarification.

PR #103, opened 2020-10-05
chapter 1

FredericEspiau:
### Solutions for Chapter 1

cc @vrom911 @chshersh

Thanks for your time 🙏

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,39 +209,33 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-expected: True :: Bool
-got:      True :: Bool
+True :: Bool
+
 >>> :t 'a'
-expected: 'a' :: Char
-got:      'a' :: Char
+'a' :: Char
+
 >>> :t 42
-expected: 42 :: Integer
-got:      42 :: Num p => p (didn't expect that !)
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-expected: (True, 'x') :: (Bool, Char)
-got:      (True, 'x') :: (Bool, Char)
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-expected: not :: Bool -> Bool (I already know a bit of Haskell :) )
-got:      not :: Bool -> Bool
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-expected: (&&) :: Bool -> Bool -> Bool
-got:      (&&) :: Bool -> Bool -> Bool
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -416,7 +416,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Num a => a -> a -> a
```

src/Chapter1.hs
```diff
@@ -541,8 +541,14 @@ value after "=" where the condition is true.
...
-
-mid x y z = error "mid: not implemented!"
+mid :: Ord a => a -> a -> a -> a
+mid x y z
+  | x <= y && y <= z = y
+  | x <= z && z <= y = z
+  | y <= x && x <= z = x
+  | y <= z && z <= x = z
+  | z <= x && x <= y = x
+  | otherwise = y
```
chshersh, line 551: You can actually get rid of the first case as well, to return `y` on those two cases at the same time and have fewer lines in your implementation :slightly_smiling_face:

src/Chapter1.hs
```diff
@@ -619,8 +633,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Integer -> Integer
+sumLast2 n = theLastDigit + digitBeforeLast
+  where
+    theLastDigit = mod (abs n) 10
+    digitBeforeLast = (mod (abs n) 100 - theLastDigit) `div` 10
```
chshersh, line 640: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
```

chshersh (approved):
Nice work! And well done with solving the advanced task :+1: 

I left some comments and my feedback, I hope you find it helpful :slightly_smiling_face:

src/Chapter1.hs
FredericEspiau: Thanks for your inputs !

Would 
```hs
squareSum :: Int a => a -> a -> a
```

be equivalent or it means something else ?

chshersh: The the following signature is invalid:

```haskell
squareSum :: Int a => a -> a -> a
```

`Num` is one of standard typeclasses (aka interfaces), we explain them in Chapter Tree. And `Int` is not a typeclass, so you can't specify `Int a` in the type signature, it will be a compiler error.

On a more advanced note, the following type signature is equivalent to `Int -> Int -> Int`:

```haskell
squareSum :: Int ~ a => a -> a -> a
```

Here we use `~` which is a type equality. So you can use the `a` type variable and it will be an alias to `Int` on simple words. But we don't cover this course, such tricks are needed only in very advanced cases or when trying to achieve something non-trivial, no need to worry about that for now :slightly_smiling_face:

vrom911:

PR #105, opened 2020-10-05
Solutions to Chapter 2, 3, and 4

guojing0:
Hi @vrom911 and @chshersh,

Thank you for this interesting project! Although I knew some Haskell before this, I really get used to some important points until now after really thinking through and working out these problems.

I hope I can give back to the Haskell community in the near future.

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -344,6 +344,11 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+data Book = Book {
+        price :: Double,
+        review :: Bool
+    } deriving (Show)
+
 {- |
 =⚔️= Task 2
 
@@ -373,6 +378,21 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+
+data Knight = Knight {
+    health :: Int,
+    attck :: Int,
+    gold :: Int
+} deriving Show
+
+data Monster = Monster {
+    health' :: Int,
+    attck' :: Int,
+    gold' :: Int
+} deriving Show
+
+-- fight :: Knight -> Monster -> Int
+
 
 {- |
 =🛡= Sum types
@@ -563,13 +583,16 @@ introducing extra newtypes.
 data Player = Player
     { playerHealth    :: Int
     , playerArmor     :: Int
-    , playerAttack    :: Int
+    , playerAttack    :: Attack
     , playerDexterity :: Int
-    , playerStrength  :: Int
+    , playerStrength  :: Strength
     }
 
-calculatePlayerDamage :: Int -> Int -> Int
-calculatePlayerDamage attack strength = attack + strength
+newtype Attack = Attack Int
+newtype Strength = Strength Int
+
+calculatePlayerDamage :: Attack -> Strength -> Int
+calculatePlayerDamage (Attack attack) (Strength strength) = attack + strength
 
 calculatePlayerDefense :: Int -> Int -> Int
 calculatePlayerDefense armor dexterity = armor * dexterity
@@ -910,6 +933,20 @@ Implement instances of "Append" for the following types:
 class Append a where
     append :: a -> a -> a
 
+newtype Gold = Gold Int deriving Show
+
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append (Gold x) (Gold y) = Gold (x + y)
+
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append x y = x ++ y
+
+instance Append a => Append (Maybe a) where
+    append :: Maybe a -> Maybe a -> Maybe a
+    append (Just x) (Just y) = Just (append x y)
+    append _ _ = Nothing
 
 {-
 =🛡= Standard Typeclasses and Deriving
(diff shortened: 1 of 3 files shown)


chshersh:
Hi @guojing0! It means a lot to us that you find our course helpful :blush: 

As you already know Haskell, you might not a review from us. But let us know if you still need :slightly_smiling_face: 
Also, I see that you solved three chapters in a single PR. To get a Hacktoberfest contribution, you need to open separate PRs per chapter total 4 PRs. I've added a label to this one, but let us know if you want to split your work into multiple PRs :slightly_smiling_face:

guojing0:
@chshersh Thank you for the suggestion, but I think one should be fine. If you have time, reviewing only solutions to chapter 3 and 4 would be really helpful, since I don't think I really know them so much.

guojing0:
@chshersh Thank you for the suggestion, but I think one should be fine. If you have time, reviewing only solutions to chapter 3 and 4 would be really helpful, since I don't think I really know them so much.

src/Chapter3.hs
```diff
@@ -374,6 +379,21 @@ after the fight. The battle has the following possible outcomes:
...
+    gold :: Int
+} deriving Show
+
+data Monster = Monster {
+    health' :: Int,
+    attck' :: Int,
+    gold' :: Int
+} deriving Show
+
+-- fight :: Knight -> Monster -> Int
```
chshersh, line 394: Do you have any troubles implementing this function? I think, the easiest way to implement it is to use fields as getters and compare a few values.

src/Chapter3.hs
```diff
@@ -563,13 +583,16 @@ introducing extra newtypes.
 data Player = Player
     { playerHealth    :: Int
     , playerArmor     :: Int
-    , playerAttack    :: Int
+    , playerAttack    :: Attack
     , playerDexterity :: Int
-    , playerStrength  :: Int
+    , playerStrength  :: Strength
```
chshersh, line 588: Ideally, all these 5 fields should be separate newtypes. Newtypes add semantic meaning to fields. `Int` is just `Int` and it can mean anything. But newtypes add additional context.

src/Chapter3.hs
```diff
@@ -910,6 +933,20 @@ Implement instances of "Append" for the following types:
...
 
+newtype Gold = Gold Int deriving Show
+
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append (Gold x) (Gold y) = Gold (x + y)
+
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append x y = x ++ y
```
chshersh, line 944: You can eta-reduce the implementation of `append` for brevity. So you can see how easy it is to create instances :slightly_smiling_face: 
```suggestion
    append = (++)
```

src/Chapter3.hs
```diff
@@ -910,6 +933,20 @@ Implement instances of "Append" for the following types:
...
+
+instance Append a => Append (Maybe a) where
+    append :: Maybe a -> Maybe a -> Maybe a
+    append (Just x) (Just y) = Just (append x y)
+    append _ _ = Nothing
```
chshersh, line 949: I don't think this implementation is valid :thinking: Though, even it makes some sense and nothing in the task description says about the implementation details, but it would be more useful, if append of `Just` and `Nothing` returned `Just`

src/Chapter3.hs
```diff
@@ -971,6 +1008,20 @@ implement the following functions:
...
+isWeekend :: Week -> Bool
+isWeekend day = day == Sat || day == Sun
+
+nextDay :: Week -> Week
+nextDay day
+    | day == maxBound = minBound
+    | otherwise = succ day
+
+daysToParty :: Week -> Int
+daysToParty day = abs (fromEnum day - fromEnum Fri)
```
chshersh, line 1023: Unfortunately, time doesn't work this way, and we can't travel back in time disappointed
For example, if today is Saturday, there are 6 days to the next Friday, not 1 (as in your current solution).

src/Chapter4.hs
```diff
@@ -268,6 +282,13 @@ instance Functor Maybe where
 @
 -}
 
+data Maybe' a = Just' a | Nothing' deriving Show
+
+instance Functor Maybe' where
+    fmap :: (a -> b) -> Maybe' a -> Maybe' b
+    fmap f (Just' a) = Just' (f a)
+    fmap _ Nothing' = Nothing'
```
chshersh, line 290: Just curious, why did you implement your own version of `Maybe` and an instance for it? For practice purposes? :slightly_smiling_face:

src/Chapter4.hs
```diff
@@ -472,10 +499,12 @@ Implement the Applicative instance for our 'Secret' data type from before.
...
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    Trap e <*> _ = Trap e
+    Reward _ <*> Trap b = Trap b
+    Reward a <*> Reward b = fmap a (Reward b)
```
chshersh, line 507: You can actually reduce these two cases to a single one. I see that you use `fmap`, and `fmap` already patterns matches on `Secret` and applies a function to it. So you can reuse `Functor` instance for `Secret` here

src/Chapter4.hs
```diff
@@ -611,6 +656,13 @@ Implement the 'Monad' instance for our lists.
   maybe a few) to flatten lists of lists to a single list.
 -}
 
+flattenLists :: List (List a) -> List a
+flattenLists Empty = Empty
+flattenLists (Cons x xs) = consHelper x (flattenLists xs)
+
+instance Monad List where
+    (>>=) :: List a -> (a -> List b) -> List b
+    xs >>= f = flattenLists (fmap f xs)
```
chshersh, line 665: Nice and idiomatic implementation!

chshersh (approved):
@guojing0 Great work on solving the tasks :+1: 
And kudos for cracking advanced tasks! I left some comments in my review. Hope you find them helpful :slightly_smiling_face: 

I've added a label and approved the PR, so it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to close it after fixes or even straightaway :slightly_smiling_face:

PR #106, opened 2020-10-05
Chapter 1

adriahadi:
### Solutions for Chapter 1

cc @vrom911 @chshersh

Thank you.

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -204,31 +204,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -503,7 +504,11 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y =
+  if abs x > (abs y) then
+    y
+  else
+    x
```
chshersh, line 511: `()` around second `abs` are redundant. And you also can write it in a single line for brevity. But that's fine to leave as it is :slightly_smiling_face:

src/Chapter1.hs
```diff
@@ -537,7 +542,12 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | (x > y && z > x) || (x > z && y > x) = x
+  | (y > x && z > y) || (y > z && x > y) = y
+  | (z > x && y > z) || (z > y && x > z) = z
+  | otherwise = x
```
chshersh, line 550: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

Because of that, you will need to add another guard – `| otherwise = x`, to tell the compiler, that your pattern matching is exhaustive 🙂 But as you can notice, you can just collapse the last two guards and return `z` in the last option.

src/Chapter1.hs
```diff
@@ -615,7 +632,17 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
+sumLast2 n
+  | n < 0 = sumLast2 (abs n)
+  | n <= 10 = mod n 10
+  | otherwise =
+      let
+        last2 = mod (abs n) 100
+        first = div last2 10
+        second = mod last2 10
+      in
+        first + second
```
chshersh, line 645: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
```

chshersh (approved):
Great work on solving our tasks!
And well done on cracking the advanced task :medal_sports: 

I left some comments in my review. I hope you find them helpful :slightly_smiling_face:

chshersh:

PR #107, opened 2020-10-05
Chapter4

jamesthesnake:
### Solutions for Chapter 4

cc @vrom911 @chshersh


src/Chapter4.hs
```diff
@@ -114,25 +114,26 @@ As always, try to guess the output first! And don't forget to insert
...
 >>> :k Trinity
-
+Trinity :: *->*->*->*
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
-
+IntBox :: (*->*)->*
 -}
 
+let go =3
```
chshersh, line 136: Hmm, this looks like an accidental error :thinking:

src/Chapter4.hs
```diff
@@ -114,25 +114,26 @@ As always, try to guess the output first! And don't forget to insert
...
+[Int] :: *
 >>> :k []
-
+[] :: *->*
 >>> :k (->)
-
+(->) :: *->*->*
 >>> :k Either
-
+Either :: *->*->*
```
chshersh, line 127: I see that spaces are removed between `*` and `->` here. To pass doctests, you need to insert the line exactly as GHCi displays, because doctest compare with the GHCi output

chshersh (commented):
Left some comments, but as I see, this is still work in progress :slightly_smiling_face:

vrom911:

PR #110, opened 2020-10-05
mullakhmeov / Chapter #3

mullakhmetov:
### Solutions for Chapter 3

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -381,6 +381,21 @@ after the fight. The battle has the following possible outcomes:
 
 -}
 
+data Unit = MkUnit
+    { unitHealth :: Int
+    , unitAttack :: Int
+    , unitGold   :: Int
+    } deriving (Show)
+
+type Knight = Unit
+type Monster = Unit
+
+fight :: Knight -> Monster -> Int
+fight k m
+  | unitHealth m <= unitAttack k = unitGold m + unitGold k
+  | unitHealth k <= unitAttack m = (-1)
+  | otherwise unitGold k
+
 {- |
 =🛡= Sum types
 
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -481,6 +481,15 @@ and provide more flexibility when working with data types.
 Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
+data BreakfastType = American | Dutch | English | French
+data Meal
+  = Breakfast BreakfastType
+  | Lunch
+  | Elevenses 
+  | Lunch
+  | Tea
+  | Supper
+  | Dinner
 
 {- |
 =⚔️= Task 4
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -830,6 +830,18 @@ parametrise data types in places where values can be of any general type.
   maybe-treasure ;)
 -}
 
+data Dragon = Dragon { dragonPower :: power }
+
+data TreasureChest x = TreasureChest
+    { treasureChestGold :: Int
+    , treasureChestLoot :: x
+    }
+
+data Lair = Lair
+    { lairDragon   :: Dragon
+    , lairTreasure :: Maybe TreasureChest
+    }
+
 {-
 =🛡= Typeclasses
 
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -996,9 +996,24 @@ Implement instances of "Append" for the following types:
   ✧ *(Challenge): "Maybe" where append is appending of values inside "Just" constructors
 
 -}
+newtype Gold = MkGold Int
+
 class Append a where
     append :: a -> a -> a
 
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append (MkGold g1) (MkGold g2) = MkGold (g1 + g2)
+
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append l1 l2 = l1 ++ l2
+
+instance (Append a) => Append (Maybe a) where
+    append :: Maybe a -> Maybe a -> Maybe a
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -374,6 +381,21 @@ after the fight. The battle has the following possible outcomes:
...
+data Unit = MkUnit
+    { unitHealth :: Int
+    , unitAttack :: Int
+    , unitGold   :: Int
+    } deriving (Show)
+
+type Knight = Unit
+type Monster = Unit
+
+fight :: Knight -> Monster -> Int
```
vrom911, line 393: Nice usage of type aliases!

```haskell
> fight knight monster
> fight monster knight
```

```haskell
newtype Monster = Monster Unit
newtype Knight = Knight Unit
```

src/Chapter3.hs
```diff
@@ -459,6 +481,15 @@ and provide more flexibility when working with data types.
...
 -}
+data BreakfastType = American | Dutch | English | French
+data Meal
+    = Breakfast BreakfastType
+    | Brunch
+    | Elevenses
+    | Lunch
+    | Tea
+    | Supper
+    | Dinner
```
vrom911, line 492: Wow, international meals 🇺🇸 🇳🇱 🇬🇧 🇫🇷

src/Chapter3.hs
```diff
@@ -480,6 +511,51 @@ After defining the city, implement the following functions:
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
 
+data City = City
+    { cityCastle  :: Maybe Castle
+    , cityWall    :: Maybe Wall
```
vrom911, line 516: While this is correct, however, this allows having unrepresentable states.
For example, nothing stops us from having `Just Wall` and `Nothing :: Maybe Castle`.

You can actually avoid that by using more flexible structures that precisely would describe all possible states in types and won't let us to create invalid ones.

For example, here, instead of having `Wall` and `Castle` separately, you can think on the may to create a data type, that would represent All: none cattle, castle and castle with walls 🙂

src/Chapter3.hs
```diff
@@ -480,6 +511,51 @@ After defining the city, implement the following functions:
...
+data Shrine = Church | Library deriving (Show)
+
+data House = House
+    { houseRoom :: Int
+    } deriving (Show)
+
+mkHouse :: Int -> Maybe House
+mkHouse rooms
+    | rooms > 4 = Nothing
+    | otherwise = Just (House rooms)
```
vrom911, line 538: ~Interesting approach. It is usually called "smart constructor". To avoid building houses with more than 4 people you need to restrict usage of the Constructor and allow to create `House`s only with your `mkHouse` function.

src/Chapter3.hs
```diff
@@ -480,6 +511,51 @@ After defining the city, implement the following functions:
...
+buildHouse city rooms = case mkHouse rooms of
+    Just house -> city { cityHouses = house : cityHouses city }
+    Nothing    -> city
+
+buildWalls :: City -> City
+buildWalls city = case cityCastle city of
+    Nothing -> city
+    Just _  ->
+        if sum (map houseRoom (cityHouses city)) >= 10
+        then city { cityWall = Just (Wall 42) }
```
vrom911, line 556: Nice! 
As you keep a number of `Wall`s, maybe it makes sense to add them together, when you already have some? 🙂

src/Chapter3.hs
```diff
@@ -560,22 +636,30 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (MkArmor armor) (MkDexterity dexterity) = MkDefense (armor * dexterity)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (MkDamage damage) (MkDefense defense) (MkHealth health) = MkHealth (health + defense - damage)
```
vrom911, line 662: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -907,9 +1005,24 @@ Implement instances of "Append" for the following types:
...
 class Append a where
     append :: a -> a -> a
 
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append (MkGold g1) (MkGold g2) = MkGold (g1 + g2)
+
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append l1 l2 = l1 ++ l2
```
vrom911, line 1019: Yeap, that is right!
You can eta-reduce this function. As you can see, `append` and `(++)` are actually the same functions, as they return the same output on any inputs 🙂

src/Chapter3.hs
```diff
@@ -970,6 +1083,31 @@ implement the following functions:
...
+nextDay :: Weekday -> Weekday
+nextDay weekday =
+    if weekday == maxBound
+    then minBound
+    else succ weekday
+
+daysToParty :: Weekday -> Int
+daysToParty weekday
+    | weekday <= Friday = fromEnum Friday - fromEnum weekday
+    | otherwise = fromEnum Sunday - (fromEnum weekday - fromEnum Friday) + 1
```
vrom911, line 1110: Great catch on the corner case! 👁️ 

You could use the fact that we know the exact number of days, that means, that we could just take the subtraction by `mod .. 7` 🥳

vrom911 (approved):
Wonderful work with types and typeclasses 💯 🔥 
Congratulations 3️⃣ 🎉

vrom911:

PR #112, opened 2020-10-05
Completed first chapter!

RajPathare:
### Solutions for Chapter 1

cc @vrom911 @chshersh
Thank you for creating this repo!

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -478,7 +478,10 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
```
chshersh, line 483: Your implementation is almost correct :ok: 

src/Chapter1.hs
```diff
@@ -541,8 +544,8 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z = (x + y + z) - (min (min x y) z) - (max (max x y) z)
```
chshersh, line 548: Very interesting solution! 💡 

src/Chapter1.hs
```diff
@@ -556,7 +559,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+    | c `elem` "aeiouAEIOU" = True
```
chshersh, line 564: Nice one! 👍🏼 

```haskell
isVowel c = c `elem` "aeiouAEIOU"
```

src/Chapter1.hs
```diff
@@ -620,8 +626,11 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n = 
+    let lastNum = mod n 10
+        secondLastNum = div (mod n 100) 10
+        in (lastNum + secondLastNum)
```
chshersh, line 633: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
```

src/Chapter1.hs
```diff
@@ -641,8 +650,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit n =
```
chshersh, line 653: You are encouraged to add type signatures to all top-level functions :top:

src/Chapter1.hs
```diff
@@ -641,8 +650,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit n = 
+    if (n>=10)
+      then firstDigit (div n 10)
+      else n
```
chshersh, line 656: Your implementation is almost correct, but it returns wrong results on negative numbers. Probably you can patch it to take this corner case into consideration :slightly_smiling_face:

chshersh (approved):
Nice solutions! And well done with solving the Advanced Task :tada: 

I left some comments and some suggestions on how to improve the code. I hope you find them helpful :slightly_smiling_face:

src/Chapter1.hs
RajPathare: I almost forgot considering negative numbers! @chshersh 
Thank you for reviewing my code! :D

chshersh:

RajPathare:
> The solutions look good  I'm going to close this PR, because we can't merge the solutions to our repo  But don't worry, it still counts towards Hacktoberfest contribution   Feel free to open PRs with solutions to other chapters!

Thanks! @chshersh

PR #113, opened 2020-10-05
My chapter 2 solutions

premkiran7:
### Solutions for Chapter 2

Would like to know if it is possible to set up github actions to run when I make commits to a branch other than master. Also, I am not sure if I understood task 10 pairMul part.

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
@@ -186,31 +186,31 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
-
+[10,2,3,1,5]
 >>> [] ++ [1, 4]  -- [] is an empty list
-
+[1,4]
 >>> 3 : [1, 2]
-
+[3,1,2]
 >>> 4 : 2 : [5, 10]  -- prepend multiple elements
-
+[4,2,5,10]
 >>> [1 .. 10]  -- list ranges
-
+[1,2,3,4,5,6,7,8,9,10]
 >>> [10 .. 1]
-
+[]
 >>> [10, 9 .. 1]  -- backwards list with explicit step
-
+[10,9,8,7,6,5,4,3,2,1]
 >>> length [4, 10, 5]  -- list length
-
+3
 >>> replicate 5 True
-
+[True,True,True,True,True]
 >>> take 5 "Hello, World!"
-
+"Hello"
 >>> drop 5 "Hello, World!"
-
+", World!"
 >>> zip "abc" [1, 2, 3]  -- convert two lists to a single list of pairs
-
+[('a',1),('b',2),('c',3)]
 >>> words "Hello   Haskell     World!"  -- split the string into the list of words
-
+["Hello","Haskell","World!"]
 
 
 👩‍🔬 Haskell has a lot of syntax sugar. In the case with lists, any
@@ -336,7 +336,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList start end list = if start > end || start < 0 || end < 0
+                            then []
+                            else take (end - start + 1) (drop start list)
 
 {- |
 =⚔️= Task 4
@@ -348,8 +350,8 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf l = subList 0 ((div (length l) 2) - 1) l
 
 
 {- |
@@ -500,7 +502,8 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird :: [Int] -> Bool
+isThird list = if (take 1 (reverse (take 3 list))) == [42] then True else False
 
 
 {- |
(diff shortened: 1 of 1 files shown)


vrom911:
> Would like to know if it is possible to set up github actions to run when I make commits to a branch other than master.

You can enable GitHub Actions on your fork, so you would be able to have it work in there 🙂 
For that, you could go to https://github.com/premkiran7/learn4haskell/actions
And press enable.

src/Chapter2.hs
```diff
@@ -348,8 +350,8 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf l = subList 0 ((div (length l) 2) - 1) l
```
vrom911, line 354: Nice reusage of the `subList` function! 👍🏼 

Note that `sublist 0` is the same as just `take` 🙂

src/Chapter2.hs
```diff
@@ -500,7 +502,8 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird :: [Int] -> Bool
+isThird list = if (take 1 (reverse (take 3 list))) == [42] then True else False
```
vrom911, line 506: Wow that is a creative implementation 🎨 

Nothe that you don;t need to put in into `if-then-else` as it already returns `Bool`.
```suggestion
isThird42 list = take 1 (reverse (take 3 list)) == [42] 
```

And, btw, tests are failing because they expect this function to be named `isThird42` 🙂

vrom911: On the other hand, here you could just use the pattern matching. It would be less expensive 🙂 

```haskell
isThird42 (_ : _ : 42 : _ ) = True
isThird42 _ = False
```

src/Chapter2.hs
```diff
@@ -605,7 +608,8 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate [] = []
+duplicate (x:xs) = [x] ++ [x] ++ duplicate xs
```
vrom911, line 612: It is more efficient to construct lists using the cons operator directly without introducing intermediate values.
```suggestion
duplicate (x:xs) = x : x duplicate xs
```

src/Chapter2.hs
```diff
@@ -620,7 +624,14 @@ Write a function that takes elements of a list only on even positions.
...
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven list = selector 0 [] list
+  where
+    selector :: Int -> [a] -> [a] -> [a]
+    selector index evenList [] = evenList
+    selector index evenList (x:xs) = if (mod index 2 == 0) 
+                              then selector (index + 1) (evenList ++ [x]) xs
+                              else selector (index + 1) evenList xs
```
vrom911, line 634: Creative again! 💡 

You actually could implement this one with just pattern matching. You would need to check a few cases of the list for that 😉

src/Chapter2.hs
```diff
@@ -727,7 +738,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concat (map (\x -> replicate x x) l)
```
vrom911, line 741: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -740,7 +751,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains num l = filter (\x -> elem num x) l
```
vrom911, line 755: 👌🏼 
You Could use eta-reduction both in the function itself and in the lambda function 🙂

src/Chapter2.hs
```diff
@@ -841,7 +855,11 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate num l = if num < 0 then [] else subList (getStartIndex num (length l)) (getEndIndex num (length l)) (cycle l)
+                where
+                  getStartIndex num lsize = mod num lsize
+                  getEndIndex num lsize = (getStartIndex num lsize) + lsize - 1
```
vrom911, line 862: `num` is in scope from the `rotate` function argument, so you are actually shadowing that one. Instead, you could amend passing it to this function at all, and use it from the scope. The same applies to the `length l` 🙂

src/Chapter2.hs
```diff
@@ -857,7 +875,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [Int] -> [Int]
+rewind [] = []
+rewind (x:xs) = rewind(xs) ++ [x]
```
vrom911, line 880: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

src/Chapter2.hs
```diff
@@ -783,9 +795,11 @@ divideTenBy :: Int -> Int
 divideTenBy x = div 10 x
```
vrom911, line 795: When eta-reducing, you need to remove arguments from both sides of the `=` sign. So the correct solution is:

```haskell
divideTenBy = div 10
```

See how it compares with the original code:

```diff
-divideTenBy x = div 10 x
+divideTenBy = div 10
```

The motivation behind eta-reduction if that if you have two functions `foo` and `bar` of the same type, and they return the same results for any arguments (so `foo x` always equal to `bar x` for any `x`). In such situations, we say that functions `foo` and `bar` are equal, and if we want to define `foo` in terms of `bar`, we can write `foo = bar` instead of a longer `foo x = bar x` (which means that `foo` returns the same result as `bar` for any `x`).

The same goes for the `pairMul`. The only difference, that you can eta-reduce both arguments in there 🙂 

Does it make sense?

vrom911 (approved):
Amazing work! Very interesting solutions 🥇

premkiran7:
@vrom911 @chshersh Would like to know if there is any code formatter plugin that I can use in VSCode for Haskell code like [prettier](https://prettier.io/). My aim is to avoid the minor formatting issues like redundant brackets, trailing spaces.

premkiran7:
> > Would like to know if it is possible to set up github actions to run when I make commits to a branch other than master.
> 
> You can enable GitHub Actions on your fork, so you would be able to have it work in there 
> For that, you could go to https://github.com/premkiran7/learn4haskell/actions
> And press enable.

Added a [workflow file](https://github.com/premkiran7/learn4haskell/blob/master/.github/workflows/haskell.yml). My intention was to run this workflow whenever I make commits to a branch name starting with **my-chapter** in my fork

I am confused if this is the correct way to set it up. Thoughts?

chshersh:
@premkiran7 Regarding the formatter. Unfortunately, there is no standard formatter in the Haskell ecosystem at the moment. In @kowainik we follow this style guide:

* https://kowainik.github.io/posts/2019-02-06-style-guide

But there is no tool that automatically formats the code according to our style guide, unfortunately :disappointed: And, as you can imagine, writing such a tool would require a lot of time and effort...

Haskell has several code formatters but all of them format differently and have different trade-offs. The one I enjoy the most is `stylish-haskell`, but it doesn't format every letter of your code. It mostly focuses on import and alignment of `=` in function bodies.

* https://github.com/jaspervdj/stylish-haskell

You can also look into `brittany` or `flosskell` as alternative Haskell formatters.

* https://github.com/lspitzner/brittany/
* https://github.com/ennocramer/floskell

I think Haskell plugin for VSCode should support most main code formatters. And they do a pretty good job in removing trailing spaces. But they don't remove redundant brackets. For this you need to use some combination of HLint and `apply-refact`, but this seems like rabbit hole you may want to follow only if you have time... I'm not using VSCode personally, so can't share ready config yet :disappointed:

PR #117, opened 2020-10-05
solution for Chapter 1

Wumanizer:
### Solutions for Chapter 1
cc @vrom911 @chshersh

Hi Veronika, hi Dmitrii,

this is my first pull request ever ;-) I hope everything is fine.
Thx in advance for your review and kudos for this project!

Best regards
Marc

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -477,8 +477,8 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = (abs n) `mod` 10
+lastDigit :: Int -> Int
+lastDigit n = abs n `mod` 10
 
 
 {- |
@@ -508,7 +508,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = if (abs x) < (abs y) 
+closestToZero x y = if abs x < abs y
                     then x
                     else y
 
@@ -633,9 +633,9 @@ specifying complex expressions.
 -}
 sumLast2 :: Int -> Int
 sumLast2 n = let m = abs n
-                 lastDigit = m `mod` 10
-                 secondLastDigit = (m `div` 10) `mod` 10
-              in lastDigit + secondLastDigit
+                 lastD = m `mod` 10
+                 secondLastD = (m `div` 10) `mod` 10
+              in lastD + secondLastD
 
 
 {- |
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -477,8 +477,8 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = abs n `mod` 10
```
vrom911, line 481: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1:

src/Chapter1.hs
```diff
@@ -508,7 +508,9 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if abs x < abs y
+                    then x
+                    else y
```
vrom911, line 513: Perfect 👌🏼 

Usually, we try to not make the indentation to depend on the previous lines, as you see, if you decide to rename the function, you will need to realign both `then` and `else` lines as well.

I would personally write like this:

```haskell
closestToZero x y = 
    if abs x < abs y
    then x
    else y
```

But this is just a note! This doesn't affect the correctness of your solution at all! 🌟

src/Chapter1.hs
```diff
@@ -541,8 +543,11 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | (y <= x && x <= z) || (z <= x && x <= y) = x
+    | (x <= y && y <= z) || (z <= y && y <= x) = y
+    | otherwise                                = z
```
vrom911, line 550: Nice solution with the minimal amount of guards used. Great catch! 🥇

src/Chapter1.hs
```diff
@@ -619,8 +631,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = let m = abs n
+                 lastD = m `mod` 10
+                 secondLastD = (m `div` 10) `mod` 10
```
vrom911, line 637: That is awesome!

You can see that you have both:

```haskell
m `mod` 10
```
and 
```haskell
m `div` 10
```

The standard library has the [`divMod`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Prelude.html#v:divMod) function, that actually combines inside both `div` and `mod`. And this is exactly what you use!

```haskell
(d, lastD) = divMod m 10
```

vrom911 (approved):
Fantastic! Congrats with finishing Chapter One 🎉 

Your solutions are awesome! And you nailed the advanced task brilliantly as well 👏🏼 

And awesome work on the first PR ever! You did very neat work 😮 Happy that you decided to work on our repo 😊

chshersh:
@Wumanizer The solutions look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to open PRs with solutions to other chapters!

PR #118, opened 2020-10-06
My chapter2 solution

adriahadi:
### Solutions for Chapter 2

cc @vrom911 @chshersh

Thank you for your help.

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -47,7 +47,7 @@ Before we start diving into the FP concepts, let's cover an important
 process-development detail.
 
 Without any additional effort, in Haskell you have access to all
-functions defined in the special module called "Prelude". This module
+functions defined in the special module called "Prelude". This module3
 is always imported by default and contains all primitive data types
 and useful functions to work with them. However, sometimes you need to
 import other modules even from other libraries.
@@ -136,34 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -327,8 +348,13 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
-
+subList x y alist 
+  | length alist == 0 = []
```
vrom911, line 352: That is the correct idea!

In Haskell, unfortunately, it is not cheap to get the length of the list. You need to get through each element of the list to get that. So it is not efficient to make this check in here.
However, you can compare this list to `[]` empty list. This operation is lazy, so works instantly:

```haskell
alist == [] = [] 
```

Alternatively, there is `null` function that does precisely the same:

```haskell
null alist = []

File: src/Chapter2.hs [hunk 2]
```diff
@@ -611,7 +644,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven (x : []) = [x]
+takeEven (x : _: xs) = x : takeEven xs
```
vrom911, line 650: Great pattern matching skills 🧑‍🎨

src/Chapter2.hs
```diff
@@ -718,7 +754,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate aList = concat (map (\x -> replicate x x) aList)
```
vrom911, line 757: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -731,7 +767,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains n aList = filter (elem n) aList
```
vrom911, line 771: Now, after you've mastered eta-reduction, you can apply such a technique to other functions as well  🙂 

src/Chapter2.hs
```diff
@@ -771,13 +808,17 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
-listElementsLessThan x l = filter (< x) l
+--listElementsLessThan x l= filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x) 
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+--pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 821: Eta-reduction award  🏆

src/Chapter2.hs
```diff
@@ -832,7 +873,16 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate n aList
+  | n < 0 || length aList == 0 = []
```
vrom911, line 878: Here the same note about length.

Also, I see that you already have it as the variable in `where`, so you can reuse it in here 🙂

src/Chapter2.hs
```diff
@@ -832,7 +873,16 @@ list.
...
+  | otherwise =
+      take listLength (drop n (take repeatCycle (cycle aList)))
+      where
+        listLength = length aList
+        repeatCycle = listLength * (n + 1)
```
vrom911, line 883: I don't quite understand this moment, could you please elaborate on your algorithm?

src/Chapter2.hs
```diff
@@ -848,7 +898,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
vrom911, line 903: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Interesting solutions for this Chapter! Great one 🥇

vrom911:

PR #120, opened 2020-10-06
added solution for Chapter 2

Wumanizer:

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -374,7 +374,7 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 firstHalf :: [a] -> [a]
-firstHalf l = let mid = (length l) `div` 2 - 1
+firstHalf l = let mid = length l `div` 2 - 1
                in subList 0 mid l
 
 
@@ -759,7 +759,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate = concat . map (\x -> replicate x x)
+smartReplicate = concatMap (\x -> replicate x x)
 
 {- |
 =⚔️= Task 9
@@ -773,7 +773,7 @@ the list with only those lists that contain a passed element.
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
 contains :: Int -> [[Int]] -> [[Int]]
-contains n = filter (\l -> n `elem` l)
+contains n = filter (n `elem`)
 
 
 {- |
@@ -881,7 +881,7 @@ rotate n l = if n < 0
              else let len = length l
                       lc = cycle l
                       m = n `mod` len
-                   in take len $ drop m lc 
+                   in take len $ drop m lc
 
 {- |
 =💣= Task 12*
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,42 +136,52 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
+[True, False] :: [Bool]
 
 
 String is a list of characters:
 >>> :t "some string"
+"some string" :: [Char]
 
 
 Empty list:
 >>> :t []
+[] :: [a]
 
 
 Append two lists:
 >>> :t (++)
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +359,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList i j xs = if 0 <= i && i <= j
+                 then take (j - i + 1) $ drop i xs
+                 else []
```
chshersh, line 364: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding a separate case to your check :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -348,8 +373,9 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf l = let mid = length l `div` 2 - 1
+               in subList 0 mid l
```
chshersh, line 378: Nice usage of variables here!

src/Chapter2.hs
```diff
@@ -620,7 +649,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven [x] = [x]
+takeEven (x:_:xs) = x : takeEven xs
```
chshersh, line 655: Pattern-matching award! :trophy:

src/Chapter2.hs
```diff
@@ -740,7 +772,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains n = filter (n `elem`)
```
chshersh, line 776: You don't need to use infix notation here :slightly_smiling_face: 
```suggestion
contains n = filter (elem n)
```

src/Chapter2.hs
```diff
@@ -727,7 +759,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate = concatMap (\x -> replicate x x)
```
chshersh, line 762: Wow, a very nice finding of a standard function!

src/Chapter2.hs
```diff
@@ -841,7 +875,13 @@ list.
...
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate n l = if n < 0
+             then []
+             else let len = length l
+                      lc = cycle l
+                      m = n `mod` len
+                   in take len $ drop m lc
```
chshersh, line 884: Nice implementation! Note, that `cycle` fails on empty lists, and `mod` fails with "division by zero" error when `length` is zero. But due to Haskell laziness and the implementation of `take`, you don't see such errors. You can write simple code, and it works on all cases automatically :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -857,7 +897,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
chshersh, line 902: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):

chshersh:

PR #126, opened 2020-10-06
[+] Chapter 3

juev:
### Solutions for Chapter 3

In progress. I have some comments.

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -40,6 +40,8 @@ Now, if you are ready, bring it on!
 
 module Chapter2 where
 
+import Data.List
+
 {-
 =🛡= Imports
 
@@ -136,34 +138,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
@@ -177,31 +188,43 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
+[10,2,3,1,5]
 
 >>> [] ++ [1, 4]  -- [] is an empty list
+[1,4]
 
 >>> 3 : [1, 2]
+[3,1,2]
 
 >>> 4 : 2 : [5, 10]  -- prepend multiple elements
+[4,2,5,10]
 
 >>> [1 .. 10]  -- list ranges
+[1,2,3,4,5,6,7,8,9,10]
 
 >>> [10 .. 1]
+[]
 
 >>> [10, 9 .. 1]  -- backwards list with explicit step
+[10,9,8,7,6,5,4,3,2,1]
 
 >>> length [4, 10, 5]  -- list length
+3
 
 >>> replicate 5 True
+[True,True,True,True,True]
 
 >>> take 5 "Hello, World!"
+"Hello"
 
 >>> drop 5 "Hello, World!"
+", World!"
 
 >>> zip "abc" [1, 2, 3]  -- convert two lists to a single list of pairs
+[('a',1),('b',2),('c',3)]
 
 >>> words "Hello   Haskell     World!"  -- split the string into the list of words
-
+["Hello","Haskell","World!"]
 
 
 👩‍🔬 Haskell has a lot of syntax sugar. In the case with lists, any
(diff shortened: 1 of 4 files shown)


src/Chapter2.hs
```diff
@@ -611,7 +651,15 @@ Write a function that takes elements of a list only on even positions.
...
 -}
-takeEven = error "takeEven: Not implemented!"
+-- takeEven :: [Int] -> [Int]
+-- takeEven [] = []
+-- takeEven (x:xs) = x : takeEven (drop 1 xs)
+
+-- version with pattern matching
+takeEven :: [Int] -> [Int]
+takeEven [] = []
+takeEven (x:[]) = x : []
```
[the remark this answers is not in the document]
juev: It's not true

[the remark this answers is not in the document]
juev: Your wrong

src/Chapter3.hs
```diff
@@ -374,6 +385,34 @@ after the fight. The battle has the following possible outcomes:
...
+fight :: Knight -> Monster -> Int
+fight knight monster
+
+--  we got Health from monster minus attack from knight, if this value less as
+--  zero, we got Gold from monster
+    | monsterHealth monster - knightAttack knight <= 0 =
+        knightGold knight + monsterGold monster
+-- if attack from monster is more then Health knigth we got zero
+    | knightHealth knight - monsterAttack monster <= 0 =
+        (-1)
```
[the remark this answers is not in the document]
juev: Controversial issue

src/Chapter3.hs
```diff
@@ -344,6 +344,17 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+-- In descriptions you provide MkKnight constructor, but in solutions you have
+-- just Book constructor. Why??
```
chshersh, line 348: Constructors can have the same name as the data type. We add the `Mk` prefix to help people understand the difference between types and constructors. Otherwise, it may be very confusing. However, in practice, when using Haskell in real-world, most people don't bother with adding `Mk` (extra two letters to write!) and always name constructors using the same name as the type itself.

src/Chapter3.hs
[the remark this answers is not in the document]
chshersh: Actually, this works without `()` to my surprise as well. Though, feel free to keep `()` for readability :slightly_smiling_face:

src/Chapter3.hs
```diff
@@ -460,6 +499,15 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
...
 -}
 
+data Breakfast
+  = Milk
+  | Eags
+  | Omelette
+  | Bread
+  | Oatmeal
+  | Sandwich
+  | Coffee
```
chshersh, line 509: Nice menu!

src/Chapter3.hs
```diff
@@ -479,6 +527,54 @@ After defining the city, implement the following functions:
...
+   { cityCastle :: Castle
+   , cityBuild  :: Building
+   , cityHouses :: [House]
+   }
+
+-- our castle
+data Castle
+    = None    -- dont have castle, but I cannot see what is None from your Chapter
+    | OnlyCastle -- one castle
+    | CastleWithWalls -- castle with wall
```
chshersh, line 542: The task tells that the castle should have a name. In this case, it can be a `String`. Sorry, if it wasn't clear before. We improved this task wording in the latest version to clarify this moment.
Could you guess, what functions would need patching in case of the type modifications?

src/Chapter3.hs
```diff
@@ -560,22 +656,38 @@ introducing extra newtypes.
...
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) =
+  Defense (armor * dexterity)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage damage) (Defense defense) (Health health) =
+  Health (health + defense - damage)
```
chshersh, line 690: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -560,22 +656,38 @@ introducing extra newtypes.
...
     }
 
-calculatePlayerDamage :: Int -> Int -> Int
-calculatePlayerDamage attack strength = attack + strength
+calculatePlayerDamage :: Attack -> Strength -> Damage
+calculatePlayerDamage (Attack attack) (Strength strength) =
+  Damage (attack + strength)
+
+-- (Attack attack) - its using our types.
+-- but this: Damage (attack + strength) - what is it?
```
chshersh, line 682: Here you are using the `Damage` constructor of the `Damage` type to create a value of type `Damage` (yes, constructor and the type has the same name). You just adding manual wrapping here :slightly_smiling_face: Your function must return value of type `Damage`, but values `attack` and `strength` have type `Int`. That's why you need to wrap `Int` to return `Damage`

src/Chapter3.hs
```diff
@@ -971,6 +1097,30 @@ implement the following functions:
...
+    Sun -> True
+    _ -> False
+
+nextDay :: Weekday -> Weekday
+nextDay wd
+    | wd == maxBound = minBound
+    | otherwise = succ wd
+
+daysToParty :: Weekday -> Int
+daysToParty wd = abs (fromEnum wd - fromEnum Fri)
```
chshersh, line 1122: This would show the shortest distance to `Friday`, however, time doesn't work like this.
For example. if it is `Saturday`, it would tell you that it is 1 day to `Friday`, however, it is 6 days to go 🥳

chshersh (approved):
Hi @juev! I reviewed your current solutions and answered some of your questions. Let me know if I missed anything, or if something is not clear :slightly_smiling_face:

src/Chapter3.hs
juev: I forgot about enums. This data just enum with my values. It's not the same as Nothing from Maybe.
I understood now.

PR #127, opened 2020-10-06
Solutions for Chapter 2

nbelluco:
### Solutions for Chapter 2

Loving this so far, thanks so much for creating this project !

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -353,7 +353,7 @@ Implement a function that returns only the first half of a given list.
 -}
 firstHalf :: [a] -> [a]
 firstHalf l
-    | length l == 0 = []
+    | null l = []
     | otherwise = take (div (length l) 2) l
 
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -841,7 +858,9 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate n l =
+    take (length l) $ drop n $ cycle l
```
chshersh, line 863: Function `rotate` should return empty list on negative numbers, but your implementation will return the list itself

src/Chapter2.hs
```diff
@@ -857,7 +876,10 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind [] = []
+rewind [x] = [x]
+rewind (x:xs) =
+    rewind xs ++ [x]
```

chshersh (approved):

We are glad you are enjoying the course :hugs:

PR #129, opened 2020-10-07
Solutions for Chapter 1

imsekun:
### Solutions for Chapter 1

Fun!

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -478,7 +478,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
```
vrom911, line 482: Your implementation is almost correct :ok: 

src/Chapter1.hs
```diff
@@ -508,7 +509,12 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y =
+    if absX < absY then
+      absX
+    else absY
```
vrom911, line 515: Actually, you need to return the value itself instead of the absolute value. In this case, we lose the sign. But it is easy to fix 👍🏼

src/Chapter1.hs
```diff
@@ -619,8 +637,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+    let tensDigit = div n 10
+        onesDigit = mod n 10
```
vrom911, line 643: One hint to make your solution a bit shorter: the standard library has the [`divMod`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Prelude.html#v:divMod) function, that actually combines inside both `div` and `mod`. And this is exactly what you use!

```haskell
(first, second) = divMod n 10
```

And also here you also would need to consider case with negative numbers 🙂

src/Chapter1.hs
```diff
@@ -541,8 +547,13 @@ value after "=" where the condition is true.
...
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | x < y && y < z = y
+    | z < y && y < x = y
+    | y < x && x < z = x
+    | z < x && x < y = x
+    | otherwise = z
```
vrom911, line 556: Yes, perfect case 💯

src/Chapter1.hs
```diff
@@ -640,8 +661,10 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n =
+    if divN == 0 then n else firstDigit divN
+    where divN = div n 10
```
vrom911, line 667: Almost there!

Here you also would need to think on how to handle negative numbers 🙂

vrom911 (approved):
Great work on Chapter One 🎉 

I added a few comments, some of the tasks need special treatments to support negative number. Other than that, your solutions look wonderful! 👏🏼

imsekun:
@vrom911 Thank you so much for the review! I have to be more mindful about the special cases.

vrom911:

PR #130, opened 2020-10-07
Solutions for Chapter 2

imsekun:
### Solutions for Chapter 2

cc @vrom911 @chshersh


src/Chapter2.hs
```diff
@@ -620,7 +632,14 @@ Write a function that takes elements of a list only on even positions.
...
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven list = go 0 list
+    where go :: Int -> [a] -> [a]
+          go _ [] = []
+          go count (x:xs) 
+             | mod count 2 == 0 = x:go (count + 1) xs
+             | otherwise = go (count + 1) xs
```
chshersh, line 642: Nice idea on the helper functions :+1:

src/Chapter2.hs
```diff
@@ -857,7 +882,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
chshersh, line 887: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
Great work on solving our challenges!

src/Chapter2.hs
imsekun: This one blew my mind. The solution is now so much shorter!

imsekun:
> Great work on solving our challenges!
> And well done with the Advanced Task medal_sports
> 
> I left my review and some comments. Hope you find them helpful slightly_smiling_face

Thanks so much for the review!

chshersh:
@imsekun You are welcome! The solutions look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to open PRs with solutions to other chapters!

P.S. Thanks for fixing the hole in the train! We really need our train working to get the end of Chapter Four... :steam_locomotive:

PR #131, opened 2020-10-07
Finish Chapter 1 in the right place this time

jacobshu:
### Solutions for Chapter 1

Thanks again!

/cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -541,8 +541,12 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | (y < x && x < z) || (z < x && x < y) = x
+  | (x < y && y < z) || (z < y && y < x) = y
+  | otherwise = z
```
vrom911, line 548: Perfect guards with a minimal amount of branches 👌🏼

src/Chapter1.hs
```diff
@@ -556,7 +560,15 @@ True
...
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+  | c == 'a' || c == 'A' = True
+  | c == 'e' || c == 'E' = True
+  | c == 'i' || c == 'I' = True
+  | c == 'o' || c == 'O' = True
+  | c == 'u' || c == 'U' = True
```
vrom911, line 569: Nice catch with the upper cases 👍🏼

src/Chapter1.hs
```diff
@@ -619,8 +631,12 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+  let last2 = mod n 100
+      tens = div last2 10
+      ones = mod last2 10
```
vrom911, line 638: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
(tens, ones) = divMod last2 10
```

src/Chapter1.hs
```diff
@@ -640,8 +656,8 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n = if div n 10 == 0 then n else firstDigit (div n 10)
```
vrom911, line 660: I see, that for a negative number like -9, this implementation will return the negative number itself instead of the last digit. 

But it works nicely on other cases 👍🏼

vrom911 (approved):
Great start! Congrats 1️⃣ Very nice solutions!

chshersh:

PR #133, opened 2020-10-07
Complete `Chapter1`

ufocoder:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -478,7 +478,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Integral a => a -> a
```
chshersh, line 481: Nice work on specifying a polymorphic type! We don't explain them, so `Int -> Int` will work for now. But that's nice :slightly_smiling_face:

src/Chapter1.hs
```diff
@@ -620,7 +636,12 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = firstLast + secondLast
+  where
+    a = abs n
+    firstLast = mod a 10
+    secondLast = mod (div a 10) 10
```
chshersh, line 644: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
```

src/Chapter1.hs
```diff
@@ -641,7 +662,9 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n =
+  if div n 10 == 0 then n else firstDigit (div n 10)
```
chshersh, line 667: I see that CI run was cancelled by timeout. This happened because the implementation of `firstDigit` hangs. You can verify it in GHCi:

```haskell
ghci> div (-1) 10
-1
```

So, your function hangs on negative numbers and you need to handle them separately.

chshersh (approved):

src/Chapter1.hs
ufocoder: Woow, I don't know about this pattern matching way 0_o
```
```

chshersh: Don't worry! We explain pattern matching in Chapter Two :slightly_smiling_face:

PR #134, opened 2020-10-07
Complete `Chapter2`

ufocoder:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -500,7 +504,9 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird42 :: [Int] -> Bool
+isThird42 (_ : _ : x : _ ) = x == 42
```
chshersh, line 508: You don't need to compare `x` with 42 separately, you can pattern match on `42` directly in the pattern.

src/Chapter2.hs
```diff
@@ -620,7 +627,11 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+
+takeEven :: [a] -> [a]
+takeEven (x:_:xs) = x : takeEven xs
+takeEven (x:_) = [x]
+takeEven x = x
```
chshersh, line 634: I would recommend writing these two patterns explicitly to utilize GHC exhaustiveness checker for patterns. The first line is changed for minor aesthetics reasons, but the last one is the important change here.
```suggestion
takeEven [x] = [x]
takeEven [] = []
```

src/Chapter2.hs
```diff
@@ -727,7 +738,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate xs = concat (map (\x -> replicate x x) xs)
```
chshersh, line 741: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -780,13 +792,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Ord a => a -> [a] -> [a]
+listElementsLessThan x = filter (<x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Integer] -> [Integer] -> [Integer]
+pairMul = zipWith (*)
```
chshersh, line 803: Eta-reduction award! :trophy:

src/Chapter2.hs
```diff
@@ -857,7 +875,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind  :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```

chshersh (approved):
Very nice work on Chapter Two!
And kudos for solving the advanced task :+1: 

I left some comments and suggestions, but overall looks fine :slightly_smiling_face:

PR #138, opened 2020-10-07
Finished tasks in Chapter1 !

Perceval62:
### Solutions for Chapter 1

cc @vrom911 @chshersh

This is an amazing course !

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -480,7 +480,7 @@ Implement a function that returns the last digit of a given number.
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit :: Int -> Int
-lastDigit n = mod n 10
+lastDigit n = mod (abs n) 10
 
 
 {- |
@@ -510,7 +510,10 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = 
+  if (abs x) > (abs y)
+    then y
+    else x
 
 
 {- |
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -542,7 +547,19 @@ value after "=" where the condition is true.
...
+  | x < z && x > y = x
+
+  | y < x && y > z = y  -- Conditions for which y is the middle number
+  | y < z && y > x = y
+
+  | z < y && z > x = z  -- Conditions for which X is the middle number
+  | z < x && z > y = z
+
+  | x == y && y == z = x  -- If all equal return any of them
+  | otherwise = y
```
chshersh, line 562: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

It looks like you can move all these 4 cases under a single otherwise and return z in here.

src/Chapter1.hs
```diff
@@ -619,8 +644,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = 
+  let num = mod (abs n) 100
+      (tens, units) = divMod num 10
```
chshersh, line 650: Wow, the first person to discover `divMod` on their own :1st_place_medal:

chshersh (approved):

chshersh:
The solutions look good :+1: Feel free to close this PR right now or after fixes, since we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Don't hesitate to open PRs with solutions to other chapters!

PR #139, opened 2020-10-07
add test chapter 1 task 7

krivtsov:
### Add test case for Chapter 1

cc @vrom911 @chshersh


test/Test/Chapter1.hs
```diff
@@ -48,6 +48,7 @@ chapter1normal = describe "Chapter1Normal" $ do
     describe "Task7: mid" $ do
         it "positives up  " $ mid 10 20 30 `shouldBe` 20
         it "positives down" $ mid 30 20 10 `shouldBe` 20
+        it "positives mix " $ mid 20 30 10 `shouldBe` 20
```
vrom911, line 51: Great catch!
Would you like to also add a similar test case with the negatives?

vrom911 (approved):
Thank you very much! 💯

test/Test/Chapter1.hs
krivtsov: Right! 
Аdd case

PR #142, opened 2020-10-07
resolve chapter1

krivtsov:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -541,9 +542,11 @@ value after "=" where the condition is true.
...
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
-
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | x >= y && x <= z || x < y && x > z = x
+  | y < x && y > z || y > x && y < z = y
+  | otherwise = z
```
chshersh, line 549: Nice solution and correct usage of otherwise!

src/Chapter1.hs
```diff
@@ -619,8 +629,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+  let secondNum = div (mod (abs n) 100) 10
+      lastNum = mod (abs n) 10
+  in secondNum + lastNum
```
chshersh, line 636: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
```

chshersh (approved):

vrom911:

PR #145, opened 2020-10-07
Chapter 2 tasks

pennychase:
### Solutions for Chapter {2}

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -337,10 +337,11 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList start end xs =
-  if start < 0 || end < 0 
-    then [] 
-    else take (end - start + 1) (drop start xs)
+subList start end xs 
+  | end < start     = []
+  | start < 0       = []
+  | end < 0         = []
+  | otherwise       = take (end - start + 1) (drop start xs)
 
 {- |
 =⚔️= Task 4
@@ -800,7 +801,7 @@ listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
 pairMul :: [Int] -> [Int] -> [Int]
-pairMul xs = zipWith (*) xs 
+pairMul = zipWith (*) 
 
 {- |
 =🛡= Lazy evaluation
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +337,10 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList start end xs =
+  if start < 0 || end < 0
```

src/Chapter2.hs
```diff
@@ -620,7 +628,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven []         = []
+takeEven [x]        = [x]
+takeEven (x:_:xs)  = (x:takeEven xs)
```
chshersh, line 634: Pattern-matching award! :trophy:

src/Chapter2.hs
```diff
@@ -740,7 +751,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Eq a => a -> [[a]] -> [[a]]
```
chshersh, line 754: Nice work on specifying polymorphic type here!

src/Chapter2.hs
```diff
@@ -857,8 +875,12 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```

Your second implementation with `foldl` is much more efficient :horse: Nice work on using standard Haskell functions!

chshersh (approved):
Awesome work on Chapter Two!
And well done with the Advanced Task! :1st_place_medal: 

I left some minor comments, but nothing critical :slightly_smiling_face: 

The solutions already look good :+1: Feel free to close this PR, since we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree:

PR #148, opened 2020-10-08
Finished Chapter2

Perceval62:

cc @vrom911 @chshersh

Having an awesome time so far !

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -335,8 +335,13 @@ from it!
...
 ghci> :l src/Chapter2.hs
 -}
+
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList x y l
+  | x > y = []
+  | x == y = take 1 (drop x l)
+  | x < 0 || y < 0 = []
+  | otherwise = take ((y + 1) - x) (drop x l)
```
vrom911, line 344: This is absolutely correct! 👌🏼 

However, the second case is actually covered by the last last. If you would have `x == y` then the last case would be:

```haskell
| otherwise = take (x + 1 - x) (drop x l) = take 1 (drop x l)
```
See how we came to the second case 🙂

src/Chapter2.hs
```diff
@@ -348,8 +353,10 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+
+firstHalf :: [a] -> [a]
+firstHalf l =
+  let half = fst (divMod (length l ) 2) in take half l
```
vrom911, line 359: Great usage of `let-in` construction! Makes code easily readable 💯 

Note, that as you use only the `fst` of `divMod`, you can use `div` function straightaway 🙂

src/Chapter2.hs
```diff
@@ -605,7 +615,8 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate (x:xs) = [x, x] ++ duplicate xs
```
vrom911, line 618: In general, it's always a good idea to try to replace `l ++ [x]` with `x : l` whenever possible, though, in some cases it might not be that trivial to achieve the same result. Remember the analogy with trains :bullettrain_side: :train: :train: :train:


Here, It is more efficient to construct lists using the cons operator directly without introducing intermediate values.
```suggestion
duplicate (x:xs) = x : x duplicate xs
```

src/Chapter2.hs
```diff
@@ -620,7 +631,11 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+
+takeEven :: [Int] -> [Int]
+takeEven (x:_:xs) = x : takeEven xs
+takeEven (x:_) = [x]
+takeEven _ = []
```
vrom911, line 638: Great pattern matching skills 👏🏼 🏅

src/Chapter2.hs
```diff
@@ -727,7 +742,8 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate (x:xs) = replicate x x ++ smartReplicate xs
+smartReplicate [] = []
```
vrom911, line 746: You can notice that you need to get through each element of the list, turn it into the list of repeating elements and then concatenate all such lists together. This is exactly what you are doing 👍🏼 

I just want to mention, that you could use standard functions to do the same as you do as well: `map`.

However after using `map (\x -> replicate x x)` on the list you will get a list of lists. To flatten such a list, you can use the `concat` function:

```haskell
[3,1,2,4,0,1]
```

src/Chapter2.hs
```diff
@@ -740,8 +756,9 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
 
+contains :: Int -> [[Int]] -> [[Int]]
+contains n l = filter (elem n) l
```
vrom911, line 761: Now, after you've mastered eta-reduction, you can apply such a technique to this function as well  🙂 

src/Chapter2.hs
```diff
@@ -780,13 +797,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Ord a => a -> [a] -> [a]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: Num c => [c] -> [c] -> [c]
+pairMul = zipWith (*)
```
vrom911, line 808: Eta-reduction award 🏆

src/Chapter2.hs
```diff
@@ -857,8 +883,13 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
-
+rewind :: [Int] -> [Int]
+rewind l =
+  if len > 0
+    then drop prev l ++ rewind (take prev l)
+    else []
```
vrom911, line 890: Very creative solution! 🧑‍🎨 

Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Chapter Two is done, bravo 🕺🏼 

I left some comments, hope it makes sense. Let me know if you have any questions 🙂

Perceval62:
Thanks for the reviews ! I will re-read the sections where I should improve !

PR #151, opened 2020-10-08
Fix a typo where wrong function is called in example

arthurborisow:
Just a small typo

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -613,7 +613,7 @@ Implement a function that returns the sum of the last two digits of a number.
 6
 >>> sumLast2 134
 7
->>> sumLast 1
+>>> sumLast2 1
 1
 
 Try to introduce variables in this task (either with let-in or where) to avoid
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -645,7 +645,7 @@ firstDigit n = error "firstDigit: Not implemented!"
 
 
 {-
-You did it! Now it is time to the open pull request with your changes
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -613,7 +613,7 @@ Implement a function that returns the sum of the last two digits of a number.
 6
 >>> sumLast2 134
 7
->>> sumLast 1
+>>> sumLast2 1
```
chshersh, line 616: @arthurborisow Thanks for the fix! Would it be for hard for you to also fix the type of this comment? Just replace

```haskell
{-
=⚔️= Task 9
```

with

```haskell
{- |
=⚔️= Task 9
```

few lines above. Doctest tests only functions inside special comments (with `|` after `{-`), and this part was missing for a comment in this task. That's why we didn't notice this error before.

chshersh (approved):

src/Chapter1.hs
arthurborisow: Sure, no problem!

arthurborisow: @chshersh Another thing that I noticed: `Now it is time to the open pull request with your changes` - `the` part is redundant I believe

vrom911: Yes, nice catch, @arthurborisow! You can add it to this PR as well, if you don't mind 🙂

arthurborisow: Done

chshersh:
@arthurborisow Many thanks! :hugs:

PR #152, opened 2020-10-08
Chapter2typo

pennychase:
### Edit for Chapter 2

The description of lambdas has an error on line 654 ("append number 3" should be "add number 3"

cc @vrom911 @chshersh


chshersh:
Hi @pennychase! Thanks for the fix :slightly_smiling_face: However, I see that your PR also contains solutions to Chapter One, which we obviously can't change. Would you be able to rewrite the history of your commits to keep only the relevant change?

PR #153, opened 2020-10-08
Chapter 2 Solutions

Volence:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -506,6 +506,7 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
+isThird42 :: [Int] -> Bool
 isThird42 (_ : _ : 42 : _) = True
 isThird42 _ = False
 
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -74,6 +74,7 @@ When working with lists, the most practical module will be "Data.List":
 
    * http://hackage.haskell.org/package/base-4.14.0.0/docs/Data-List.html
 -}
+import Control.Applicative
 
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -832,7 +859,8 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate rotateLength l = drop rotateLength $ take (length l + rotateLength) (cycle l)
```
vrom911, line 863: Unfortunately, `cycle` fails in runtime on empty lists :recycle: So you need to handle this case separately for this function to work properly.

Also, you should consider how to handle negative `rotateLength` as well 🙂

src/Chapter2.hs
```diff
@@ -848,7 +876,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = (rewind xs) ++ [x]
```
vrom911, line 881: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Amazing! You have really impressive solutions with interesting ideas! Well done 🎉

chshersh:

PR #155, opened 2020-10-08
finished chap 1

cobrien99:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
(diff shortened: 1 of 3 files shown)


src/Chapter1.hs
```diff
@@ -416,7 +416,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Num a => a -> a -> a
```

src/Chapter1.hs
```diff
@@ -478,7 +478,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Integer -> Integer --Integral a => a -> a didn't work for some reason
+lastDigit n = mod n 10
```
chshersh, line 482: Your implementation is almost correct :ok: 

src/Chapter1.hs
```diff
@@ -640,8 +654,8 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Integer -> Integer
+firstDigit n = if n < 10 then n else firstDigit (div n 10)
```
chshersh, line 658: I see, that for a negative number like -19, this implementation will return the negative number itself instead of the last digit because the first check n < 10 will always succeed for negative numbers.

chshersh (approved):
Nice work on solving our tasks!
And well done with the Advanced Task :+1: 

I reviewed only Chapter One. I see that you submitted solutions to first three chapters in a single PR. Our course assumes, that you will open 4 PRs, so you can get your contribution towards Hacktoberfest T-Shirt :shirt: or plant :deciduous_tree: But let us know, if you want to have reviews of all chapters in this single PR :slightly_smiling_face:

cobrien99:
Hi. Thanks for the feedback? Sorry for the noob question but how do I open a different PR for every chapter? DO I need to be on a separate branch every time?

chshersh:
@cobrien99 No worries, it's not a noob question!

Our course recommends opening PRs in your own fork. So, for example, you start from `master` branch, you create new branch `solutions-chapter1` (or something like that), you implement solutions in this branch, and then you open a PR to `master` in your own fork. The same process is repeated for each chapter. And in the end you have solutions for all 4 chapters and all CI checks are green.

The same process can be repeated for our repository as well. But for forks, it's nicer. Since we can't merge your PRs in our repo, you'll either end up having separate solutions in 4 different branches without having all solutions in a single place, or you will open consequence PRs with solutions to one chapter, two chapters at the same time, three chapters (where two are already reviewed by us), and so on. So yeah, different git workflows with different trade-offs...

vrom911:

PR #157, opened 2020-10-08
Solutions to Chapter Two

cobrien99:
### Solutions for Chapter 2

I hope this is the right place

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,52 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
+[True, False] :: [Bool]
 
 
 String is a list of characters:
 >>> :t "some string"
+"some string" :: [Char]
 
 
 Empty list:
 >>> :t []
+[] :: [a]
 
 
 Append two lists:
 >>> :t (++)
+(++) :: [a] -> [a] -> [a]
 
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
+(:) :: a -> [a] -> [a]
 
 
 Reverse a list:
 >>> :t reverse
+reverse :: [a] -> [a]
 
 
 Take first N elements of a list:
 >>> :t take
+take :: Int -> [a] -> [a]
 
 
 Create list from N same elements:
 >>> :t replicate
+replicate :: Int -> a -> [a]
 
 
 Split a string by line breaks:
 >>> :t lines
+lines :: String -> [String]
 
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
@@ -186,31 +195,31 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
-
+[10,2,3,1,5]
 >>> [] ++ [1, 4]  -- [] is an empty list
-
+[1,4]
 >>> 3 : [1, 2]
-
+[3,1,2]
 >>> 4 : 2 : [5, 10]  -- prepend multiple elements
-
+[4,2,5,10]
 >>> [1 .. 10]  -- list ranges
-
+[1,2,3,4,5,6,7,8,9,10]
 >>> [10 .. 1]
-
+[]
 >>> [10, 9 .. 1]  -- backwards list with explicit step
-
+[10,9,8,7,6,5,4,3,2,1]
 >>> length [4, 10, 5]  -- list length
-
+3
 >>> replicate 5 True
-
+[True,True,True,True,True]
 >>> take 5 "Hello, World!"
-
+"Hello"
 >>> drop 5 "Hello, World!"
-
+", World!"
 >>> zip "abc" [1, 2, 3]  -- convert two lists to a single list of pairs
-
+[('a',1),('b',2),('c',3)]
 >>> words "Hello   Haskell     World!"  -- split the string into the list of words
-
+["Hello","Haskell","World!"]
 
 
 👩‍🔬 Haskell has a lot of syntax sugar. In the case with lists, any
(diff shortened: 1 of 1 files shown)


cobrien99:
cc @vrom911 @chshersh

I think I have uploaded the right file this time

src/Chapter2.hs
```diff
@@ -336,7 +337,12 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList _ _ [] = []
+subList a b (x:xs) | b < a = []
+                    | a < 0 = []
+                    | a > 0 = subList (a-1) (b-1) xs
+                    | b >= 0 = x : subList a (b - 1) xs
+                    | otherwise = []
```
vrom911, line 345: That is an interesting idea of the implementation through the recursion 🎨 

However, you could use the standard functions `take` and `drop` to solve this problem as well 🙂 
You would still need to make the checks on negative indexes and if `b < a` as you rightly did in here 👍🏼 

Additional side note: in Haskell, the operators are usually separated from variables with spaces: `a - 1` instead of `a-1` 🙂

src/Chapter2.hs
```diff
@@ -348,8 +354,12 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf [] = []
+firstHalf (x:xs)  | l > 0 = x : firstHalf xs
+                  | otherwise = []
+                  where l = div (length (x:xs) - 1) 2
```
vrom911, line 361: `length (x:xs) - 1` is the same as `length xs` 🙂 

But generally note, that this solution would be slow, as `length` function on lists in Haskell is not cheap. It walks each element of the list in order to get the length. So when you recursively ask `length` on each interaction, this would be too expensive 😞 

Luckily, here, you can also solve it with `take` function. You already found how many elements you need to take (by the way, nice job putting it into `where` construction, very readable 👍🏼 ).

src/Chapter2.hs
```diff
@@ -620,7 +633,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven (x:_:xs) = x : takeEven xs
```
vrom911, line 637: Great pattern matching skills 👏🏼

src/Chapter2.hs
```diff
@@ -727,7 +743,9 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate (x:xs) = replicate x x ++ smartReplicate xs
+--negative numbers will break it :)
```
vrom911, line 748: The way replicate works, it won't break it 🙂 

```haskell
>>> replicate (-1) 9
[]
```

So, it will just put an empty list instead of the elements 🙂

src/Chapter2.hs
```diff
@@ -727,7 +743,9 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate (x:xs) = replicate x x ++ smartReplicate xs
```
vrom911, line 747: This is right! 💯 

```haskell
[3,1,2,4,0,1]
```

src/Chapter2.hs
```diff
@@ -740,7 +758,9 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Integer -> [[Integer]] -> [[Integer]]
+contains _ [] = []
+contains n (x:xs) = if elem n x == True then x : contains n xs else contains n xs
```
vrom911, line 763: `elem n x` is bool itself, so no need to compare it to `True`, just `if elem n x then...`.

On a general note, there is the standard `filter` function, that would do this check for you 🙂

src/Chapter2.hs
```diff
@@ -780,13 +800,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Integer -> [Integer] -> [Integer]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Integer] -> [Integer] -> [Integer]
+pairMul = zipWith (*)
```
vrom911, line 811: Eta-reduction award 🏆

src/Chapter2.hs
```diff
@@ -841,7 +863,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate _ [] = []
+rotate n xs | n < 1 = []
```
vrom911, line 868: On negative numbers it should return an empty list, however, if it is `0`, then it means only that the list don't rotate, so you need to return the list as it is 👍🏼

src/Chapter2.hs
```diff
@@ -857,7 +882,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = (rewind xs) ++ [x]
```
vrom911, line 887: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Great job! Chapter Two is done ✔️ 👏🏼 🥇

vrom911:

PR #164, opened 2020-10-09
Chapter 2 completed

lprakashv:
### Solutions for Chapter Two

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,44 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
+[True, False] :: [Bool]
 
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +350,7 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList start end list = take (end - start + 1) (drop start list)
```
chshersh, line 353: The main idea is correct :+1: Though, it looks like this implementation doesn't work correctly on negative bounds. Corner cases sometimes can be tricky to discover and think about...

src/Chapter2.hs
```diff
@@ -534,6 +550,7 @@ For example, we can patch the previous function to count the number of
 steps we need to take in order to reduce the number to zero.
 
 🤔 Blitz question: can you guess what this number represents?
+It represents "Binary traversal / search" divide and conquer strategy / algorithm => O (log (N))
```
chshersh, line 553: Smart answer :nerd_face:

src/Chapter2.hs
```diff
@@ -620,7 +638,12 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven = go ([], 0)
+  where
+    go :: ([a], Int) -> [a] -> [a]
+    go (res, _) [] = res
+    go (res, i) (x:xs) = if even i then go (res ++ [x], i+1) xs else go (res, i+1) xs
```
chshersh, line 646: Usually, adding elements to the end of the list is costly. Your implementation is correct :+1: But you can try to implement this function more efficiently by using `:` instead `++`. If you can think, how to avoid adding elements to the end of the list, and instead add to the beginning, but get the same result, that would be amazing!

In general, it's always a good idea to try to replace `l ++ [x]` with `x : l` whenever possible, though, in some cases it might not be that trivial to achieve the same result. Remember the analogy with trains :bullettrain_side: :train: :train: :train:

src/Chapter2.hs
```diff
@@ -727,7 +750,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concatMap (\i -> replicate i i) l
```
chshersh, line 753: Nice usage of standard function `concatMap`! Now, after you've mastered eta-reduction, you can apply such a technique to other functions as well  🙂 

src/Chapter2.hs
```diff
@@ -857,7 +884,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
chshersh, line 889: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Again, remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
And well done with the Advanced Challenge :clap: 

I left some comments, thoughts and suggestions. I hope they can help to move forward and improve existing solutions :slightly_smiling_face:

src/Chapter2.hs
lprakashv: Thank you :)

src/Chapter2.hs
lprakashv: Fixed

src/Chapter2.hs
lprakashv: Fixed

src/Chapter2.hs
```diff
@@ -642,8 +644,8 @@ takeEven :: [a] -> [a]
 takeEven = go ([], 0)
   where
     go :: ([a], Int) -> [a] -> [a]
-    go (res, _) [] = res
-    go (res, i) (x:xs) = if even i then go (res ++ [x], i+1) xs else go (res, i+1) xs
+    go (res, _) [] = reverse res
+    go (res, i) (x:xs) = if even i then go (x:res, i+1) xs else go (res, i+1) xs
```
chshersh, line 648: This is correct and almost perfect :ok_hand: 

Though, `reverse` is slow as it traverses the whole list (as you know because you've implemented `rewind`. Note, that here, if you replace `go (x : res)` with `x : go xs`, you will get a proper order of elements without the need to use `reverse` in the end, and you will also get rid of the additional variable `res`

chshersh (approved):

chshersh:

PR #165, opened 2020-10-10
Complete `Chapter3`

ufocoder:
### Solutions for Chapter 3

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -398,11 +398,11 @@ data Monster = Monster
 fight :: Knight -> Monster -> Coins
 fight knight monster
       | kinghtSteps < monsterSteps = knightGold knight + monsterGold monster
-      | kinghtSteps > monsterSteps = (-1)
+      | kinghtSteps > monsterSteps = -1
       | otherwise = knightGold knight
     where
-      kinghtSteps = (monsterHealth monster) / (knightAttack knight)
-      monsterSteps = (knightHealth knight) / (monsterAttack monster)
+      kinghtSteps = monsterHealth monster / knightAttack knight
+      monsterSteps = knightHealth knight / monsterAttack monster
 
 {- |
 =🛡= Sum types
@@ -490,7 +490,7 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
-data Breakfast 
+data Breakfast
     = Tea
     | Coffee
     | Juice
@@ -523,24 +523,24 @@ After defining the city, implement the following functions:
 data Building = Church | Library deriving (Show)
 data House = One | Two | Three | Four deriving (Enum, Show)
 
-data City = City 
+data City = City
     { cityCastle :: Maybe Castle
     , cityWalls  :: Bool
     , cityBuilding  :: Building
     , cityHouses  :: [House]
     } deriving Show
 
-data Castle = Castle String deriving Show
+newtype Castle = Castle String deriving Show
 
 buildCastle :: String -> City -> City
-buildCastle castleName city = 
+buildCastle castleName city =
     city { cityCastle = Just (Castle castleName) }
 
 buildHouse :: City -> City
-buildHouse city = 
+buildHouse city =
     city { cityHouses = Two : cityHouses city }
 
-buildWalls :: City -> City 
+buildWalls :: City -> City
 buildWalls city =
   case cityCastle city of
     Just _ ->
@@ -837,9 +837,7 @@ data TreasureChest x = TreasureChest
     , treasureChestLoot :: x
     }
 
-data Dragon x = Dragon 
-    { dragonPower :: x
-    }
+newtype Dragon x = Dragon {dragonPower :: x}
 
 data DragonLair treasure dragon = DragonLair
     { dragonLairChest  :: Maybe (TreasureChest treasure)
@@ -1097,7 +1095,7 @@ nextDay day
   | otherwise = succ day
 
 daysToParty :: DayOfWeek -> Int
-daysToParty day 
+daysToParty day
     | day > Friday = fromEnum Sunday - fromEnum day + fromEnum Friday + 1
     | otherwise = fromEnum Friday - fromEnum day
 
@@ -1143,7 +1141,7 @@ My doctests:
 
 >>> :{
   fight'
-    FightKnight (FightHealth 10) (FightAttack 2) (FightDefense 0) [KnightAttack, KnightDrinkPotion 1] 
+    FightKnight (FightHealth 10) (FightAttack 2) (FightDefense 0) [KnightAttack, KnightDrinkPotion 1]
     FightMonster (FightHealth 10) (FightAttack 3) [MonsterAttack]
 :}
 Victory
@@ -1168,18 +1166,18 @@ newtype FightHealth = FightHealth Int deriving (Show, Eq)
 newtype FightAttack = FightAttack Int deriving (Show, Eq)
 newtype FightDefense = FightDefense Int deriving (Show, Eq)
 
-data FightAction 
-  = KnightAttack 
+data FightAction
+  = KnightAttack
   | KnightDrinkPotion Int
   | KnightCastSpell Int
-  | MonsterAttack 
-  | MonsterRun 
+  | MonsterAttack
+  | MonsterRun
   deriving Show
 
-data FightResult 
-  = Round FightKnight FightMonster 
-  | Wait 
-  | Victory 
+data FightResult
+  = Round FightKnight FightMonster
+  | Wait
+  | Victory
   | Defeat deriving Show
 
 data FightKnight = FightKnight
@@ -1189,7 +1187,7 @@ data FightKnight = FightKnight
     , fightKnightActions :: [FightAction]
     } deriving Show
 
-data FightMonster = FightMonster 
+data FightMonster = FightMonster
     { fightMonsterHealth  :: FightHealth
     , fightMonsterAttack :: FightAttack
     , fightMonsterActions :: [FightAction]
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -374,6 +381,29 @@ after the fight. The battle has the following possible outcomes:
...
+    } deriving (Show)
+
+fight :: Knight -> Monster -> Coins
+fight knight monster
+      | kinghtSteps < monsterSteps = knightGold knight + monsterGold monster
+      | kinghtSteps > monsterSteps = -1
+      | otherwise = knightGold knight
+    where
+      kinghtSteps = monsterHealth monster / knightAttack knight
+      monsterSteps = knightHealth knight / monsterAttack monster
```
vrom911, line 405: Interesting decision! ⚔️ 
Could you please elaborate a bit on what the `step` is in this context? 🙂

src/Chapter3.hs
```diff
@@ -374,6 +381,29 @@ after the fight. The battle has the following possible outcomes:
 
 -}
 
+type Coins = Int
```
vrom911, line 384: Great usage of the type alias 👌🏼

src/Chapter3.hs
```diff
@@ -460,6 +490,15 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
...
 -}
 
+data Breakfast
+    = Tea
+    | Coffee
+    | Juice
+    | Pancake
+    | Jam
+    | Toast
+    | Oatmeal
```
vrom911, line 500: Mmm 🥞 🍞 ☕

src/Chapter3.hs
```diff
@@ -480,6 +519,37 @@ After defining the city, implement the following functions:
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
 
+
+data Building = Church | Library deriving (Show)
+data House = One | Two | Three | Four deriving (Enum, Show)
+
+data City = City
+    { cityCastle :: Maybe Castle
+    , cityWalls  :: Bool
```
vrom911, line 528: This is the correct data type, that can represent what we have in requirements 👍🏼 
However, this also allows having some states that should not ideally exist: like having walls without any castle. With this representation, this is possible. However, we actually can restrict this on type-level by providing more representative types 🙂 This is actually the greatest power of Haskell and type system 😍

src/Chapter3.hs
```diff
@@ -480,6 +519,37 @@ After defining the city, implement the following functions:
...
+
+newtype Castle = Castle String deriving Show
+
+buildCastle :: String -> City -> City
+buildCastle castleName city =
+    city { cityCastle = Just (Castle castleName) }
+
+buildHouse :: City -> City
+buildHouse city =
+    city { cityHouses = Two : cityHouses city }
```
vrom911, line 541: Nice one! 🏠 

If you want to be able to build any type of the house, you can just pass it into the function itself. So the type could also be:

```haskell
buildHouse :: House -> City -> City
```

src/Chapter3.hs
```diff
@@ -480,6 +519,37 @@ After defining the city, implement the following functions:
...
+
+buildWalls :: City -> City
+buildWalls city =
+  case cityCastle city of
+    Just _ ->
+      if people > 10 then city { cityWalls = True } else city
+    Nothing ->
+      city
+  where
+    people = sum $ map ((+1) . fromEnum) $ cityHouses city
```
vrom911, line 551: Very neat solution with the `Enum` instance 🧠 🎨

src/Chapter3.hs
```diff
@@ -560,22 +630,31 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defense (armor * dexterity)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage damage) (Defense defense) (Health health) = Health (health + defense - damage)
```
vrom911, line 657: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -907,9 +998,21 @@ Implement instances of "Append" for the following types:
...
+instance Append Gold where
+    append (Gold x) (Gold y) = Gold (x + y)
+
+instance Append [a] where
+    append = (++)
+
+instance Append a => Append (Maybe a) where
+    append :: Maybe a -> Maybe a -> Maybe a
+    append (Just x) (Just y) = Just (append x y)
+    append _ _ = Nothing
```
vrom911, line 1015: Here, even if one of the arguments is `Just` it will return `Nothing`, though more correct to be to return the `Just`.

src/Chapter3.hs
```diff
@@ -971,6 +1074,31 @@ implement the following functions:
...
+    deriving (Show, Enum, Eq, Ord)
+
+isWeekend :: DayOfWeek -> Bool
+isWeekend Saturday = True
+isWeekend Sunday = True
+isWeekend _ = False
+
+nextDay :: DayOfWeek -> DayOfWeek
+nextDay day
+  | day == Saturday = Monday
```
vrom911, line 1094: That is right!

```suggestion
  | day == maxBound = minBound
```

src/Chapter3.hs
```diff
@@ -971,6 +1074,31 @@ implement the following functions:
...
+  | otherwise = succ day
+
+daysToParty :: DayOfWeek -> Int
+daysToParty day
+    | day > Friday = fromEnum Sunday - fromEnum day + fromEnum Friday + 1
+    | otherwise = fromEnum Friday - fromEnum day
```
vrom911, line 1100: This is correct!
However, this could be a bit simplified and made a bit more independent of the data type structure 👍🏼 

We know for sure that it is 7 days in a week. So we can use `mod` function on the difference, that will tell us exactly how many days left to the Friday:

```haskell
>>> mod (-1) 7
6
```

src/Chapter3.hs
```diff
@@ -1007,6 +1135,145 @@ contestants, and write a function that decides the outcome of a fight!
 -}
 
 
+
+{-
+My doctests:
```
vrom911, line 1140: Wow 👏🏼 This is SOO cool 😎 

Thanks a lot for doing this 🥇

src/Chapter3.hs
```diff
@@ -1007,6 +1135,145 @@ contestants, and write a function that decides the outcome of a fight!
...
+newtype FightHealth = FightHealth Int deriving (Show, Eq)
+newtype FightAttack = FightAttack Int deriving (Show, Eq)
+newtype FightDefense = FightDefense Int deriving (Show, Eq)
+
+data FightAction
+  = KnightAttack
+  | KnightDrinkPotion Int
+  | KnightCastSpell Int
+  | MonsterAttack
+  | MonsterRun
```
vrom911, line 1174: Nice sum type!

But I think it could be more type-safe, if we would have Monster and Knight actions separately. In that case, we will restrict the possible state to only relevant ones. 🙂

src/Chapter3.hs
```diff
@@ -1007,6 +1135,145 @@ contestants, and write a function that decides the outcome of a fight!
...
+processFightActions Victory _ = Victory
+processFightActions Defeat _ = Defeat
+processFightActions Wait _ = Wait
+processFightActions (Round knight monster) [] =
+    let newActions = createFightActions knight monster in
+      if length newActions == 0 then Wait else
+        processFightActions (Round knight monster) newActions
+
+processFightActions (Round knight monster) (action:actions) =
+    processFightActions (processRoundAction action knight monster) actions
```
vrom911, line 1267: Oh, I see now that you have a shared pool of actions for both of them!

This is a nice idea, however, it doesn't give guarantees that they will go in turns then 🤔

src/Chapter3.hs
```diff
@@ -1007,6 +1135,145 @@ contestants, and write a function that decides the outcome of a fight!
...
+
+createFightActions :: FightKnight -> FightMonster -> [FightAction]
+createFightActions knight monster =
+  fightKnightActions knight ++ fightMonsterActions monster
```
vrom911, line 1271: Oh, Interesting!
Yeah, wouldn't in this case, knight perform all its actions before monster even starts the fight?

vrom911 (approved):
Fantastic job on this Chapter!
Very interesting solutions! 👏🏼 

Oh, also, amazing and solid work on the advanced task, very thoughtful solution 👍🏼

src/Chapter3.hs
ufocoder: As I understand knight and monster should hit each other one by one in turn until one of them will die. 
So I need to know how much rounds (steps) need for each of them. The one with fewer steps will win ;)

src/Chapter3.hs
ufocoder: Thanks for this information!

src/Chapter3.hs
ufocoder: Cool, I've made a note

src/Chapter3.hs
ufocoder: I've implemented as I've understood the following >_< :
```
Each fighter starts with some list of actions they can do, performs them in sequence one after another...
```
Now I've got that action list should be look like:

```
[KnightAttack, MonsterAttack, KnightDrinkPotion Int, .. ]
```

src/Chapter3.hs
vrom911: Oh, very clever! I see now! 👍🏼

src/Chapter3.hs
vrom911: Yeap, I see that this is just another understanding of the task, no worries 🙂

PR #166, opened 2020-10-10
chapter 2

FredericEspiau:
### Solutions for Chapter 2

cc @vrom911 @chshersh

Thanks for the hard work 🙏 

This chapter was awesome

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +336,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList start end list
+  | start < 0 || end < start = []
+  | otherwise = take (end - start + 1) (drop start list)
```
chshersh, line 341: Perfect implementation :gem:

src/Chapter2.hs
```diff
@@ -605,7 +611,8 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate [] = []
+duplicate (x:xs) = [x, x] ++ duplicate xs
```
chshersh, line 615: It is more efficient to construct lists using the cons operator directly without introducing intermediate values.
```suggestion
duplicate (x:xs) = x : x duplicate xs
```

src/Chapter2.hs
```diff
@@ -727,7 +737,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate x = concatMap  (\n -> replicate n n) x
```
chshersh, line 740: Now, after you've mastered eta-reduction, you can apply such a technique to other functions as well  🙂 

src/Chapter2.hs
```diff
@@ -857,7 +874,10 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind [x] = [x]
+rewind (x:xs) = rewind xs ++ [x]
```

chshersh (approved):
Congrats on solving Chapter Two tasks!

We are really happy that you are enjoying our course :blush: 

I left some comments and suggestions in my review, but your solutions are already great :top:

chshersh:

PR #170, opened 2020-10-10
Add solution for chapter 1

jborkowski:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -416,7 +416,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Num a => a -> a -> a
```
vrom911, line 419: We explain `Num` in the following chapters, so you are one step ahead 😸

src/Chapter1.hs
```diff
@@ -542,7 +542,14 @@ value after "=" where the condition is true.
...
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | x <= y && y <= z = y
+    | x <= z && z <= y = z
+    | y <= x && x <= z = x
+    | y <= z && z <= x = z
+    | z <= x && x <= y = x
+    | otherwise        = y
```
vrom911, line 552: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive` if you don't put the `otherwise` case, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

However, you can simplify your cases a bit. The first case `x <= y && y <= z = y` could be covered by this `otherwise`. So you can safely remove it 👌🏼 

src/Chapter1.hs
```diff
@@ -620,8 +633,11 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n =
+    let restAndLast    = divMod (abs n) 10
```
vrom911, line 638: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `divMod`, cool 😎 

```haskell
(aDiv, aMod) = divMod (abs n) 10
```

So you don't need to use `fst` and `snd` 👍🏼

src/Chapter1.hs
```diff
@@ -620,8 +633,11 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
+        restAndPreLast = divMod (fst restAndLast) 10
```
vrom911, line 639: As you only need the `mod` part of `divMod`, you can just use `mod` function on its own in here:

```haskell
        preLast = mod (fst restAndLast) 10
```

src/Chapter1.hs
```diff
@@ -641,8 +657,12 @@ You need to use recursion in this task. Feel free to return to it later, if you
...
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit n = go (divMod (abs n) 10)
+  where
+    go :: (Int, Int) -> Int
+    go (0, lastD) = lastD
+    go (rest, _)  = go (divMod rest 10)
```
vrom911, line 665: Great and clean solution! 👏🏼

vrom911 (approved):
Very thoughtful solutions 🥇

vrom911:

PR #171, opened 2020-10-11
Chapter 2 completed!

RajPathare:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +336,10 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList x y z = 
+    if x >=0 && y >=x
+    then take (y - x + 1) (drop x z)
+    else []
```
chshersh, line 342: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding a separate case to your check :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -348,8 +351,10 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf l = 
+    let a = length l `div` 2 - 1
+    in subList 0 a l
```
chshersh, line 357: Nice usage of `subList` here! Though, `subList` does more than required, and here it's enough to _take_ the first half of the list.

src/Chapter2.hs
```diff
@@ -604,9 +612,10 @@ Implement a function that duplicates each element of the list
 "aabbaacc"
 
 -}
-duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
 
+duplicate :: [a] -> [a]
+duplicate [] = []
+duplicate (x:y) = x : x : duplicate y
```
chshersh, line 618: Idiomatically, the list tail is called _xs_ usually when the variable is `x`
```suggestion
duplicate (x:xs) = x : x : duplicate xs
```

src/Chapter2.hs
```diff
@@ -620,7 +629,11 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven [x] = [x]
+takeEven (x:_:y) = x : takeEven y
```
chshersh, line 636: Great pattern-matching skills! :trophy:

src/Chapter2.hs
```diff
@@ -726,8 +739,10 @@ value of the element itself
 
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
+
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate l = concat (map (\x -> replicate x x) l)
```
chshersh, line 745: Great! 👍🏼 

Also, a separate case with `[]` is redundant. `map` and `concat` already handle empty lists without problems :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -857,7 +883,11 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+
+rewind :: [a] -> [a]
+rewind [] = []
+rewind [x] = [x]
+rewind (x:y) = rewind y ++ [x]
```
chshersh, line 890: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
Nice solutions for Chapter Two tasks! And well done with the Advanced Task :clap: 

I left some comments and suggestions. Your solutions are great, but you can make them even more awesome :100:

vrom911:

PR #174, opened 2020-10-11
Chapter 1 Solutions

siv12345:
### Solutions for Chapter {1}

cc @vrom911 @chshersh
 When I run the test suite
I get some errors in Task 5 where I have to find the number which is closer to zero from 2 negative numbers, but other than that it seems to be alright

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -508,7 +509,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = min (abs x) (abs y)
```
chshersh, line 512: This solution is almost correct :ok_hand: However, on negative numbers, it returns the absolute value of those negative numbers, note the value itself.

src/Chapter1.hs
```diff
@@ -541,8 +542,11 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | x > y = mid y x z
+  | y > z = mid x z y
+  | otherwise = y
```
chshersh, line 549: Short and clever solution!

src/Chapter1.hs
```diff
@@ -619,9 +626,8 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n = (mod (div (abs n) 10) 10) + (mod (abs n) 10)
```
chshersh, line 630: You were encouraged to use variables here to increase code readability :slightly_smiling_face: 

```haskell
```

src/Chapter1.hs
```diff
@@ -640,8 +646,10 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+  | div n 10 == 0 = n
```
chshersh, line 651: `div` on negative numbers returns negative number:

```haskell
ghci> div (-1) 10
-1
```

That's why this function hangs on CI and runs into an infinite loop. So, to make tests pass, you need to handle corner cases in this function :slightly_smiling_face:

chshersh (approved):
Congrats with finishing the first chapter!

I left some comments and suggestions. Hope they help you to move forward :slightly_smiling_face:

siv12345:
Thank you very much for the review of my PR @chshersh. I will make the changes you mentioned, locally on my machine. 

Can I close this PR?

vrom911:
@siv12345 No, that is fine! Feel free to close this PR. The contributions would be counted toward Hacktoberfest 😉

siv12345:
okay thanks, closing this pr

PR #175, opened 2020-10-11
Add solutions for chapter 2

jborkowski:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -739,7 +739,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate = concat . map (\n -> replicate n n)
+smartReplicate = concatMap (\n -> replicate n n)
 
 {- |
 =⚔️= Task 9
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -337,7 +337,9 @@ ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
 subList from to ls
-  | from < 0 || to < 0 = []
+  | from < 0  = []
+  | to < 0 = []
+  | from > to = []
   | otherwise = take n . drop from $ ls
   where
     n = to - from + 1
@@ -353,7 +355,7 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 firstHalf :: [a] -> [a]
-firstHalf ls = fst . splitAt half $ ls
+firstHalf ls = take half ls
   where
     half = length ls `div` 2
 
@@ -507,10 +509,8 @@ True
 False
 -}
 isThird42 :: (Eq a, Num a) => [a] -> Bool
-isThird42 []         = False
 isThird42 (_:_:42:_) = True
 isThird42 _          = False
-
 
 {- |
 =🛡= Recursion
@@ -629,7 +629,7 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven :: (Eq a, Num a) => [a] -> [a]
+takeEven :: [a] -> [a]
 takeEven []      = []
 takeEven [a]     = [a]
 takeEven (h:_:t) = h : takeEven t
@@ -752,7 +752,7 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains :: (Eq a, Num a) => a -> [[a]] -> [[a]]
+contains :: Eq a => a -> [[a]] -> [[a]]
 contains n = filter (elem n)
 
 {- |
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +336,11 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList from to ls
+  | from < 0 || to < 0 = []
+  | otherwise = take n . drop from $ ls
```

src/Chapter2.hs
```diff
@@ -348,8 +352,10 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf ls = fst . splitAt half $ ls
+  where
+    half = length ls `div` 2
```
chshersh, line 358: This is a correct solution! Though, `splitAt` does more than required. Here, it's enough to _take_ the first half of the list

src/Chapter2.hs
```diff
@@ -500,7 +506,10 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird42 :: (Eq a, Num a) => [a] -> Bool
+isThird42 []         = False
```
chshersh, line 510: This case is redundant, since it will be covered by the `_` pattern on the last line

src/Chapter2.hs
```diff
@@ -620,7 +629,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: (Eq a, Num a) => [a] -> [a]
```
chshersh, line 632: Constraints here are redundant. Notice that you don't use any specific information about values of type `a`, so no need to require anything special here :slightly_smiling_face: 
```suggestion
takeEven :: [a] -> [a]
```

src/Chapter2.hs
```diff
@@ -740,8 +752,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: (Eq a, Num a) => a -> [[a]] -> [[a]]
```
chshersh, line 755: The `Num` constraint is redundant here. Note, that `elem` needs only `Eq` for this function to work. We don't require people to write very polymorphic functions here, so by saying _list of lists of numbers_, we mean that a type signature like `Int -> [[Int]] -> [[Int]]` will be enough for this task. Nevertheless, nice work on specifying a polymorphic type signature :+1: 
```suggestion
contains :: (Eq a) => a -> [[a]] -> [[a]]
```

src/Chapter2.hs
```diff
@@ -780,13 +792,14 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
--- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: (Ord a) => a -> [a] -> [a]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: (Num a) => [a] -> [a] -> [a]
+pairMul = zipWith (*)
```
chshersh, line 802: Eta-reduction award! :trophy: 

And nice work on general types :+1:

src/Chapter2.hs
```diff
@@ -841,7 +854,12 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: (Num a) => Int -> [a] -> [a]
+rotate n ls
+  | n < 0 = []
+  | otherwise = take listLen . drop n . cycle $ ls
+  where
+    listLen = length ls
```
chshersh, line 862: Your implementation is correct. Though, you can optimize it even further. Note, that if `n` is a big number, `drop` will do a lot of work, if the list size is much smaller than the number `n`. It looks like it possible to optimize this function a bit :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -857,7 +875,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind []    = []
+rewind (h:t) = rewind t ++ [h]
```

chshersh (approved):

I left some notes in my review. Keep up the great work :muscle:

jborkowski:
@chshersh Thanks for comments! I improved my code :)

chshersh:
@jborkowski Nice work! The solutions look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to open PRs with solutions to other chapters!

PR #176, opened 2020-10-11
Chapter3 completed

pennychase:
### Solutions for Chapter {3}

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -1218,6 +1218,7 @@ mkMonster name attack health actions =
           , monsterActions = filter isMonsterAction actions
           }
 
+-- Typeclass for Fighters
 class (Fighter b) where
   getName :: b -> String
   getAttack :: b -> Int
@@ -1293,8 +1294,8 @@ roundKnight (k, f) =
   case action of
     Nothing                     -> (k, f)
     Just AttackFoe              -> (k', calculateDamage f (getAttack k))
-    Just (CastSpell n)          -> (k' { knightHealth = (knightHealth k') + n }, f)
-    Just (DrinkHealthPotion n)  -> (k' { knightDefense = (knightDefense k') + n}, f)
+    Just (CastSpell n)          -> (k' { knightDefense = (knightDefense k') + n }, f)
+    Just (DrinkHealthPotion n)  -> (k' { knightHealth = (knightHealth k') + n}, f)
   where
     (action, k') = nextAction k
 
@@ -1308,12 +1309,17 @@ roundMonster (m, f) =
   where 
     (action, m') = nextAction m
 
+-- Winner checks if fight is over (one fighter's health <= 0). Return  Nothing if
+-- no winner
 winner :: (Fighter a, Fighter b) => a -> b -> Maybe String
 winner f1 f2 
   | getHealth f1 <= 0 = Just (getName f2)
   | getHealth f2 <= 0 = Just (getName f1)
   | otherwise         = Nothing
 
+-- Fight
+-- First make sure at least one fighter has actions
+-- Then fight a round and check if there's a winner. If not, recurse
 fight :: (Fighter a, Fighter b) => a -> b -> String
 fight fighter1 fighter2 =
   if (noAction fighter1) && (noAction fighter2)
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -344,6 +344,14 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+data Book = Book
+    {  bookTitle :: String
+     , bookAuthor :: String
+     , bookPublisher :: String
+     , bookYearPublished :: Int
+     , bookPages :: Int
+    }
+
 {- |
 =⚔️= Task 2
 
@@ -373,6 +381,40 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+
+data Fighter' = 
+    KnightF  { fighterHealth :: Int
+             , fighterAttack :: Int
+             , fighterGold :: Int
+             }
+
+  | MonsterF { fighterHealth :: Int
+             , fighterAttack :: Int
+             , fighterGold :: Int
+             }
+  deriving (Show)
+
+-- The instructions aren't totally clear, so I'm making some assumptions:
+--  1. A fight is a single round in which knight attacks monster and monster attacks knight
+--  2. The fight is to the death, so loser has health < 0 at the end of the round
+--  3. If both knight and monster have health >=0 neither wins
+
+-- fight consists of two attacks:
+--  1. knight attacks monster, resulting in m1 (monster with updated health)
+--  2. m1 attacks knight (resulting in knight with updated health)
+fight' :: Fighter' -> Fighter' -> Int
+fight' k m = result (attack m1 k) m1
+  where
+    m1 = attack k m
+    result k' m'
+      | fighterHealth k' > 0 && fighterHealth m' < 0  = fighterGold k' + fighterGold m'
+      | fighterHealth k' < 0 && fighterHealth m' > 0  = -1
+      | otherwise                                   = fighterGold k'
+
+-- Attack is half a round of the fight: fighter 1 attacks fighter 2, and the return value
+-- is fighter 2 with its health reduced by fighter 1's attack.
+attack :: Fighter' -> Fighter' -> Fighter'   
+attack f1 f2 = f2 { fighterHealth = fighterHealth f2 - (fighterAttack f1) }
 
 {- |
 =🛡= Sum types
@@ -460,6 +502,15 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data MealType =
+    Breakfast
+  | Lunch
+  | Brunch
+  | Tea
+  | Dinner
+  | Supper
+  | Snack
+
 {- |
 =⚔️= Task 4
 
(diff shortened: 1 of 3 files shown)


src/Chapter3.hs
```diff
@@ -374,6 +382,40 @@ after the fight. The battle has the following possible outcomes:
...
+
+  | MonsterF { fighterHealth :: Int
+             , fighterAttack :: Int
+             , fighterGold :: Int
+             }
+  deriving (Show)
+
+-- The instructions aren't totally clear, so I'm making some assumptions:
+--  1. A fight is a single round in which knight attacks monster and monster attacks knight
+--  2. The fight is to the death, so loser has health < 0 at the end of the round
```
vrom911, line 399: Zero health is equal to death. But this is not a critical assumption, I guess 🙂

src/Chapter3.hs
```diff
@@ -374,6 +382,40 @@ after the fight. The battle has the following possible outcomes:
...
+    KnightF  { fighterHealth :: Int
+             , fighterAttack :: Int
+             , fighterGold :: Int
+             }
+
+  | MonsterF { fighterHealth :: Int
+             , fighterAttack :: Int
+             , fighterGold :: Int
+             }
+  deriving (Show)
```
vrom911, line 395: Nice usage of the sum type!

The only issue with that (the reusage of the same data type for Monster and Knight) in here is that you can easily mix the constructors when calling the function, and the compiler won't tell you anything about that:

```haskell
> fight knight monster
> fight monster knight
```

Here, we had in mind to make two types for each Monster and Knight.

src/Chapter3.hs
```diff
@@ -374,6 +382,40 @@ after the fight. The battle has the following possible outcomes:
...
+    result k' m'
+      | fighterHealth k' > 0 && fighterHealth m' < 0  = fighterGold k' + fighterGold m'
+      | fighterHealth k' < 0 && fighterHealth m' > 0  = -1
+      | otherwise                                   = fighterGold k'
+
+-- Attack is half a round of the fight: fighter 1 attacks fighter 2, and the return value
+-- is fighter 2 with its health reduced by fighter 1's attack.
+attack :: Fighter' -> Fighter' -> Fighter'   
+attack f1 f2 = f2 { fighterHealth = fighterHealth f2 - (fighterAttack f1) }
+
```
vrom911, line 418: That is a nice solution!

I also see that the attack function can be inlined into the `fight'` function itself.
That means that you could just check is `fighterHealth moster < fighterAttack knight` and vice versa 🙂 
But this is just a suggestion :)

src/Chapter3.hs
```diff
@@ -460,6 +502,15 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
...
 -}
 
+data MealType =
+    Breakfast
+  | Lunch
+  | Brunch
+  | Tea
+  | Dinner
+  | Supper
+  | Snack
```
vrom911, line 512: Yummy 😋 ☕ 🥐

src/Chapter3.hs
```diff
@@ -480,6 +531,55 @@ After defining the city, implement the following functions:
...
+data Institution = Church | Library
+                deriving (Show)
+
+data House = House NumPeople
+          deriving (Show)
+
+data City =   CityWithCastle Castle Institution [House]
+            | WalledCity Wall Institution [House]
+            | City Institution [House]
+            deriving (Show)
```
vrom911, line 552: That is an interesting representation of the city as a sum type 🏙️ 

However, you can see a lot of repetition in all constructors.
I feel like it would be more convenient to work `City` if it was a record type instead of a sum type. 

First of all, if you change the castele to have walls inside, it would then be possible to only have the castle represented with three states in one record type rather than three constructors with almost the same types. Let's think about how to do that. The description says:

1. Castle with a name
3. No castle

This looks like a sum type indeed!

```haskell
data Castle 
    = Castle String
    | WallCastle String
    | None
```

Note that your idea with castle suits here perfectly as well. These zrfe just the suggestions and simplifications, or just exploration of different approaches 🙂 

So, instead of having three top-level constructors and duplicating all fields on both of them, you can define `City` as a record with the new castle:

```haskell
data City = City
    , cityInstitution :: Institution
    , cityHouses :: [House]
    }
```

src/Chapter3.hs
```diff
@@ -480,6 +531,55 @@ After defining the city, implement the following functions:
...
+
+getHouses :: City -> [House]
+getHouses (CityWithCastle _ _ houses) = houses
+getHouses (WalledCity _ _ houses) = houses
+getHouses (City _ houses) = houses
+
+peopleInHouse :: House -> Int
+peopleInHouse (House n) = (fromEnum n) + 1
```
vrom911, line 560: Perfect usage of the enum instance 👍🏼 🧑🏼‍🤝‍🧑🏼

src/Chapter3.hs
```diff
@@ -560,22 +661,32 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defense (armor * dexterity)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage damage) (Defense defense) (Health health) = Health (health + defense - damage)
```
vrom911, line 689: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -753,6 +864,21 @@ parametrise data types in places where values can be of any general type.
...
+
+data Treasure a = Treasure
+    { treasureHoard :: a
+    }
+
+data Dragon a = Dragon
+    { dragonPower :: a
+    }
+
+data Lair p t = Lair (Dragon p) (Maybe (TreasureChest t)) (Maybe (Treasure t))
```
vrom911, line 880: Great!

You could also give names to these fields as these are going to be useful while working with such non-trivial types 👍🏼

src/Chapter3.hs
```diff
@@ -910,6 +1036,23 @@ Implement instances of "Append" for the following types:
...
+  append g1 g2 = Gold (unGold g1 + unGold g2)
+
+instance Append [a] where
+  append :: [a] -> [a] -> [a]
+  append = (++)
+
+instance (Append a) => Append (Maybe a) where
+  append :: Maybe a -> Maybe a -> Maybe a
+  append (Just x) (Just y) = Just (append x y)
+  append _ _               = Nothing
```
vrom911, line 1055: While this is valid implementation and nothing in the task description says differently, in reality, it would be more beneficial is this function returned `Just` when at least one argument is `Just`.

This is how the real instance of `Maybe` works for the similar typeclass 🙂

src/Chapter3.hs
```diff
@@ -971,6 +1114,22 @@ implement the following functions:
 🕯 HINT: to implement this task, derive some standard typeclasses
 -}
 
+data Weekday = Sunday | Monday | Tuesday | Wednesday | Thursday | Friday | Saturday
+            deriving (Show, Eq, Ord, Bounded, Enum)
+
+isWeekend :: Weekday -> Bool
+isWeekend day = day == minBound || day == maxBound
```
vrom911, line 1121: This is clever 🧠

src/Chapter3.hs
```diff
@@ -971,6 +1114,22 @@ implement the following functions:
...
+
+nextDay :: Weekday -> Weekday
+nextDay day = if day == maxBound
+              then minBound
+              else succ day
+
+daysToParty :: Weekday -> Int
+daysToParty day = if day > Friday
+                  then (fromEnum day - fromEnum Friday) + fromEnum Friday
+                  else fromEnum Friday - fromEnum day
```
vrom911, line 1131: That is right!
You can actually simplify it a bit. If you use the fact that there are 7 days in the week, and the way how `mod` works on negatives, then it could be done easier:

```haskell
>>> mod (-1) 7
6
```

src/Chapter3.hs
```diff
@@ -1006,6 +1165,175 @@ Implement data types and typeclasses, describing such a battle between two
 contestants, and write a function that decides the outcome of a fight!
 -}
 
+-- Data Types
+
+-- Actions - all possible actions
+data Action = AttackFoe | CastSpell Int | DrinkHealthPotion Int | RunAway
+                deriving (Show)
```
vrom911, line 1172: To make it more typesafe, you could have two separate data types for `Knight` actions and `Monster` actions 🙂

src/Chapter3.hs
```diff
@@ -1006,6 +1165,175 @@ Implement data types and typeclasses, describing such a battle between two
...
+
+-- Knight
+data Knight = Knight { knightName :: String
+                     , knightAttack :: Int
+                     , knightHealth :: Int
+                     , knightDefense ::Int
+                     , knightActions :: [Action]          
+                     } deriving (Show)
+
+-- smart constructor for Knight that enforces the correct subset of actions
```
vrom911, line 1195: If you have two separate data types here, it makes sense to separate actions as well. then it would be safe to work with constructors itself without this smart constructor function 👍🏼

src/Chapter3.hs
```diff
@@ -1006,6 +1165,175 @@ Implement data types and typeclasses, describing such a battle between two
...
+  nextAction monster = 
+    let
+      actions = monsterActions monster
+      (maybeAction, actions') = nextAction' actions
+    in (maybeAction, monster { monsterActions = actions' })
+
+-- Typeclass helper functions
+nextAction' :: [Action] -> (Maybe Action, [Action])
+nextAction' [] = (Nothing, [])
+nextAction' (a:as) = (Just a, as ++ [a])
```
vrom911, line 1289: Adding the element at the end of the list is not very efficient.

src/Chapter3.hs
```diff
@@ -1006,6 +1165,175 @@ Implement data types and typeclasses, describing such a battle between two
...
+  where
+    (action, k') = nextAction k
+
+-- Monster implementation of nextRound
+roundMonster :: Fighter b => (Monster, b) -> (Monster, b)
+roundMonster (m, f) =
+  case action of
+    Nothing         -> (m, f)
+    Just AttackFoe  -> (m', calculateDamage f (getAttack m))
+    Just RunAway    -> (m, f)
```
vrom911, line 1308: When the monster runs away, it means that the battle is over and knight wins. 🙂

src/Chapter3.hs
```diff
@@ -1006,6 +1165,175 @@ Implement data types and typeclasses, describing such a battle between two
...
+  | getHealth f2 <= 0 = Just (getName f1)
+  | otherwise         = Nothing
+
+-- Fight
+-- First make sure at least one fighter has actions
+-- Then fight a round and check if there's a winner. If not, recurse
+fight :: (Fighter a, Fighter b) => a -> b -> String
+fight fighter1 fighter2 =
+  if (noAction fighter1) && (noAction fighter2)
+    then "No fight"
```
vrom911, line 1326: When this could happen?

vrom911 (approved):
Very huge work on this chapter! Very creative and interesting solutions 👏🏼 🎉

pennychase:
Thanks for the review. There was a lot of helpful advice on style and idiomatic usage.

PR #179, opened 2020-10-12
Solved Chapter 1

nasirhm:
### Solutions for Chapter 1

Thank you very much for the awesome tutorial on FP and Haskell

cc @vrom911 @chshersh Please share your feedback on the solutions and how can I improve it too :)

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -650,7 +650,8 @@ aren't ready for this boss yet!
 -}
 
 firstDigit :: Int -> Int
-firstDigit n = let evaluated_expr = div n 10
+firstDigit n = let m = abs n
+                   evaluated_expr = div m 10
                    in (if evaluated_expr < 10 then evaluated_expr else firstDigit evaluated_expr)
 
 {-
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -475,7 +475,7 @@ Implement a function that returns the last digit of a given number.
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit :: Int -> Int
-lastDigit n = mod n 10
+lastDigit n = mod (abs n) 10
 
 
 {- |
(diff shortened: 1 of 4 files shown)


nasirhm:
Fixed breaking tests and warnings.

src/Chapter1.hs
```diff
@@ -537,7 +539,11 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | (x >= y && x <= z) || (x >= z && x <= y) = x
+    | (y >= x && y <= z) || (y >= z && y <= x) = y
+    | otherwise = z
```
vrom911, line 546: Perfect solution with the minimal amount of guard cases 👍🏼

src/Chapter1.hs
```diff
@@ -614,9 +623,13 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n = let last2 = mod (div (abs n) 10) 10
+                 last1 = mod (abs n) 10
```
vrom911, line 628: Great one!
You can see that this `last1` is actually the `lastDigit` function from the previous task.

Also, I notice that you use

```haskell
mod (abs n) 10
```
and 
```haskell
div (abs n) 10
```

```haskell
(a, b) = divMod (abs n) 10
```

src/Chapter1.hs
```diff
@@ -636,8 +649,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit n = let m = abs n
+                   evaluated_expr = div m 10
```
vrom911, line 654: in Haskell, it is the standard to use camel case instead of the snake case 🐍 🐫 
```suggestion
                   evaluatedExpr = div m 10
```

vrom911 (approved):
Very well done 👏🏼 Congrats with finishing Chapter One 🥳 
Very nice solutions and thoughtful work!

chshersh:

PR #182, opened 2020-10-12
Chapter 2 Solution

khvmaths:
### Solutions for Chapter 2

Loving the chapter! Thank you so much 😊  

cc @vrom911 @chshersh


src/Chapter2.hs
```diff
@@ -740,7 +747,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: a -> [[a]] -> [[a]]
+contains n = filter (elem n)
```
chshersh, line 751: I'm afraid, this don't compile. `elem` tests values on equality, and you can't compare values of any type without telling the compiler so. A proper fix would be add `Eq` constraint:

```suggestion
contains :: Eq a => a -> [[a]] -> [[a]]
contains n = filter (elem n)
```

but we explain typeclasses and constraints only in the next chapter. For now, it would be enough to just use some monomorphic type like `Int`:


```suggestion
contains :: Int -> [[Int]] -> [[Int]]
contains n = filter (elem n)
```

src/Chapter2.hs
```diff
@@ -780,13 +788,13 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
--- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: a -> [a] -> [a]
+listElementsLessThan x = filter (< x)
 
--- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [a] -> [a] -> [a]
+pairMul = zipWith (*)
```
chshersh, line 797: The same comment about type signatures applies to these functions as well. They won't compile without constraints, so it's better to use some specific types at the moment

src/Chapter2.hs
```diff
@@ -841,7 +849,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: a -> [a] -> [a]
+rotate n l
+  | n < 0 || null l = []
+  | otherwise = drop n (take (length 1+n) (cycle l))
```
chshersh, line 855: That's a correct solution! Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -857,9 +868,13 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
-
+rewind :: [a] -> [a]
+rewind = go []
+  where
+    go a [] = a
+    go a (x:xs) = go (x:a) xs
```
chshersh, line 875: Absolutely correct implementation :+1:

chshersh (approved):
Congrats on finishing Chapter Two!
And great work on the Advanced Task :clap: 

src/Chapter2.hs
khvmaths: I see. Thanks for the reply. I always taught that i can just let compiler to choose, but I overlooked them. Thanks again.

chshersh:

PR #184, opened 2020-10-12
Solutions for Chapter 1

rahulbhargav:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -510,7 +510,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = if (abs x) < (abs y) then x else y
+closestToZero x y = if abs x < abs y then x else y
 
 
 {- |
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -417,6 +417,7 @@ task is to specify the type of this function.
 49
 -}
 
+squareSum :: Int -> Int -> Int
 squareSum x y = (x + y) * (x + y)
 
 
@@ -437,7 +438,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next x = x + 1
 
 {- |
 After you've implemented the function (or even during the implementation), you
@@ -478,8 +479,9 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
 
+lastDigit :: Integral a => a -> a
+lastDigit n = mod n 10
 
 {- |
 =⚔️= Task 6
@@ -508,7 +510,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if (abs x) < (abs y) then x else y
 
 
 {- |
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -209,31 +209,30 @@ So, the output in this example means that 'False' has type 'Bool'.
...
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
-
```
vrom911, line 217: You need to keep this empty line in order to have tests pass 🙂

src/Chapter1.hs
```diff
@@ -478,8 +489,9 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
 
+lastDigit :: Integral a => a -> a
+lastDigit n = mod n 10
```

src/Chapter1.hs
```diff
@@ -556,8 +572,15 @@ True
...
 -}
-isVowel c = error "isVowel: not implemented!"
 
+isVowel :: Char -> Bool
+isVowel c
+  | c == 'a' || c == 'A' = True
+  | c == 'e' || c == 'E' = True
+  | c == 'i' || c == 'I' = True
+  | c == 'o' || c == 'O' = True
+  | c == 'u' || c == 'U' = True
```
vrom911, line 582: Nice catch on the uppercase letters! 😮

src/Chapter1.hs
```diff
@@ -620,8 +643,11 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n =
+  let last = n `mod` 10
+      sndLast = (n `div` 10) `mod` 10
```
vrom911, line 649: Great one!
You can see that this `last1` is actually the `lastDigit` function from the previous task.

Also, I notice that you use

```haskell
mod n 10
```
and 
```haskell
div n 10
```

```haskell
(a, b) = divMod n 10
```

src/Chapter1.hs
```diff
@@ -620,8 +643,11 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
+  in (last + sndLast)
```
vrom911, line 650: No need to put this in parenthesis 🙂 
```suggestion
  in last + sndLast
```

Also, `last` is the name for the standard function, so it is better to name variables in other names 🙂

src/Chapter1.hs
```diff
@@ -641,8 +667,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit:: Int -> Int
+firstDigit n
+  | n `div` 10 == 0 = n
```
vrom911, line 672: I see, that for a negative number like -19, this implementation will return the negative number itself instead of the last digit because the first check n < 10 will always succeed for negative numbers.

vrom911 (approved):
Nice work! Congrats on finishing Chapter One 🎉

src/Chapter1.hs
rahulbhargav: Great suggestion. I will use this. Thank you

vrom911:
The solutions look awesome 😍  I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: 
We will review your PR with solutions to chapter two next! ⏭️

PR #188, opened 2020-10-12
Solutions for Chapter 3

nbelluco:
### Solutions for Chapter, skipped Task 9 for now

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -344,6 +344,15 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+data Book = MkBook
+    { name       :: String
+    , author     :: String
+    , publisher  :: String
+    , pages      :: String
+    , general    :: String
+    , prince     :: String
+    }
+
 {- |
 =⚔️= Task 2
 
@@ -373,6 +382,42 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+
+data Knight = MkKnight
+    { knightHealth  :: Int
+    , knightAttack  :: Int
+    , knightGold    :: Int
+    }
+
+data Monster = MkMonster
+    { monsterHealth  :: Int
+    , monsterAttack  :: Int
+    , monsterGold    :: Int
+    }
+
+
+fight :: Knight -> Monster -> Int
+fight knight monster
+    | (knightAttack knight) >= (monsterHealth monster) = (knightGold knight) + (monsterGold monster)
+    -- is there any way to access records' fields in a more succint way ? like a dot notation or something ?
+    | otherwise =
+        case (knightHealth knight) < (monsterAttack monster) of
+          True -> (-1)
+          False -> (knightGold knight)
+
+perceval = MkKnight
+    { knightHealth = 50
+    , knightAttack = 30
+    , knightGold = 20
+    }
+
+dracor = MkMonster
+    { monsterHealth = 31
+    , monsterAttack = 60
+    , monsterGold = 45
+    }
+
+-- TODO
 
 {- |
 =🛡= Sum types
@@ -460,6 +505,13 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data MealType
+    = Breakfast
+    | Launch
+    | Dinner
+    | MidnightSnack
+
+
 {- |
 =⚔️= Task 4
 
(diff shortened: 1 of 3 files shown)


src/Chapter3.hs
```diff
@@ -344,6 +344,15 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+data Book = MkBook
+    { name       :: String
+    , author     :: String
+    , publisher  :: String
+    , pages      :: String
+    , general    :: String
+    , prince     :: String
```
vrom911, line 353: Because every story should have its prince 🤴🏼 👸🏼  😄

src/Chapter3.hs
```diff
@@ -374,6 +383,42 @@ after the fight. The battle has the following possible outcomes:
...
+data Monster = MkMonster
+    { monsterHealth  :: Int
+    , monsterAttack  :: Int
+    , monsterGold    :: Int
+    }
+
+
+fight :: Knight -> Monster -> Int
+fight knight monster
+    | (knightAttack knight) >= (monsterHealth monster) = (knightGold knight) + (monsterGold monster)
```
vrom911, line 401: In Haskell, you should not use parenthesis in every expression, only where necessary for the parser to separate some cases. In here, everything can be without parenthesis 🙂 
```suggestion
    | knightAttack knight >= monsterHealth monster = knightGold knight + monsterGold monster
```

src/Chapter3.hs
```diff
@@ -374,6 +383,42 @@ after the fight. The battle has the following possible outcomes:
...
+    -- is there any way to access records' fields in a more succint way ? like a dot notation or something ?
+    | otherwise =
+        case (knightHealth knight) < (monsterAttack monster) of
+          True -> (-1)
+          False -> (knightGold knight)
```
vrom911, line 406: You can write it as two other guards:
```suggestion
    | knightHealth knight < monsterAttack monster = -1
    | otherwise = knightGold knight
```

src/Chapter3.hs
```diff
@@ -374,6 +383,42 @@ after the fight. The battle has the following possible outcomes:
...
+    { knightHealth = 50
+    , knightAttack = 30
+    , knightGold = 20
+    }
+
+dracor = MkMonster
+    { monsterHealth = 31
+    , monsterAttack = 60
+    , monsterGold = 45
+    }
```
vrom911, line 418: Nice examples! 👍🏼 ⚔️

src/Chapter3.hs
```diff
@@ -480,7 +532,81 @@ After defining the city, implement the following functions:
...
+data Center
+    = Church
+    | Library
+
+data House = MkHouse
+    { people  :: Int }
+
+data City = MkCity
+    { cityCastle  :: Castle
+    , cityWall    :: Wall
```
vrom911, line 555: This is the correct data type, that can represent what we have in requirements 👍🏼 

 You already have `Castle` that stores walls inside. 

Let's look at the task:

1. Castle with a name
3. No castle

This looks like a sum type!:

```haskell
data Castle 
    = Castle String
    | WallCastle String
    | None
```

src/Chapter3.hs
```diff
@@ -480,7 +532,81 @@ After defining the city, implement the following functions:
...
+
+-- tried to give both Wall and Castle a None alternative but the compilare complained
+-- about multiple declarations of None, is there any way to bypass this ?
+
+data Center
+    = Church
+    | Library
+
+data House = MkHouse
+    { people  :: Int }
```
vrom911, line 551: That is a nice solution, however, it allows having any amount of people in one house, however, task specifies only a few cases of possible outcomes. You probably need another sum type to represent this 🙂

src/Chapter3.hs
```diff
@@ -480,7 +532,81 @@ After defining the city, implement the following functions:
...
+    { cityCastle  :: Castle
+    , cityWall    :: Wall
+    , cityCenter :: Center
+    , cityHouses  :: [House]
+    }
+
+
+countPeople :: [House] -> Int
+countPeople [] = 0
+countPeople [x] = (people x)
```
vrom911, line 563: This case is asctually covered by the next one.

Also, you can notice that it is actually `sum` function + standard `map` functions 🙂

src/Chapter3.hs
```diff
@@ -560,22 +686,32 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defense (armor * dexterity)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage damage) (Defense defense) (Health health) = Health (health + defense - damage)
```
vrom911, line 714: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -907,10 +1094,24 @@ Implement instances of "Append" for the following types:
...
+
+
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append (Gold x) (Gold y) = Gold (x + y)
+
+
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append listOne listTwo = listOne ++ listTwo
```
vrom911, line 1112: You can eta-reduce the implementation of `append` for brevity. So you can see how easy it is to create instances :slightly_smiling_face: 
```suggestion
    append = (++)
```

src/Chapter3.hs
```diff
@@ -971,6 +1172,36 @@ implement the following functions:
...
+
+nextDay :: WeekDay -> WeekDay
+nextDay day
+    | day == maxBound = minBound
+    | otherwise = succ day
+
+
+daysToParty :: WeekDay -> Int
+daysToParty day =
+    length (enumFromTo day Sunday) - 2
```
vrom911, line 1202: Interesting solution!

However, it would be faster if you use `mod` function instead of calculating the length of days.

vrom911 (approved):
Great solution! Very well done 👍🏼

PR #189, opened 2020-10-12
resolve chapter 2

krivtsov:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -361,7 +361,7 @@ Implement a function that returns only the first half of a given list.
 firstHalf :: [a] -> [a]
 firstHalf l = halfList where
   lengthList = length l
-  half = div lengthList 2 
+  half = div lengthList 2
   halfList = take half l
 
 
@@ -621,7 +621,7 @@ Implement a function that duplicates each element of the list
 -}
 duplicate :: [a] -> [a]
 duplicate [] = []
-duplicate (x: xs) = [x,x] ++ duplicate(xs)
+duplicate (x: xs) = [x,x] ++ duplicate xs
 
 
 {- |
@@ -745,7 +745,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = concat (map ((\x -> replicate x x)) l)
+smartReplicate l = concatMap (\x -> replicate x x) l
 
 {- |
 =⚔️= Task 9
@@ -759,7 +759,7 @@ the list with only those lists that contain a passed element.
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
 contains :: Int -> [[Int]] -> [[Int]]
-contains n = filter (\x -> elem n x)
+contains n = filter (elem n)
 
 
 {- |
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,42 +136,52 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
+[True, False] :: [Bool]
 
 
 String is a list of characters:
 >>> :t "some string"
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +346,7 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList start end range = if start < 0 then [] else take (end - start + 1) (drop start range)
```
chshersh, line 349: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding a separate case to your check :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -605,7 +620,8 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate [] = []
+duplicate (x: xs) = [x,x] ++ duplicate xs
```
chshersh, line 624: It is more efficient to construct lists using the cons operator directly without introducing intermediate values.
```suggestion
duplicate (x:xs) = x : x : duplicate xs
```

src/Chapter2.hs
```diff
@@ -620,7 +636,9 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+takeEven [] = []
+takeEven (x : xs) = x : takeEven (drop 1 xs)
```
chshersh, line 641: Nice usage of standard function `drop`! Though, you can pattern-match on the first two elements of the list first and avoid using `drop` at all

src/Chapter2.hs
```diff
@@ -727,7 +745,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concatMap (\x -> replicate x x) l
```
chshersh, line 748: Now, after you've mastered eta-reduction, you can apply such a technique to other functions as well  🙂 

src/Chapter2.hs
```diff
@@ -841,7 +861,9 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate _ [] = []
+rotate n l = if n < 0 then [] else drop n (take (length l + n) (cycle l))
```
chshersh, line 866: Unfortunately, `cycle` fails in runtime on empty lists :recycle: So you need to handle this case separately for this function to work properly.

Also, your current solution does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -857,7 +879,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [Int] -> [Int]
+rewind [] = []
+rewind (x: xs) = rewind xs ++ [x]
```
chshersh, line 884: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
Well done with chapter two tasks!
And bonus points for solving the Advanced Task :100: 

I left some comments and suggestions in my review :slightly_smiling_face:

chshersh:

PR #190, opened 2020-10-13
Chapter1 Push

danielcpham:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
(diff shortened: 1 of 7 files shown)


src/Chapter1.hs
```diff
@@ -417,9 +417,9 @@ task is to specify the type of this function.
 49
 -}
 
+squareSum :: Num a => a -> a -> a
```
vrom911, line 420: We explain `Num` in the following chapters, so you are one step ahead 😸

src/Chapter1.hs
```diff
@@ -478,7 +479,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit ::  Int -> Int
+lastDigit n = abs n `mod` 10
```
vrom911, line 483: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1:

src/Chapter1.hs
```diff
@@ -541,8 +543,12 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | min x y == min y z = min x z
+  | otherwise = max (min x y) (min y z)
```
vrom911, line 549: Creative solution! 🧑‍🎨 
As you reuse some of the `max` and `min` values, it makes sense to move them into separate variables in the `where` clause, e.g. `minXy = min x y` , etc. 🙂

src/Chapter1.hs
```diff
@@ -619,9 +632,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n =
+  let ones = mod (abs n) 10
+      tens = mod (div (abs n - ones) 10) 10
```
vrom911, line 638: Great!

vrom911, line 638: You can see that this `ones` is actually the `lastDigit` function from the previous task.

Also, I notice that you use

```haskell
mod (abs n) 10
```
and 
```haskell
div (abs n) 10
```

```haskell
(a, b) = divMod (abs n) 10
```

src/Chapter1.hs
```diff
@@ -641,7 +656,9 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+-- firstDigit n = if div n 10 == 0 then n else firstDigit (div n 10)
```
vrom911, line 660: Yeap, this is almost right. You only need to think of the case of the negative numbers. For that, you would need to use `abs` function 👍🏼

vrom911 (approved):
Wonderful start! Congrats with finishing Chapter One 🎉

PR #191, opened 2020-10-13
Solution for Chapter 3

khvmaths:
### Solutions for Chapter 3

Thank you for the fun! I have learnt a lot, even though the syntax is very strange to me, but somehow I have learnt to adopt them. haha

cc @vrom911 @chshersh


src/Chapter3.hs
```diff
@@ -1006,7 +1122,78 @@ Implement data types and typeclasses, describing such a battle between two
...
 -}
 
-
+--Assuming data for each player (monster/knight)
+newtype PlayerHealth = PlayerHealth Int
+newtype PlayerAttack = PlayerAttack Int
+newtype PlayerDefense = PlayerDefense Int
+
+--All possible actions
+data BattleAction = Attack | DrinkPotion Int | CastSpell Int | MonsterAttack | MonsterRun
```
chshersh, line 1131: Ideally, you want to have two separate data type `FighterAction` and `MonsterAction` so it won't be possible to mix `MonsterRun` and `DrinkPotion` in a single list

chshersh (approved):
Nice work on Chapter Three!
And well done with the Advanced Task :top: 

Left some comments for code improvements, and general suggestions. Looks good overall, but there are a few things regarding the language syntax. I understand that the Haskell syntax may feel weird and look very unusual, especially when coming from other languages... But at the same time, I really like it, because it simple and concise :slightly_smiling_face:

chshersh:

PR #193, opened 2020-10-13
Do chapter 1

cherealnice:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -522,7 +522,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = (if abs x < abs y then x else y)
+closestToZero x y = if abs x < abs y then x else y
 
 {- |
 =⚔️= Task 7
@@ -576,7 +576,7 @@ False
 -}
 isVowel :: Char -> Bool
 isVowel c
-  | elem c vowels = True
+  | c `elem` vowels = True
   | otherwise = False
   where vowels = ['a', 'e', 'i', 'o', 'u']
 
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -429,6 +429,7 @@ task is to specify the type of this function.
 49
 -}
 
+squareSum :: Num a => a -> a -> a
 squareSum x y = (x + y) * (x + y)
 
 
@@ -559,8 +560,8 @@ mid :: Ord a => a -> a -> a -> a
 mid x y z
   | x < y && x < z = min y z
   | y < x && y < z = min x z
-  | z < x && z < y = min x y
-  | otherwise = x
+  | otherwise = min x y
+  | otherwise = z
 
 {- |
 =⚔️= Task 8
@@ -575,11 +576,7 @@ True
 False
 -}
 isVowel :: Char -> Bool
-isVowel c
-  | c `elem` vowels = True
-  | otherwise = False
-  where vowels = ['a', 'e', 'i', 'o', 'u']
-
+isVowel c = c `elem` ['a', 'e', 'i', 'o', 'u']
 
 {- |
 == Local variables and functions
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -478,7 +490,9 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Integral a => a -> a
+lastDigit n = mod absn 10
+  where absn = abs n
```
chshersh, line 495: Nice usage of `where`!

src/Chapter1.hs
```diff
@@ -542,7 +555,12 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Ord a => a -> a -> a -> a
+mid x y z
+  | x < y && x < z = min y z
+  | y < x && y < z = min x z
+  | z < x && z < y = min x y
+  | otherwise = x
```
chshersh, line 563: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

It looks like you can move last 2 cases under a single otherwise and return some `min` in there

src/Chapter1.hs
```diff
@@ -556,7 +574,11 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+  | c `elem` vowels = True
+  | otherwise = False
+  where vowels = ['a', 'e', 'i', 'o', 'u']
```
chshersh, line 581: Nice one! 👍🏼 

```haskell
```

src/Chapter1.hs
```diff
@@ -620,7 +642,13 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Integer -> Integer
+sumLast2 n = ones + tens
+  where
+    absn = abs n
+    moded = mod absn 100
+    ones = mod moded 10
+    tens = div moded 10
```
chshersh, line 651: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
```

src/Chapter1.hs
```diff
@@ -641,7 +669,12 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Integral p => p -> p
+firstDigit 0 = 0
+firstDigit n = last result
+  where
+    absn = abs n
+    result = [div absn (10 ^ x) | x <- [0..10], absn >= (10 ^ x)]
```
chshersh, line 677: Interesting solution with list comprehensions :brain:

chshersh (approved):
Congrats on finishing Chapter One!
And well done with the advanced task :clap: 

I left my review and some comments with suggestions to further code improvements :slightly_smiling_face:

src/Chapter1.hs
cherealnice: Sounds good. Interesting that explicit pattern matching is faster 🤔

chshersh: @cherealnice The future version of GHC actually will implement such optimization to automatically rewrite `elem` with hardcoded list to manual pattern matching. But it was discovered recently, that `elem` is much slower, and rewriting it to pattern matching in one of the Haskell projects significantly improved the performance of application, because such `elem` was called a lot in there.

chshersh:
Your solutions to chapter one look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face:

PR #196, opened 2020-10-13
chapter 1

jovyntls:
### Solutions for Chapter 1

cc @vrom911 @chshersh 

Thank you for the comprehensive guide!

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -213,8 +213,7 @@ True :: Bool
 >>> :t 'a'
 'a' :: Char
 >>> :t 42
-42 :: Int
-[actual result: `42 :: Num p => p`]
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -542,7 +543,10 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid x y z
+    | x < y && x < z = min y z
+    | y < x && y < z = min x z
+    | z < x && z < y = min x y
```
chshersh, line 549: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

It looks like you can move the last case under a single otherwise and return some `min` in there

src/Chapter1.hs
```diff
@@ -641,7 +649,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit n
+  | div n 10 == 0 = n
```
chshersh, line 653: I'm afraid, this function hangs on negative numbers because of how `div` behaves on them:

```haskell
ghci> div  (-1) 10
-1
```

Also, remember to add top-level signatures for functions!

chshersh (approved):
Congrats with finishing Chapter One! :tada: 
And well done with the Advanced task :+1: 

I left some comments on possible code improvements, but your code already looks nice :slightly_smiling_face:

jovyntls:
@chshersh Thank you for all the feedback! I'll keep those in mind :)

vrom911:

PR #199, opened 2020-10-13
Do chapter 2

cherealnice:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -644,11 +644,11 @@ Write a function that takes elements of a list only on even positions.
 [2,3,4]
 -}
 takeEven :: [a] -> [a]
-takeEven l = go l
+takeEven = go
   where
     go :: [a] -> [a]
     go [] = []
-    go (x:[]) = [x]
+    go [x] = [x]
     go (x:_:xs) = x : takeEven xs
 
 {- |
@@ -756,7 +756,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = concat $ map (\x -> replicate x x) l
+smartReplicate l = concatMap (\x -> replicate x x) l
 
 {- |
 =⚔️= Task 9
@@ -770,7 +770,7 @@ the list with only those lists that contain a passed element.
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
 contains :: (Foldable t, Eq a) => a -> [t a] -> [t a]
-contains num = filter (\x -> isJust $ find (==num) x)
+contains num = filter (isJust . find (== num))
 
 {- |
 =🛡= Eta-reduction
@@ -873,7 +873,7 @@ list.
 🕯 HINT: Use the 'cycle' function
 -}
 rotate :: Int -> [a] -> [a]
-rotate offset list = subList offset ((length list) + (offset - 1)) $ cycle list
+rotate offset list = subList offset (length list + (offset - 1)) $ cycle list
 
 {- |
 =💣= Task 12*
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -783,10 +812,12 @@ divideTenBy :: Int -> Int
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Ord a => a -> [a] -> [a]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Integer] -> [Integer] -> [Integer]
+pairMul = zipWith (*)
```
vrom911, line 820: Eta-reduction award 🏆

src/Chapter2.hs
```diff
@@ -841,7 +872,8 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate offset list = subList offset (length list + (offset - 1)) $ cycle list
```
vrom911, line 876: Interesting solution with `subList`! 💯 

Unfortunately, `cycle` fails in runtime on empty lists :recycle: So you need to handle this case separately for this function to work properly.

Also, even hat's a correct solution, though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -857,7 +889,8 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind list = map (list !!) [length list - 1, length list - 2..0]
```
vrom911, line 893: Creative solution! 🎨 

Though it is slow as `!!`, `length` walks each element of the list to get its size, so on each step, you will go through all the elements, which will take a lot! And imagine making both of them for each element at every step.
Also `!!` is not a safe function and can fail in runtime if the function will be outside the length of the list.

That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Great job! Chapter Two is done ✔️ Great work 👏🏼

src/Chapter2.hs
```diff
@@ -500,7 +521,9 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird42 :: (Eq a, Num a) => [a] -> Bool
+isThird42 (_:_:42:_) = True
+isThird42 _ = False;
```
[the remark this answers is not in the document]
cherealnice: Oops! Habit...

src/Chapter2.hs
```diff
@@ -620,7 +643,13 @@ Write a function that takes elements of a list only on even positions.
...
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven = go
+  where
+    go :: [a] -> [a]
+    go [] = []
+    go [x] = [x]
+    go (x:_:xs) = x : takeEven xs
```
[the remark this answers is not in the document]
cherealnice: Yea I wanted to use `go` pattern here, but maybe it wasn't necessary :)

src/Chapter2.hs
cherealnice: Thanks for the suggestions! I took some more time for this one, and might have cheated a bit but I understand it a lot better now :D

src/Chapter2.hs
cherealnice: Thanks for the suggestion, I took another shot at it

vrom911:
The solutions for this Chapter look awesome 😍 
I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to open PRs with solutions to other chapters!

PR #201, opened 2020-10-14
Chapter2 Push

danielcpham:

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +336,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList start stop list 
+  | start < 0 = []
+  | otherwise = take (stop - start + 1) (drop start list)
```
vrom911, line 341: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding a separate case to your check :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -620,7 +626,13 @@ Write a function that takes elements of a list only on even positions.
...
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven (x:xs) = takeEvenHelper True (x:xs)
+  where
+    takeEvenHelper :: Bool -> [a] -> [a]
+    takeEvenHelper _ [] = []
+    takeEvenHelper isEven (y:ys) = if isEven then y : takeEvenHelper False ys else takeEvenHelper True ys
```
vrom911, line 635: Very creative solution with a helper function! 🎨 

src/Chapter2.hs
```diff
@@ -727,9 +739,13 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+-- smartReplicate [] = []
+-- smartReplicate (x:xs) = replicate x x ++ smartReplicate xs
+-- smartReplicate xs = foldr (\ x -> (++) (replicate x x)) [] xs
+smartReplicate = foldr (\ x -> (++) (replicate x x)) []
```
vrom911, line 745: This is right! 💯 

```haskell
[3,1,2,4,0,1]
```

src/Chapter2.hs
```diff
@@ -780,13 +797,13 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
-divideTenBy x = div 10 x
+divideTenBy = div 10
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul = zipWith (*)
```
vrom911, line 806: Eta-reduction award 🏆

src/Chapter2.hs
```diff
@@ -780,13 +797,13 @@ Let's now try to eta-reduce some of the functions and ensure that we
 mastered the skill of eta-reducing.
 -}
 divideTenBy :: Int -> Int
-divideTenBy x = div 10 x
+divideTenBy = div 10
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan x = filter (< x)
```
vrom911, line 803: We advise adding type signatures to all functions 🙂

src/Chapter2.hs
```diff
@@ -841,7 +858,12 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate num xs
+  | num < 0 = []
+  | num == 0 = xs
+  | otherwise = take (length xs) (drop num (cycle xs))
```
vrom911, line 865: Nice catch with `num == 0` 💯 

Note that you can also check `mod num (length xs)` to operate with smaller numbers in the last clause 🙂

src/Chapter2.hs
```diff
@@ -857,7 +879,8 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind = foldl (flip (:)) []
```
vrom911, line 883: Wow, that's a clever solution using standard functions! 🤯

vrom911 (approved):
Great job! Chapter Two is done as well! Very tidy and thoughtful solutions 👏🏼 
And the advanced task is just perfect, bravo 🎉

PR #202, opened 2020-10-14
Solutions for Chapter 2

jovyntls:

cc @vrom911 @chshersh

Thanks again! This guide has been very helpful and instructive 😄

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,45 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
+[] :: [a]
 
 
 Append two lists:
 >>> :t (++)
+(++) :: [a] -> [a] -> [a]
 
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -620,7 +625,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+takeEven (x:y:xs) = x : takeEven xs
```
chshersh, line 629: Since the name of variable `y` doesn't matter here, you can use the `_` underscore pattern :slightly_smiling_face: 
```suggestion
takeEven (x:_:xs) = x : takeEven xs
```

src/Chapter2.hs
```diff
@@ -727,7 +735,8 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate (x:xs) = replicate x x ++ smartReplicate xs
```
chshersh, line 738: This is right! 💯 

```haskell
[3,1,2,4,0,1]
```

src/Chapter2.hs
```diff
@@ -740,7 +749,7 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains n ls = filter (\x -> elem n x) ls
```
chshersh, line 752: Now, after you've mastered eta-reduction, you can apply such a technique to other functions as well  🙂 

Here you have two opportunities for eta-reduction: lambda function and the last argument to `filter`

src/Chapter2.hs
```diff
@@ -841,7 +850,11 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate n l = 
+  let llen = length l
+      ls = take (2 * llen) (cycle l)
```
chshersh, line 856: I think you don't need to take `2 * llen` elements first. `cycle` returns an infinite list, so if you drop from it and take, then you'll have a finite list in the end

src/Chapter2.hs
```diff
@@ -841,7 +850,11 @@ list.
...
+  in take llen (drop n ls)
```
chshersh, line 857: That's a correct solution! Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Otherwise, if `n > 2 * llen`, `drop` returns an empty list.

Maybe you can think of a way to optimize this and handle these special cases :slightly_smiling_face:  :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -857,7 +870,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind (x:xs) = rewind xs ++ [x]
```
chshersh, line 874: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
Congrats on solving Chapter Two Tasks!
And nice work on the advanced task :medal_sports: 

I left some comments and suggestions on how to improve code further :slightly_smiling_face:

vrom911:

PR #204, opened 2020-10-14
Chapter 1

kopackiw:
### Solutions for Chapter 1

cc @vrom911 @chshersh

Really nice introduction, thanks 🙇

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -641,7 +657,13 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Integer -> Integer
+firstDigit n
+      | isLastDigit = n
+      | otherwise   = firstDigit currentN
```
kopackiw, line 663: I didn't have better name for `currentN`. What is a usual convention when dealing with something being "current" during folding?

PR #205, opened 2020-10-15
chapter 1 completed

piushkumar:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -437,7 +438,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next x = x+1
```
vrom911, line 441: Great 👍🏼 
In Haskell operator and arguments are usually separated with spaces:
```suggestion
next x = x + 1
```

src/Chapter1.hs
```diff
@@ -477,9 +478,8 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
-
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
```
vrom911, line 482: Your implementation is almost correct :ok: 

src/Chapter1.hs
```diff
@@ -508,8 +508,12 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
-
+--closestToZero x y = min (abs (x - 0)) (abs (y - 0))
+closestToZero x y = if (abs(x-0) < abs(y-0))
```
vrom911, line 512: You can omit subtracting 0 as it gives no difference when working with `Int`s 🙂 

Also you can remove parenthesis in the `if` condition:
```suggestion
closestToZero x y = if abs x < abs y
```

src/Chapter1.hs
```diff
@@ -556,7 +563,14 @@ True
...
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+        | (c == 'a' || c == 'A') = True
+        | (c == 'e' || c == 'E') = True
+        | (c == 'i' || c == 'I') = True
+        | (c == 'o' || c == 'O') = True
+        | (c == 'u' || c == 'U') = True
```
vrom911, line 572: `()` are redundant here. I see how it can be confusing to see when to use them and when not to use, especially in the beginning. But it's usually better to remove redundant brackets for cleaner code :broom:

src/Chapter1.hs
```diff
@@ -619,8 +633,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+  let lastDigit = mod n 10
+      secLastDigit = mod (div n 10) 10
```
vrom911, line 639: Great one!
You can see that this `lastDigit` here is actually the `lastDigit` function from the previous task.
Also, the negative case in here as well 🙂 

Also, I notice that you use

```haskell
mod x 10
```
and 
```haskell
div x 10
```

```haskell
(a, b) = divMod x 10
```

src/Chapter1.hs
```diff
@@ -640,8 +657,10 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n 
+             | ((div n 10) == 0) = n
```
vrom911, line 662: I see that for a negative number like -19, this implementation will return the negative number instead of the last digit 👍🏼

vrom911 (approved):
Great start! Congrats with finishing Chapter One 🎉

vrom911:
The solutions for Chapter One look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to open PRs with solutions to other chapters!

PR #210, opened 2020-10-16
Finished Chapter1

ramkumarr:
### Solutions for Chapter 1

cc @vrom911 @chshersh

The only annoying thing for me, in this chapter, is this warning for Task 7
```haskell
src/Chapter1.hs:547:1: warning: [-Wincomplete-patterns]
    Pattern match(es) are non-exhaustive
    In an equation for ‘mid’: Patterns not matched: _ _ _
    |
547 | mid x y z
    | ^^^^^^^^^...
```

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -417,6 +417,7 @@ task is to specify the type of this function.
 49
 -}
 
+squareSum :: Num a => a -> a -> a
```
vrom911, line 420: We explain `Num` in the following chapters, so you are one step ahead 😸

src/Chapter1.hs
```diff
@@ -477,8 +478,8 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = (abs n) `mod` 10
```
vrom911, line 482: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1: 

```suggestion
lastDigit n = abs n `mod` 10
```

src/Chapter1.hs
```diff
@@ -542,7 +543,14 @@ value after "=" where the condition is true.
...
 
-mid x y z = error "mid: not implemented!"
+mid :: Ord a => a -> a -> a -> a
+mid x y z
+        | x <= y && y <= z = y
+        | z <= y && y <= x = y
+        | y <= x && x <= z = x
+        | z <= x && x <= y = x
+        | x <= z && z <= y = z
+        | y <= z && z <= x = z
```
vrom911, line 553: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

Because of that, you will need to add another guard – `| otherwise`, to tell the compiler, that your pattern matching is exhaustive 🙂
However, in your case, you rightly noticed, that you covered all cases. So actually if the first 4 cases won't fit, then the answer would definitely be `z`. That means that you can move all these 2 last cases under a single otherwise and return z in here.
```suggestion
        | otherwise = z
```

src/Chapter1.hs
```diff
@@ -556,7 +564,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+        | elem c "aeiouAEIOU" = True
+        | otherwise = False
```
vrom911, line 570: Nice one! 👍🏼 

```haskell
isVowel c = elem c "aeiouAEIOU"
```

src/Chapter1.hs
```diff
@@ -620,7 +631,11 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = lastD + secondlastD
+  where
+    lastD = (abs n) `mod` 10
```
vrom911, line 637: Great!
You can notice that this is actually the `lastDigit` function from the previous task 🙂

src/Chapter1.hs
```diff
@@ -620,7 +631,11 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
+    secondlastD = ((abs n) `mod` 100) `div` 10
```
vrom911, line 638: Here you also don't need parenthesis with infixly specified functions :)

src/Chapter1.hs
```diff
@@ -641,7 +656,8 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n = read [head $ show (abs n)]
```
vrom911, line 660: Creative solution 🧠 🎨

vrom911 (approved):
Awesome! Congrats with finishing Chapter One 🎉

vrom911:

PR #212, opened 2020-10-16
Chapter 4 exercises complete

pennychase:
### Solutions for Chapter {4}

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -671,9 +671,9 @@ Can you implement a monad version of AND, polymorphic over any monad?
 -}
 
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM b1 b2 = b1 >>= (\x ->
-             b2 >>= (\y ->
-             pure (x && y)))
+andM b1 b2 = b1 >>= (\x -> if x == False
+                           then pure x
+                           else b2 >>= (\y -> pure y))
 
 {- |
 =🐉= Task 9*: Final Dungeon Boss
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -114,22 +114,30 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
+Char :: *
 
 >>> :k Bool
+Bool :: *
 
 >>> :k [Int]
+[Int] :: *
 
 >>> :k []
+[] :: * -> *
 
 >>> :k (->)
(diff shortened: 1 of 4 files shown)


src/Chapter4.hs
```diff
@@ -472,10 +487,12 @@ Implement the Applicative instance for our 'Secret' data type from before.
 -}
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    Trap x <*> _ = Trap x
+    Reward f <*> y = fmap f y
```
chshersh, line 494: Nice work on reusing `Functor` instance for `Secret`! :world_map:

src/Chapter4.hs
```diff
@@ -489,6 +506,25 @@ Implement the 'Applicative' instance for our 'List' type.
...
 
+instance Applicative List where
+  pure :: a -> List a
+  pure x = Cons x Empty
+
+  (<*>) :: List (a -> b) -> List a -> List b
+  -- Original version
+  -- Empty <*> _ = Empty
+  -- (Cons f fs) <*> xs = (fmap f xs) `myAppend` (fs <*> xs)
+  fs <*> xs = myConcat $ fmap (\f -> fmap f xs) fs
```
chshersh, line 517: Your original version is already good enough :+1: 
But the new one is clever :brain:

src/Chapter4.hs
```diff
@@ -628,8 +669,11 @@ Can you implement a monad version of AND, polymorphic over any monad?
 
 🕯 HINT: Use "(>>=)", "pure" and anonymous function
 -}
+
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM b1 b2 = b1 >>= (\x ->
+             b2 >>= (\y ->
+             pure (x && y)))
```
chshersh, line 676: The ordinary `&&` if lazy on the second argument. Meaning that if it meets `False` as the first argument, there is no need to calculate another arg. In `andM` we would like to keep the same behaviour, so the

```
Just False `andM` _
```

Your current solution returns `Nothing` on `andM (Just False) Nothing`

chshersh (approved):
:tada: Fantastic work!

PR #214, opened 2020-10-16
Fix typos in chapter 2.

graceful-potato:
Fix 2 small typos.

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -259,7 +259,7 @@ values of variables you defined before.
 
 Let's talk a bit about list implementation details. Lists in Haskell
 are implemented as __linked lists__ (or cons-lists). And because
-everything is Haskell is immutable, adding elements at the beginning
+everything in Haskell is immutable, adding elements at the beginning
 of the lists is cheap. Haskell doesn't need to allocate new memory and
 copy the whole list there; it can just create a new list from a new
 element and a pointer to an already existing list. In other words,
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -401,7 +401,7 @@ always matches (the same as a variable), but we don't use its value.
   the breakpoint, 'case' in Haskell always returns only a single
   expression for a single branch. You can think of this process as
   trying to match all patterns from the first one to the last one and
-  returning the expression on the right side of "=" only for the pattern
+  returning the expression on the right side of "->" only for the pattern
```
vrom911, line 404: Now when I look closer I see that we are not talking about `case-of` in here, but about _casing_ in Haskell as in the example above.
Maybe putting that instead of the case above would make it clearer.
We are talking about `cafe-of` with arrows in the following paragraph 🙂

src/Chapter2.hs
```diff
@@ -401,7 +401,7 @@ always matches (the same as a variable), but we don't use its value.
   the breakpoint, 'case' in Haskell always returns only a single
```
vrom911, line 401:
```suggestion
  the breakpoint, _casing_ in Haskell always returns only a single
```

vrom911 (commented):
Thanks @graceful-potato !I notice one thing there, sorry that I didn't spot it before 🙂

src/Chapter2.hs
graceful-potato: Oh, I got what you mean, but is `casing` a good term? What you think about `pattern matching on function parameters`?

vrom911: Yeap, that is much better, thank you! 👍🏼 ❤️

graceful-potato:
Updated

vrom911 (approved):
Great! Thank you 🤗

PR #215, opened 2020-10-16
Chapter2

ramkumarr:
### Solutions for Chapter 1

cc @vrom911 @chshersh

I enjoyed the higher order functions and lists.
Coming from Erlang, lists are subtly different in Haskell, even without considering the infinite lists/lazy evaluation etc. Nice exercises :-)

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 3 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +349,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList x y lst
+    | x < 0 || y < 0 = []
```
chshersh, line 353: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding a separate case to your check :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -605,8 +622,11 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate xs = duplicate' xs []
 
+duplicate' :: [a] -> [a] -> [a]
+duplicate' [] acc = reverse acc
+duplicate' (x:xs) acc = duplicate' xs (x:x:acc)
```
chshersh, line 629: Your implementation is correct :+1: But it can be made more performant. `reverse` is not a fast function, since it traverses the whole list. Here it's possible to slightly rewrite the implementation by adding elements to the beginning of the result instead adding to a separate variable `acc`.

src/Chapter2.hs
```diff
@@ -620,7 +640,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+-- takeEven [] = []
+takeEven (x0:_:xs) = (x0 : takeEven xs)
+takeEven lst = lst
```
chshersh, line 646: Since this function is rather small, I would recommend adding patterns for the remaining cases explicitly. Just to make sure that GHC pattern checker does it work properly, as well as for self-checking
```suggestion
takeEven [] = []
takeEven [x] = [x]
```

src/Chapter2.hs
```diff
@@ -727,7 +750,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concat $ map (\x -> replicate x x) l
```
chshersh, line 753: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -841,7 +867,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate n lst
+    | n < 0 = []
+    | otherwise = take (length lst) $ drop n $ cycle lst
```
chshersh, line 873: That's a correct solution! Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -857,7 +886,12 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind lst = rewind' lst []
+
+rewind' :: [a] -> [a] -> [a]
+rewind' [] acc = acc
+rewind' (x:xs) acc = rewind' xs (x:acc)
```
chshersh, line 894: Usually, instead of creating a separate top-level function and taking names from the global namespace, it's recommended to create local functions inside the `where` clause. In most cases, this function is called `go` and it's known as the "Recursive go" pattern.

chshersh (approved):
Great work on chapter two!
And well done with the Advanced task :medal_sports: 

I left some comments and suggestions in my review :slightly_smiling_face: I'm glad, you've enjoyed this chapter :blush:

vrom911:
I'm going to close this PR, because we can't merge the solutions to our repo 👌🏼

PR #216, opened 2020-10-17
Solutions for Chapter 1

pislas:
### Solutions for Chapter {1}
I find this way of learning Haskell very interesting. Thank you very much
cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -403,6 +403,7 @@ ghci> :q
 --
 -- >>> squareSum 3 4
 -- 49
+squareSum :: Int -> Int -> Int 
 squareSum x y = (x + y) * (x + y)
 
 -- |
@@ -529,7 +530,7 @@ mid x y z
   | x >= z && x <= y = x
   | y >= x && y <= z = y
   | y >= z && y <= x = y
-  | otherwise = zs
+  | otherwise = z
 
 -- |
 -- =⚔️= Task 8
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -403,7 +403,7 @@ ghci> :q
 --
 -- >>> squareSum 3 4
 -- 49
-squareSum :: Int -> Int -> Int 
+squareSum :: Int -> Int -> Int
 squareSum x y = (x + y) * (x + y)
 
 -- |
@@ -465,7 +465,7 @@ next x = x + 1
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit :: Int -> Int
 --lastDigit n = error "lastDigit: Not implemented!"
-lastDigit n = mod n 10
+lastDigit n = mod (abs n) 10
 
 -- |
 -- =⚔️= Task 6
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -70,52 +64,53 @@ Each Haskell module starts with the "module <MODULE_NAME> where" line.
...
-Additionally to static typing, Haskell has __type inference__. This means that
-you _don't need_ to specify the type of each expression as it is going to be
-found out for each expression and subexpression by the powerful compiler.
-
-However, you are __strongly encouraged to write top-level function type
-signatures__ and provide types in different situations where you don't
-immediately see what types will be inferred.
--}
+-- | This tutorial uses block comments to explain various concepts and provide
+-- task description.
```
chshersh, line 69: Would it be possible for you to configure your autoformatter not to change the style of comments?
We still can review, but it's easier for us when the diff is only for solutions and not for the whole text.

src/Chapter1.hs
```diff
@@ -187,462 +181,462 @@ ghci> :q
...
-0
-
-The 'if-then-else' constructions must always have both __then__ and __else__
-branches because it is an expression and it must always return some value.
-
-👩‍🔬 Due to lazy evaluation in Haskell, only the expression from the branch
-  satisfying the check will be returned and, therefore, evaluated.
--}
+lastDigit :: Int -> Int
+--lastDigit n = error "lastDigit: Not implemented!"
```
chshersh, line 467: You can safely remove these lines :broom: 
```suggestion
```

src/Chapter1.hs
```diff
@@ -187,462 +181,462 @@ ghci> :q
...
+--   Remember, that Haskell is indentation- and layout-sensitive language.
+--
+-- Casual reminder about adding top-level type signatures for all functions :)
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | x >= y && x <= z = x
+  | x >= z && x <= y = x
+  | y >= x && y <= z = y
+  | y >= z && y <= x = y
+  | otherwise = z
```
chshersh, line 533: Nice and clean implementation!

src/Chapter1.hs
```diff
@@ -187,462 +181,462 @@ ghci> :q
...
-5
-
-You need to use recursion in this task. Feel free to return to it later, if you
-aren't ready for this boss yet!
--}
-
-firstDigit n = error "firstDigit: Not implemented!"
-
+    beforeLast :: Int -> Int
+    beforeLast x = lastDigit (div ((abs x) - lastDigit x) 10)
```
chshersh, line 617: Great!

chshersh (approved):
Congratulations on finishing chapter one!
And well done with the Advanced Task :top: 

I left some comments about your solutions. Mostly just minor feedback, the solutions already look good :+1:

src/Chapter1.hs
pislas: Sorry, it was a mistake. Thank you for your patience @chshersh

src/Chapter1.hs
pislas: Thanks for your feedback @chshersh , I will try to solve the following chapters!!

PR #218, opened 2020-10-17
Chapter 1 solutions

jovyntls:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -479,7 +479,7 @@ Implement a function that returns the last digit of a given number.
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit :: Int -> Int
-lastDigit n = mod n 10
+lastDigit n = let m = abs n in mod m 10
 
 
 {- |
@@ -627,9 +627,10 @@ specifying complex expressions.
 
 sumLast2 :: Int -> Int
 sumLast2 n =
-  let last = mod n 10
-      last2 = div (mod n 100) 10
-  in last + last2
+  let m = abs n
+      last1 = mod m 10
+      last2 = div (mod m 100) 10
+  in last1 + last2
 
 
 {- |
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -508,7 +509,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if (abs x) <= (abs y) then x else y
```
vrom911, line 512: Great one 👍🏼 
`()` are redundant here. I see how it can be confusing to see when to use them and when not to use, especially in the beginning. But it's usually better to remove redundant brackets for cleaner code :broom: 
```suggestion
closestToZero x y = if abs x <= abs y then x else y
```

src/Chapter1.hs
```diff
@@ -620,7 +625,12 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+  let m = abs n
+      last1 = mod m 10
```
vrom911, line 631: Neat solition! 👌🏼 
You can also notice that this is the `lastDigit` function from the previous task 🐱

src/Chapter1.hs
```diff
@@ -640,8 +650,12 @@ Implement a function that returns the first digit of a given number.
...
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+  | n < 0 = firstDigit (-n)
+  | div n 10 == 0 = n
+  | div n 100 == 0 = div n 10
+  | otherwise = firstDigit (div n 10)
```
vrom911, line 658: I think these two cases could be collapsed into the only last one 🙂

vrom911 (approved):
Awesome work! Congrats with finishing Chapter One! 🎉

vrom911:
@jovyntls I have just noticed that you already had PR with this Chapter (#196 ), is this one opened by mistake?

vrom911:

PR #219, opened 2020-10-17
Do chapter 3

cherealnice:
### Solutions for Chapter 3
cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -53,6 +53,8 @@ provide more top-level type signatures, especially when learning Haskell.
 
 module Chapter3 where
 
+import Data.Tuple (swap)
+
 {-
 =🛡= Types in Haskell
 
@@ -343,6 +345,18 @@ Define the Book product data type. You can take inspiration from our description
 of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
+data Book = MkBook
+  { bookTitle    :: String
+  , bookAuthor   :: String
+  , bookGenre    :: String
+  } deriving (Show)
+
+goneWithTheWind :: Book
+goneWithTheWind = MkBook
+  { bookTitle = "Gone With The Wind"
+  , bookAuthor = "Some Old Racist Lady"
+  , bookGenre = "Old Timey"
+  }
 
 {- |
 =⚔️= Task 2
@@ -373,6 +387,25 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+
+data Character = MkCharacter
+  { characterHealth   :: Int
+  , characterAttack   :: Int
+  , characterGold     :: Int
+  } deriving (Show)
+
+fight :: Character -> Character -> Character
+fight monster knight
+  | knightAttack >= monsterHealth = knight { characterGold = knightGold + monsterGold }
+  | monsterAttack >= knightHealth = knight { characterHealth = -1}
+  | otherwise = knight
+  where
+    knightAttack = characterAttack knight
+    monsterHealth = characterHealth monster
+    knightGold = characterGold knight
+    monsterGold = characterGold monster
+    monsterAttack = characterAttack monster
+    knightHealth = characterHealth knight
 
 {- |
 =🛡= Sum types
@@ -460,6 +493,13 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data MealType
+  = Breakfast
+  | Lunch
+  | Dinner
+  | Brunch
+  | Snack
+
 {- |
 =⚔️= Task 4
 
@@ -479,6 +519,65 @@ After defining the city, implement the following functions:
    complicated task, walls can be built only if the city has a castle
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
+
+data Building
+  = Church
+  | Library deriving (Show)
+
+data HousePeople
+  = P1
+  | P2
+  | P3
+  | P4 deriving (Show)
+
+peopleCount :: HousePeople -> Int
+peopleCount P1 = 1
+peopleCount P2 = 2
+peopleCount P3 = 3
+peopleCount P4 = 4
+
+data House = House { housePeople :: HousePeople } deriving (Show)
+
+data Wall = Wall deriving (Show)
+
+data City = City
+  { cityCastleName   :: String
+  , cityWalls        :: [Wall]
+  , cityBuildingType :: Building
+  , cityHouses       :: [House]
+  } deriving (Show)
+
+buildCity :: Building -> [House] -> [Char] -> City
+buildCity bulidingType houses "" = City
+  { cityBuildingType = bulidingType
+  , cityHouses       = houses
+  , cityWalls        = []
+  , cityCastleName   = ""
+  }
+buildCity bulidingType houses castleName = City
+  { cityBuildingType = bulidingType
+  , cityHouses       = houses
+  , cityWalls        = [Wall]
+  , cityCastleName   = castleName
+  }
+
+buildCastle :: City -> String -> City
+buildCastle city castleName = city { cityCastleName = castleName }
+
+buildHouse :: City -> House -> City
+buildHouse city house = city { cityHouses = house:cityHouses city }
+
+buildWalls :: City -> City
+buildWalls city = city { cityWalls = Wall:cityWalls city }
+
+canBuildWall :: City -> Bool
+canBuildWall city = hasEnoughPeople city && hasCastle city
+
+hasCastle :: City -> Bool
+hasCastle = (> 0) . length . cityCastleName
+
+hasEnoughPeople :: City -> Bool
+hasEnoughPeople = (> 10) . sum . map peopleCount . map housePeople . cityHouses
 
 {-
 =🛡= Newtypes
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -374,6 +388,25 @@ after the fight. The battle has the following possible outcomes:
 
 -}
 
+data Character = MkCharacter
+  { characterHealth   :: Int
+  , characterAttack   :: Int
+  , characterGold     :: Int
+  } deriving (Show)
+
+fight :: Character -> Character -> Character
```
chshersh, line 397: The only issue with the current implementation and reusages of the same data type for `Monster` and `Knight` in here is that you can easily mix them when calling the function:

```haskell
> fight knight monster
> fight monster knight
```

Both of the calls are absolutely valid, but the results are going to be different! Because your implementation expects a particular order of arguments, but it's so easy to provide a wrong order.

```haskell
```

src/Chapter3.hs
```diff
@@ -480,6 +520,67 @@ After defining the city, implement the following functions:
...
+
+newtype House
+  = House { housePeople :: HousePeople }
+  deriving Show
+
+data Wall = Wall deriving (Show)
+
+data City = City
+  { cityCastleName   :: String
+  , cityWalls        :: [Wall]
```
chshersh, line 547: This is the correct data type, that can represent what we have in requirements 👍🏼 

src/Chapter3.hs
```diff
@@ -480,6 +520,67 @@ After defining the city, implement the following functions:
...
+  }
+
+buildCastle :: City -> String -> City
+buildCastle city castleName = city { cityCastleName = castleName }
+
+buildHouse :: City -> House -> City
+buildHouse city house = city { cityHouses = house:cityHouses city }
+
+buildWalls :: City -> City
+buildWalls city = city { cityWalls = Wall:cityWalls city }
```
chshersh, line 573: This function should build a wall only if all criteria are satisfied. But from the implementation, I see that it always builds walls unconditionally. I see that you've implemented separate helper functions, which is nice :+1: Looks like you can reuse them here :slightly_smiling_face:

src/Chapter3.hs
```diff
@@ -561,21 +662,36 @@ introducing extra newtypes.
...
+
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense armor dexterity = MkDefense
+  { defenseValue = armorValue armor * dexterityValue dexterity }
+
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit damage defense health = MkHealth
+  { healthValue = newHealthValue }
+  where
+    newHealthValue = healthValue health + defenseValue defense - damageValue damage
```
chshersh, line 694: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -910,6 +1038,22 @@ Implement instances of "Append" for the following types:
 class Append a where
     append :: a -> a -> a
 
+newtype Gold = Gold { goldCount :: Int } deriving (Show)
+
+instance Append Gold where
+  append :: Gold -> Gold -> Gold
+  append g1 g2 = Gold { goldCount = sum $ map goldCount [g1,g2] }
```
chshersh, line 1045: It's a nice implementation and a good way of using standard functions :+1: Though, allocating an intermediate list can be a bit inefficient. Usually, it's a good idea to implement something simple when possible :slightly_smiling_face: 
```suggestion
  append (Gold g1) (Gold g2) = Gold (g1 + g2)
```

Or are you wanted to use `$` here just to increase the amount of money? :smile:

src/Chapter3.hs
```diff
@@ -910,6 +1038,22 @@ Implement instances of "Append" for the following types:
...
+instance Append [a] where
+  append :: [a] -> [a] -> [a]
+  append = (++)
+
+instance (Append a) => Append (Maybe a) where
+    append :: Maybe a -> Maybe a -> Maybe a
+    append (Just x) (Just y)  = Just (append x y)
+    append (Just x) Nothing   = Just x
+    append Nothing (Just y)   = Just y
+    append Nothing Nothing    = Nothing
```
chshersh, line 1056: This is an absolutely correct solution!

src/Chapter3.hs
```diff
@@ -971,6 +1115,26 @@ implement the following functions:
...
+  = Monday
+  | Tuesday
+  | Wednesday
+  | Thursday
+  | Friday
+  | Saturday
+  | Sunday deriving (Show, Eq, Enum)
+
+isWeekend :: WeekDay -> Bool
+isWeekend day = any ((== True) . (== day)) [Saturday,Sunday]
```
chshersh, line 1128: `== True` is redundant call here because `== day` already returns `Bool` :slightly_smiling_face:

src/Chapter3.hs
```diff
@@ -971,6 +1115,26 @@ implement the following functions:
...
+
+nextDay :: WeekDay -> WeekDay
+nextDay Sunday = Monday
+nextDay d = succ d
```
chshersh, line 1132: That is right!

```suggestion
  | day == maxBound = minBound
```

src/Chapter3.hs
```diff
@@ -971,6 +1115,26 @@ implement the following functions:
...
+
+daysToParty :: WeekDay -> Int
+daysToParty Friday = 0
+daysToParty d = 1 + daysToParty (nextDay d)
```
chshersh, line 1136: Your implementation is correct :+1: And nice way of using already implemented `nextDay`!

Though, this implementation can be a bit more efficient, if you use the `fromEnum` method of `Enum` typeclass

src/Chapter3.hs
```diff
@@ -1006,6 +1170,143 @@ Implement data types and typeclasses, describing such a battle between two
...
+data Action
+  = Attack
+  | DrinkPotion
+  | CastSpell
+  | RunAway deriving (Show)
+
+data Monster = Monster
+  { monsterHealth  :: Int
+  , monsterAttack  :: Int
+  , monsterActions :: [Action]
```
chshersh, line 1182: By the task defintion, `Monster` can't cast spells or drink potions. So it would be better to have two separate types `KnightAction` and `MonsterAction` to avoid having invalid states

src/Chapter3.hs
```diff
@@ -1006,6 +1170,143 @@ Implement data types and typeclasses, describing such a battle between two
...
+
+  doAction :: (Fighter b) => Monster -> b -> Action -> (Monster, b)
+  doAction monster opponent RunAway =
+    ( monster
+    , opponent
+    )
+  doAction monster opponent Attack =
+    ( monster
+    , takeDamage opponent $ monsterAttack monster
+    )
```
chshersh, line 1229: Also, because of how you structured the `Action` type, this function is partial. If a `Monster` by accident have a `CastSpell` action in their sequence, this function fails in runtime and it might be unpleasant to debug :disappointed:

src/Chapter3.hs
```diff
@@ -1006,6 +1170,143 @@ Implement data types and typeclasses, describing such a battle between two
...
+
+  doActions :: (Fighter a) => Monster -> a -> (Monster, a)
+  doActions monster opponent = head $ map (doAction monster opponent) $ monsterActions monster
+
+instance Fighter Knight where
+  takeDamage :: Knight -> Int -> Knight
+  takeDamage knight amount = knight
+    { knightHealth = knightHealth knight - amount + knightDefense knight
+    , knightDefense = maximum [knightDefense knight - amount, 0]
```
chshersh, line 1238: Since it's only two values, you can use a standard function `max`
```suggestion
    , knightDefense = max 0 (knightDefense knight - amount)
```

src/Chapter3.hs
```diff
@@ -1006,6 +1170,143 @@ Implement data types and typeclasses, describing such a battle between two
...
+    , takeDamage opponent $ knightAttack knight
+    )
+  doAction knight opponent DrinkPotion =
+    ( heal knight $ knightPotion knight
+    , opponent
+    )
+  doAction knight opponent CastSpell =
+    ( applySpell knight $ knightSpell knight
+    , takeDamage opponent $ knightAttack knight
+    )
```
chshersh, line 1269: To avoid repeating top-level function definition in such cases, it's usually recommended to use the `case-of` expression:

```haskell
doAction knight opponent action = case action of
    Attack -> ...
    DrinkPotion -> ...
    ...
```

src/Chapter3.hs
```diff
@@ -1006,6 +1170,143 @@ Implement data types and typeclasses, describing such a battle between two
...
+
+makeMonster :: MonsterHealth -> MonsterAttack -> Monster
+makeMonster health attack = Monster
+  { monsterHealth  = monsterHealthValue health
+  , monsterAttack  = monsterAttackValue attack
+  , monsterActions = [Attack, RunAway]
+  }
+
+-- Return Int for which player won
+doFight :: (Fighter p1, Fighter p2) => p1 -> p2 -> Int
```
chshersh, line 1301: `Int` is a too wide type to represent this information. I think it's better to introduce a custom sum type:

```haskell
data BattleResult
    = First
    | Second
```

chshersh (approved):
Congrats on finishing Chapter Three!
And well done on the advanced task :crossed_swords: :shield: 

I left some comments with suggestions and other thoughts. I hope you find it helpful :slightly_smiling_face:

src/Chapter3.hs
cherealnice: Nice! I was wondering if this was possible 👍

src/Chapter3.hs
cherealnice: Thanks! I made an update, not sure if I handled `Either` properly. Please let me know :)

chshersh:
@cherealnice I reviewed your updated solutions. Look good :+1: I'm going to close this PR, but feel free to open another PR for the Final Chapter :slightly_smiling_face:

PR #221, opened 2020-10-17
Chapter2 solutions

gantu:
### Solutions for Chapter 2

cc @vrom911 @chshersh
Thank you for awesome tutorial.

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -337,7 +337,7 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList a b l | b - a < 0 = []
+subList a b l | b < a = []
               | a < 0 || b < 0 = []
               | otherwise = take ((b - a) + 1) $ drop a l
 
@@ -630,7 +630,8 @@ takeEven = go []
   where
     go :: [Int] -> [Int] -> [Int]
     go acc [] = acc
-    go acc (x:xs) = if even x then go (acc ++ [x]) xs else go acc xs
+    go acc [x] = x : acc
+    go acc (x : _ : xs) = go (x : acc) xs
 
 {- |
 =🛡= Higher-order functions
@@ -737,7 +738,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = concat $ map (\x -> replicate x x) l
+smartReplicate = concatMap (\ x -> replicate x x)
 
 {- |
 =⚔️= Task 9
@@ -853,6 +854,7 @@ list.
 -}
 rotate :: Int -> [a] -> [a]
 rotate n l | n < 0 = []
+           | null l = []
            | otherwise = drop r $ take n' (cycle l) 
               where
                 s = length l
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -40,6 +40,7 @@ Now, if you are ready, bring it on!
 
 module Chapter2 where
 
+import Data.List
 {-
 =🛡= Imports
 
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +337,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList a b l | b - a < 0 = []
```
chshersh, line 340: I think subtraction is redundant here, and you can just move `a` to the right side of `a` :slightly_smiling_face: 
```suggestion
subList a b l | b < a = []
```

src/Chapter2.hs
```diff
@@ -620,7 +625,12 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+takeEven = go []
+  where
+    go :: [Int] -> [Int] -> [Int]
+    go acc [] = acc
+    go acc (x:xs) = if even x then go (acc ++ [x]) xs else go acc xs
```
chshersh, line 633: Usually, adding elements to the end of the list is costly. Your implementation is correct :+1: But you can try to implement this function more efficiently by using `:` instead `++`. If you can think, how to avoid adding elements to the end of the list, and instead add to the beginning, but get the same result, that would be amazing!

In general, it's always a good idea to try to replace `l ++ [x]` with `x : l` whenever possible, though, in some cases it might not be that trivial to achieve the same result. Remember the analogy with trains :bullettrain_side: :train: :train: :train:

Also, your implementation takes even elements, not _elements on even positions_. Taking only even elements can be implemented simpler as `filter even` but that's not what we wanted here :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -727,7 +737,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concat $ map (\x -> replicate x x) l
```
chshersh, line 740: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -740,8 +750,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: Eq a => a -> [[a]] -> [[a]]
+contains x = filter (elem x)
```
chshersh, line 754: Nice work on polymorphic type signatures and eta-reduced implementation! :trophy:

src/Chapter2.hs
```diff
@@ -841,7 +851,13 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate n l | n < 0 = []
+           | otherwise = drop r $ take n' (cycle l)
```
chshersh, line 856: Unfortunately, `cycle` fails in runtime on empty lists :recycle: So you need to handle this case separately for this function to work properly.

chshersh (approved):
Great solutions for Chapter Two!
And well done on the advanced task :1st_place_medal: 

I left some comments with suggestions and general feedback :slightly_smiling_face:

src/Chapter2.hs
gantu: I got your point. I will try to solve in efficient way. Thank you for detailed review.

src/Chapter2.hs
```diff
@@ -630,7 +630,8 @@ takeEven = go []
   where
     go :: [Int] -> [Int] -> [Int]
     go acc [] = acc
-    go acc (x:xs) = if even x then go (acc ++ [x]) xs else go acc xs
+    go acc [x] = x : acc
+    go acc (x : _ : xs) = go (x : acc) xs
```
chshersh, line 634: Unfortunately, by adding elements from the beginning of the argument to the beginning of the result, you get elements in the reversed order (similar to the `rewind` implementation). You need to add the beginning of the result to get in the same order. After that, you can also notice, that the variable `acc` becomes redundant :slightly_smiling_face: 
```suggestion
    go acc (x : _ : xs) = x : go acc xs
```

chshersh (approved):
I have only a minor comment. Otherwise, your solutions look good :+1:

chshersh:
@gantu Nice work! Closing this PR, but feel free to open PRs with solutions to other chapters :slightly_smiling_face:

PR #222, opened 2020-10-17
Solution for chapter 2

piushkumar:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,44 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
+[True, False] :: [Bool]
 
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -727,7 +765,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concat (map (\x -> replicate x x) l)
```
chshersh, line 768: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -740,7 +778,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains n l = filter (\x -> (elem n x)) l
```
chshersh, line 782: Now, after you've mastered eta-reduction, you can apply such a technique to this function as well  🙂 

Here you have two opportunities for eta-reduction: the last argument to `filter` and the lambda function.

src/Chapter2.hs
```diff
@@ -841,7 +882,8 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate n l = take (length l) (drop n (cycle l))
```
chshersh, line 886: It looks like this function doesn't return empty lists on negative numbers. Also, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -857,7 +899,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
chshersh, line 904: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
Well done in Chapter Two!
And kudos for solving the Advanced Task :clap: 

I left some comments with suggestions on possible code improvements and general feedback :slightly_smiling_face:

PR #225, opened 2020-10-18
Chapter 1 Solutions

Chandu-4444:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'X') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -430,14 +433,22 @@ Implement the function that takes an integer value and returns the next 'Int'.
...
 ♫ NOTE: The current function body is defined using a special function called
   "error". Don't panic, it is not broken. 'error' is like a placeholder, that
   evaluates to an exception if you try evaluating it. And it also magically fits
   every type ｡.☆.*｡. No need to worry much about "error" here, just replace the
   function body with the proper implementation.
 -}
+
 next :: Int -> Int
-next x = error "next: not implemented!"
+next x = x+1
```
vrom911, line 451: Nice! 👍🏼 
In Haskell functions and operators are always separated from arguments with spaces:
```suggestion
next x = x + 1
```

src/Chapter1.hs
```diff
@@ -477,8 +488,9 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
```

src/Chapter1.hs
```diff
@@ -508,7 +520,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = min (abs (x)) (abs(y))
```
vrom911, line 523: Very interesting solution! 💡
However, on negative numbers, it returns the absolute value of those negative numbers, note the value itself.

Also, no need to add parenthesis around one argument 🙂 And spaces should be added 👾 
```suggestion
closestToZero x y = min (abs x) (abs y)
```

vrom911 (approved):
Great start! 🎉 
I left a few suggestions 🙂 Hope you find it helpful 👍🏼

vrom911:

PR #228, opened 2020-10-18
Chapter 4 - completed

lprakashv:
### Solutions for Chapter Four

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -667,8 +667,7 @@ Can you implement a monad version of AND, polymorphic over any monad?
 
 
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM ma mb = ma >>= (\a -> fmap ((&&) a) mb)
--- andM ma mb = (fmap (\a -> (&&) a) ma) <*> mb
+andM ma mb = ma >>= (\a -> if a then mb else pure False)
 
 {- |
 =🐉= Task 9*: Final Dungeon Boss
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -114,22 +114,30 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
+Char :: *
 
 >>> :k Bool
+Bool :: *
 
 >>> :k [Int]
+[Int] :: *
 
 >>> :k []
+[] :: * -> *
 
 >>> :k (->)
+(->) :: * -> * -> *
 
 >>> :k Either
+Either :: * -> * -> *
 
 >>> data Trinity a b c = MkTrinity a b c
 >>> :k Trinity
+Trinity :: * -> * -> * -> *
 
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
+IntBox :: (* -> *) -> *
 
 -}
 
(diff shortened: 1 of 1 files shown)


src/Chapter4.hs
```diff
@@ -472,10 +488,11 @@ Implement the Applicative instance for our 'Secret' data type from before.
 -}
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    (<*>) (Trap e) _        = Trap e
+    (<*>) (Reward f) secret = fmap f secret
```
vrom911, line 495: Nice work on reusing `Functor` instance for `Secret`! :world_map:

src/Chapter4.hs
```diff
@@ -628,8 +664,11 @@ Can you implement a monad version of AND, polymorphic over any monad?
 
 🕯 HINT: Use "(>>=)", "pure" and anonymous function
 -}
+
+
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM ma mb = ma >>= (\a -> fmap ((&&) a) mb)
+-- andM ma mb = (fmap (\a -> (&&) a) ma) <*> mb
```
vrom911, line 671: The ordinary `&&` if lazy on the second argument. Meaning that if it meets `False` as the first argument, there is no need to calculate another arg. In `andM` we would like to keep the same behaviour, so the

```
Just False `andM` _
```

If you don't use the `>>=` operator, then the `Monad` constraint is too strong, and it is enough to specify only `Applicative`. However, with `Applicative` you are not able to implement a lazier version of `andM` due to the nature of the `Applicative`. `Monad` though, can stop when needed, so you can use that for laziness 🙂

src/Chapter4.hs
```diff
@@ -673,6 +712,28 @@ Specifically,
...
+    = EmptyTree
+    | Node a (Tree a) (Tree a)
+
+instance Functor Tree where
+    fmap :: (a -> b) -> Tree a -> Tree b
+    fmap _ EmptyTree    = EmptyTree
+    fmap f (Node v l r) = Node (f v) (fmap f l) (fmap f r)
+
+{-
+Assuming reversing a tree means mirroring it along a vertical axis
```
vrom911, line 725: Yes, this is absolutely correct 👍🏼

vrom911 (approved):
:tada: Fantastic work! You finished all 4 Chapters 🥳 
Thank you so much for taking this course! I hope you had fun 🎈 👯  

src/Chapter4.hs
lprakashv: Okay, a little hard to understand. But, I will try to make the lazier version of this and I guess the `pure` would be helpful with this.

lprakashv: Done

vrom911:
The solutions look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo 🙂 
Well done, and thank you very much 🤗 
Don't hesitate to reach out if you have any questions!

PR #232, opened 2020-10-19
Chapter3

ramkumarr:
### Solutions for Chapter 3 (Without the final task :-/

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -343,6 +343,10 @@ Define the Book product data type. You can take inspiration from our description
 of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
+data Book = Book { bookName   :: String,
+                   bookAuthor :: String,
+                   bookPages  :: Int,
+                   bookISDN   :: String }
 
 {- |
 =⚔️= Task 2
@@ -373,6 +377,33 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+
+data Props = Props {
+    health :: Int,
+    attack :: Int,
+    gold :: Int
+} deriving (Show)
+data Knight = Knight {
+    knightName :: String,
+    kprops :: Props
+} deriving (Show)
+data Monster = Monster {
+    monsterName :: String,
+    mprops :: Props
+} deriving (Show)
+
+fight :: Monster -> Knight -> Either Int Knight
+fight monster knight = go monster knight "knight" where
+    go :: Monster -> Knight -> String -> Either Int Knight
+    go Monster{mprops = Props{health = h, gold = g}} k@Knight{kprops = p} _
+      | h <= 0 = Right k{kprops=p{gold=gold p + g}}
+    go _ Knight{kprops = Props{health = h}} _ | h <= 0 = Left (-1)
+    go m@Monster{mprops = mp@Props{health = h}}
+       k@Knight{kprops = Props{attack = a}} "knight" =
+       go m{mprops = mp{health = h-a}} k "monster"
+    go m@Monster{mprops = Props{attack = a}}
+       k@Knight{kprops = kp@Props{health = h}} "monster" =
+       go m k{kprops = kp{health = h-a}} "knight"
 
 {- |
 =🛡= Sum types
@@ -459,6 +490,7 @@ and provide more flexibility when working with data types.
 Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
+data Breakfast = Pancake String | Cereal Int String | Pizza
 
 {- |
 =⚔️= Task 4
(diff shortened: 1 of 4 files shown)


src/Chapter3.hs
```diff
@@ -374,6 +378,33 @@ after the fight. The battle has the following possible outcomes:
...
+    go :: Monster -> Knight -> String -> Either Int Knight
+    go Monster{mprops = Props{health = h, gold = g}} k@Knight{kprops = p} _
+      | h <= 0 = Right k{kprops=p{gold=gold p + g}}
+    go _ Knight{kprops = Props{health = h}} _ | h <= 0 = Left (-1)
+    go m@Monster{mprops = mp@Props{health = h}}
+       k@Knight{kprops = Props{attack = a}} "knight" =
+       go m{mprops = mp{health = h-a}} k "monster"
+    go m@Monster{mprops = Props{attack = a}}
+       k@Knight{kprops = kp@Props{health = h}} "monster" =
+       go m k{kprops = kp{health = h-a}} "knight"
```
chshersh, line 406: Nice recursive solution!

src/Chapter3.hs
```diff
@@ -480,6 +512,31 @@ After defining the city, implement the following functions:
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
 
+data House = House Int deriving (Show) -- house with number of people in it
+data Church = Church deriving (Show)
+data Library = Library deriving (Show)
+data Castle = Castle String deriving (Show)
+data City = City {
+  castle :: Maybe Castle,
+  wall :: Bool,
```
chshersh, line 521: This is the correct data type, that can represent what we have in requirements 👍🏼 

src/Chapter3.hs
```diff
@@ -480,6 +512,31 @@ After defining the city, implement the following functions:
...
+buildHouse c@City{houses=h} num = c{houses=House num : h}
+
+buildWalls :: City -> Either String City
+buildWalls c =
+  case castle c of
+    Nothing -> Left "No castle in the city!"
+    Just _ ->
+      case foldl (\acc (House n) -> acc+n) 0 (houses c) of
+        n | n < 10 -> Left "Not enough people in the city!"
+        _ -> Right c{wall=True}
```
chshersh, line 539: That's very good modelling of the result type with sum types. Great work :+1: 

If you want to improve the code even further, you can introduce a more precise custom data types:

```haskell
data BuildWallsResult
    = NoCastle
    | NotEnoughPeople
    | NewCity City
```

With a custom data type, it's more convenient to perform unit testing of various corner cases, rather than with comparing strings.

src/Chapter3.hs
```diff
@@ -560,22 +617,30 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor a) (Dexterity d) = Defense (a*d)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage dmg) (Defense def) (Health h) = Health (h+def-dmg)
```
chshersh, line 643: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -910,6 +988,21 @@ Implement instances of "Append" for the following types:
...
     append :: a -> a -> a
 
+newtype Gold = Gold Int
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append (Gold g1) (Gold g2) = Gold (g1+g2)
+
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append lst1 lst2 = lst1 ++ lst2
```
chshersh, line 998: You can eta-reduce the implementation of `append` for brevity. So you can see how easy it is to create instances :slightly_smiling_face: 
```suggestion
    append = (++)
```

src/Chapter3.hs
```diff
@@ -910,6 +988,21 @@ Implement instances of "Append" for the following types:
...
+
+instance (Append a) => Append (Maybe a) where
+    append :: Maybe a -> Maybe a -> Maybe a
+    append Nothing Nothing = Nothing
+    append Nothing (Just x) = Just x
+    append (Just x) Nothing = Just x
```
chshersh, line 1004: This is an absolutely correct solution!

src/Chapter3.hs
```diff
@@ -971,6 +1064,38 @@ implement the following functions:
...
+nextDay Saturday = Sunday
+
+daysToParty :: WeekDay -> Int
+daysToParty Sunday = 5
+daysToParty Monday = 4
+daysToParty Tuesday = 3
+daysToParty Wednesday = 2
+daysToParty Thursday = 1
+daysToParty Friday = 0
+daysToParty Saturday = 6
```
chshersh, line 1097: These functions don't use standard functions of standard typeclasses. I would recommend looking into `Bounded` and `Enum` typeclasses, derive them and try to see what of their methods you can use :slightly_smiling_face:

chshersh (approved):
Nice work on Chapter Three so far! :+1: 

chshersh:

PR #233, opened 2020-10-19
Chapter 2 Solutions

Chandu-4444:

Hey, Thanks for this valuable tutorial. Hope you are sound and safe 😉 .
cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -40,6 +40,8 @@ Now, if you are ready, bring it on!
 
 module Chapter2 where
 
+import Data.List
+
 {-
 =🛡= Imports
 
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -621,7 +633,13 @@ Write a function that takes elements of a list only on even positions.
...
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+takeEven = go []
+  where
+    go :: [Int] -> [Int] -> [Int]
+    go acc [] = acc
+    go acc [x] = x : acc
+    go acc (x : _ : xs) = go (x : acc) xs
```
chshersh, line 642: By adding elements to `acc`, you get the list in reverse order, So you need to use the `reverse` function in the end. But there's a better solution actually. You can just add elements to the beginning of the result. Thus, you will get all elements in the proper order and don't need to use `reverse`. Morever. you can notice that the `acc` variable becomes redundant
```suggestion
takeEven = go []
  where
    go :: [Int] -> [Int] -> [Int]
    go acc [] = acc
    go acc [x] = x : acc
    go acc (x : _ : xs) = x : go acc xs
```

chshersh (approved):
Nice solutions for Chapter Two tasks so far! :+1: 
I left some comments and suggestions :slightly_smiling_face:

vrom911:

PR #234, opened 2020-10-19
Chapter 3 commit

danielcpham:
### Solutions for Chapter 3

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -343,6 +343,12 @@ Define the Book product data type. You can take inspiration from our description
 of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
+data Book = Book
+  { title           :: String,
+    author          :: String,
+    tableOfContents :: [Int]
+  }
+  deriving (Show)
 
 {- |
 =⚔️= Task 2
@@ -373,6 +379,26 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+
+data Monster = Monster
+  { monsterHealth :: Int,
+    monsterAttack :: Int,
+    monsterGold   :: Int
+  }
+  deriving (Show)
+
+data Knight = Knight
+  { knightHealth :: Int,
+    knightAttack :: Int,
+    knightGold   :: Int
+  }
+  deriving (Show)
+
+fight :: Monster -> Knight -> Int
+fight m k
+  | knightAttack k >= monsterAttack m = knightGold k + monsterGold m
+  | monsterAttack m >= knightHealth k = -1
+  | otherwise = knightGold k
 
 {- |
 =🛡= Sum types
@@ -460,6 +486,12 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data Meals
+  = Breakfast String
+  | Lunch String
+  | Dinner String
+  | Wine String
+
 {- |
 =⚔️= Task 4
 
@@ -479,7 +511,49 @@ After defining the city, implement the following functions:
    complicated task, walls can be built only if the city has a castle
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
+data House
+  = One
+  | Two
+  | Three
+  | Four
 
+houseSize :: House -> Int
+houseSize house = case house of
+  One   -> 1
+  Two   -> 2
+  Three -> 3
+  Four  -> 4
+
+data Building = Church | Library
+
+data Castle
+  = None
+  | JustCastle String
+  | CastleAndWalls String
+
+data City = City
+  { cityCastle   :: Castle,
+    cityBuilding :: Building,
+    cityHouses   :: [House]
+  }
+buildHouse :: House -> City -> City
+buildHouse newHouse city =
+  city
+    { cityHouses = newHouse : cityHouses city
+    }
+
+buildCastle :: String -> City -> City
+buildCastle newName city = case cityCastle city of
+  CastleAndWalls _ -> city {cityCastle = CastleAndWalls newName}
+  _                -> city {cityCastle = JustCastle newName}
+
+buildWalls :: City -> City
+buildWalls city = case cityCastle city of
+  JustCastle castleName ->
+    if sum (map houseSize (cityHouses city)) >= 10
+      then city {cityCastle = CastleAndWalls castleName}
+      else city
+  _ -> city
 {-
 =🛡= Newtypes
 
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -68,417 +67,482 @@ Haskell has several different ways to create entirely new data types. Let's talk
...
-
-The same definition of 'Knight' but with records should be written in the
-following way:
-
-@
-data Knight = MkKnight
-    { knightName      :: String
-    , knightVictories :: Int
+-- |
+-- =🛡= Type aliases
```
chshersh, line 71: Would it be possible for you to configure your autoformatter not to change the style of comments?
It's easier for us when the diff is only for solutions and not for the whole text :slightly_smiling_face:

danielcpham: Did y'all use a specific formatter/do you know how to turn off autoformatting comments on VS code? If not, I'll just be more conscious of what I'm committing!

chshersh: We use the [stylish-haskell](https://github.com/jaspervdj/stylish-haskell) code formatter, it's quite nice :slightly_smiling_face: But at the same time, I didn't expect VSCode plugin to enable any formatters by default :thinking:

danielcpham: I updated the PR to just include the actual code changes! While we're here, this isn't technically related, but there's a couple things that look like they should be markdown formatted (like where it says "You are on __Chapter X__"); is that able to be formatted as well, and how would I set that up in VS Code?

vrom911: Unfortunately, comments in Haskell are not in Markdown. They are using a special syntax. The docs are called Haddock and you can generate full docs out of the comments. I am not sure about the plugin, as I am not a VSCode user, but I am sure, that it would be possible to generate in there.

src/Chapter3.hs
```diff
@@ -343,6 +343,12 @@ Define the Book product data type. You can take inspiration from our description
 of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
+data Book = Book
+  { title           :: String,
+    author          :: String,
+    tableOfContents :: [Int]
```
vrom911, line 349: Interesting idea 💡 

We usually suggest to name fields with the prefixes, like `bookTitle`, so there won't be any problems with the duplicate functions in the module 🙂 But this is just FYI!

src/Chapter3.hs
```diff
@@ -374,6 +380,26 @@ after the fight. The battle has the following possible outcomes:
...
+data Knight = Knight
+  { knightHealth :: Int,
+    knightAttack :: Int,
+    knightGold   :: Int
+  }
+  deriving (Show)
+
+fight :: Monster -> Knight -> Int
+fight m k
+  | knightAttack k >= monsterAttack m = knightGold k + monsterGold m
```
vrom911, line 399: If the knight attack is bigger than monsters, it still doesn't mean, than the monster would be defeated. In fact, it is better to compare to health to see if it is enough to kill the monster ⚔️

src/Chapter3.hs
```diff
@@ -460,6 +486,12 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data Meals
+  = Breakfast String
+  | Lunch String
+  | Dinner String
+  | Wine String
```
vrom911, line 493: Cheers 🍷

src/Chapter3.hs
```diff
@@ -479,7 +511,49 @@ After defining the city, implement the following functions:
...
+  | Two
+  | Three
+  | Four
+
+houseSize :: House -> Int
+houseSize house = case house of
+  One   -> 1
+  Two   -> 2
+  Three -> 3
+  Four  -> 4
```
vrom911, line 525: Wonderful idea to have it as a sum type! This is a very nice way to use the type system to make incorrect states unrepresentable 👏🏼 
It is actually the pattern of FP programming, and you can read more about it here:

src/Chapter3.hs
```diff
@@ -561,21 +635,28 @@ introducing extra newtypes.
...
+newtype Defense = Defense Int
+
+calculatePlayerDamage :: Attack -> Strength -> Damage
+calculatePlayerDamage (Attack attack) (Strength strength) = Damage (attack + strength)
+
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defense (armor * dexterity)
+
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage damage) (Defense defense) (Health health) = Health (health + defense - damage)
```
vrom911, line 659: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -910,7 +999,20 @@ Implement instances of "Append" for the following types:
...
     append :: a -> a -> a
 
+newtype Gold = Gold Int
+instance Append Gold where
+  append :: Gold -> Gold -> Gold
+  append (Gold g1) (Gold g2) = Gold (g1 + g2)
 
+instance Append [a] where
+  append :: [a] -> [a] -> [a]
+  append l1 l2 = l1 ++ l2
```
vrom911, line 1009: You can eta-reduce the implementation of `append` for brevity. So you can see how easy it is to create instances :slightly_smiling_face: 
```suggestion
    append = (++)
```

src/Chapter3.hs
```diff
@@ -971,6 +1073,31 @@ implement the following functions:
...
+
+isWeekend :: DayOfWeek -> Bool
+isWeekend day = case day of
+  Saturday -> True
+  Sunday   -> True
+  _        -> False
+
+nextDay :: DayOfWeek -> DayOfWeek
+nextDay day
+  | day == maxBound = minBound
```
vrom911, line 1094: Great! 🎉 
Using `Bounded` in here makes code more maintainable. If you decide, for example, to have the "American" system of week days in the data type constructors, you won't need to change this function at all! 🙂

src/Chapter3.hs
```diff
@@ -971,6 +1073,31 @@ implement the following functions:
...
+  | otherwise = succ day
+
+
+daysToParty :: DayOfWeek -> Int
+daysToParty day = (fromEnum Friday - fromEnum day) `mod` 7
```
vrom911, line 1099: 🔥 Very clever 👍🏼

vrom911 (approved):
Great Chapter! Very nice and interesting structures you created here 👨🏼‍🎨

chshersh:

PR #237, opened 2020-10-20
Fix test assertion for Ch4 Task8

### LINKED ISSUE #236: Incorrect test assertion for Chapter4, Task 8
The following assertion in Chapter 4, Task 8 is incorrect, it should be `Nothing` as well

https://github.com/kowainik/learn4haskell/blob/0bdad6f127ab5d7eb357a9ca3fcca2f1dd45caf8/test/Test/Chapter4.hs#L59

CharlesSchimmel:
I know enough to be pretty sure the test was incorrect, but I did double check with `liftM2 (&&) (Just False) Nothing`.

Fixes #236


vrom911:
This is not correct.

PR #241, opened 2020-10-20
hacktoberfest Chapter1

piotrcirocki:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -489,8 +489,8 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = (abs n) `mod` 10
```
vrom911, line 493: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1: 

```suggestion
lastDigit n = abs n `mod` 10
```

src/Chapter1.hs
```diff
@@ -553,8 +555,12 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: Integer -> Integer -> Integer -> Integer 
+mid x y z =
+  let maximum = max x (max y z)
+      minimum = min x (min y z)
```
vrom911, line 561: That's a very clever implementation! :brain: 

Note that `maximum` and `minimum` are names of the standard functions. Haskell allows name shadowing, that's why it works. Though, in practice, it's usually better to avoid name clashes with other functions cause it could be confusing 🙂

src/Chapter1.hs
```diff
@@ -568,7 +574,11 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c 
+  | c `elem` ['a', 'e', 'i', 'o', 'u', 'y'] = True
+  | otherwise = False
```
vrom911, line 580: Nice one! 👍🏼 

```haskell
isVowel c = c `elem` ['a', 'e', 'i', 'o', 'u', 'y']
```

src/Chapter1.hs
```diff
@@ -631,8 +641,15 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
+sumLast2 :: Int -> Int 
+sumLast2 n = last n + penultimate n
+  where 
+    last :: Int -> Int
+    last n = (abs n) `mod` 10
```
vrom911, line 648: `last` is the name of a standard function as well, so it's better to rename this one 🙂 Though you can notice that it is actually the `lastDigit` function from the previous task 🙂

vrom911 (approved):
Nice work on solving our challenges for Chapter One! 🥳 

piotrcirocki:
> Nice work on solving our challenges for Chapter One! 🥳
> And well done with the Advanced Task 🏅
> I left my review and some comments. Hope you find them helpful 🙂

Hi, thank you for your comments. I find it very helpful during a process of learning a Haskell! 👍  Cheers!

chshersh:

PR #242, opened 2020-10-21
typos in Chapter1.hs

aelbozie:
I think these should be 1st, not trying to include this towards hacktoberfest; just correcting the typo. :)

### Solutions for Chapter 1

cc @vrom911 @chshersh


vrom911 (approved):
Thanks 💯

PR #245, opened 2020-10-22
Solve 1st, 2nd chapter, start 3rd chapter

HenriDellal:
### Solutions for Chapter 1,2

cc @vrom911 @chshersh

Hello!
First of all, thank you very much for this project! I've wanted to learn more about functional programming, but it seemed scary and difficult to me (well, it is in fact difficult, but your tutorials made it easier :D). Now I know something.
I've almost finished the third chapter, but then got ill and couldn't continue. It's been two weeks since then so I forgot some parts of Haskell and may not continue later. Still, I'd appreciate some feedback.

Regards,
Henri

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -343,7 +343,11 @@ Define the Book product data type. You can take inspiration from our description
 of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
-
+data Book = Book
+	{ bookTitle	:: String
+	, bookYear	:: Int
+	, bookDesc	:: String
+	} deriving (Show)
 {- |
 =⚔️= Task 2
 
@@ -372,6 +376,41 @@ after the fight. The battle has the following possible outcomes:
  ⊛ Neither the knight nor the monster wins. On such an occasion, the knight
    doesn't earn any money and keeps what they had before.
 
+-}
+data Knight = Knight
+	{ knightHealth	:: Int
+	, knightAttack	:: Int
+	, knightGold	:: Int
+	} deriving (Show)
+
+data Monster = Monster
+	{ monsterHealth	:: Int
+	, monsterAttack :: Int
+	, monsterGold	:: Int
+	} deriving (Show)
+
+fight :: Knight -> Monster -> Int
+fight knight monster =
+	if (monsterHealth monster - knightAttack knight <= 0)
+	then
+		knightGold knight + monsterGold monster
+	else if (knightHealth knight - monsterAttack monster <= 0)
+	then
+		-1
+	else
+		knightGold knight
+
+{-
+*Chapter3> k = Knight 30 20 20
+*Chapter3> m = Monster 40 40 29
+*Chapter3> fight k m
+-1
+*Chapter3> k2 = k { knightAttack = 50 }
+*Chapter3> fight k2 m
+49
+*Chapter3> m2 = m { monsterAttack = 20 }
+*Chapter3> fight k m2
+20
 -}
 
 {- |
@@ -459,6 +498,19 @@ and provide more flexibility when working with data types.
 Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
+data Breakfast
+	= FishAndChips
+	| Coffee
+	| Tea
+	| Toast
+	| Porridge
+	| HotDog
+	| ApplePie
+	| Sandwich
+	| KelpSalad -- not many knights like kelps, so may replace it with KrabbyPatty
+	| EggsWithBacon
+	| Lasagna
+	| Lemonade
 
 {- |
 =⚔️= Task 4
@@ -480,6 +532,41 @@ After defining the city, implement the following functions:
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
 
+data Building = Church | Library
+
+{-
+params:	Int - population
+		Building - either church or library
+		String - name of castle
+		Bool - if wall is present
+-}
+data City
+	= SmallCity Int Building
+	| CityOfWizards Int Building String Bool
+
+buildCastle :: City -> String -> City
+buildCastle c name = case c of
+	SmallCity p b -> CityOfWizards p b name False
+	CityOfWizards p b _ w -> CityOfWizards p b name w
+
+buildWalls :: City -> City
+buildWalls c = case c of
+	SmallCity _ _ -> error "Build a castle before building walls"
+	CityOfWizards p b n w ->
+		if (p >= 10)
+		then
+			if length n > 0
+			then
+				CityOfWizards p b n True
+			else
+				error "Build a castle before building walls"
+		else
+			error "Build more houses before building walls"
+
+buildHouse :: City -> City
+buildHouse c = case c of
+	SmallCity p b -> SmallCity (p+1) b
+	CityOfWizards p b n w -> CityOfWizards (p+2) b n w
 {-
 =🛡= Newtypes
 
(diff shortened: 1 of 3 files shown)


HenriDellal:
Sorry, I should have submitted it to my fork, but I forgot about that

chshersh:
Hi @HenriDellal! I'm sorry to hear about your health :disappointed: I hope you feel better now! :hugs: 

I see, that you've opened a single PR with solutions to all three chapters. To complete the Hacktoberfest, you need to submit 4 accepted PRs. Would you be willing to split your work in separate PRs to finish the Hacktoberfest contribution checklist? Or is it okay for you, and we can review this one only?

HenriDellal:
@chshersh it's fine to review it as is, thanks.

src/Chapter3.hs
```diff
@@ -480,6 +532,41 @@ After defining the city, implement the following functions:
...
+
+{-
+params:	Int - population
+		Building - either church or library
+		String - name of castle
+		Bool - if wall is present
+-}
+data City
+	= SmallCity Int Building
+	| CityOfWizards Int Building String Bool
```
chshersh, line 545: I feel like it would be more convenient to work `City` if it was a record type instead of a sum type. You already have `Castle` that stores walls inside. 

First, we need to decide with the castle. Let's look at the task:

1. Castle with a name
3. No castle

This looks like a sum type!:

```haskell
data Castle 
    = Castle String
    | WallCastle String
    | None
```

Note that your idea with castle suits here perfectly as well. But you need to define such type separately (as you did):

```haskell
data Castle = Castle 
    { castleName :: String
    , castleWall :: Bool 
    }
```

And in the city, we will have `Maybe Castle`.

So, instead of having two top-level constructors and duplicating all fields on both of them, you can define `City` as a record with optional castle:

After all that is done, Then let's build the city:

```haskell
data City = City
    -- OR you can do it with your type
    -- cityCastle :: Maybe Castle
    , cityChurchOrLibrary :: ChurchOrLibrary
    , cityHouses :: [House]
    }
```

src/Chapter3.hs
```diff
@@ -480,6 +532,41 @@ After defining the city, implement the following functions:
...
+
+buildCastle :: City -> String -> City
+buildCastle c name = case c of
+	SmallCity p b -> CityOfWizards p b name False
+	CityOfWizards p b _ w -> CityOfWizards p b name w
+
+buildWalls :: City -> City
+buildWalls c = case c of
+	SmallCity _ _ -> error "Build a castle before building walls"
```
chshersh, line 554: It's usually not a good idea to fail in runtime with `error`, because you can't easily catch such errors. The use case for `error` is rather something like `TODO` or to handle cases that are impossible to happen. But in this scenario, it's totally possible :heavy_check_mark: 

One of possible solutions to this problem is to return the original `city` without changes.

Alternatively, you can create a sum type to model all possible outcomes:

```haskell
data BuildWallsResult
    = NoCastle
    | NotEnoughPeople
    | NewCity City
```

src/Chapter3.hs
```diff
@@ -480,6 +532,41 @@ After defining the city, implement the following functions:
...
+				CityOfWizards p b n True
+			else
+				error "Build a castle before building walls"
+		else
+			error "Build more houses before building walls"
+
+buildHouse :: City -> City
+buildHouse c = case c of
+	SmallCity p b -> SmallCity (p+1) b
+	CityOfWizards p b n w -> CityOfWizards (p+2) b n w
```
chshersh, line 569: According to our task description, houses can contain from 1 to 4 people each. But in your solution it looks like each house is either for a single person or for two people :thinking:

src/Chapter3.hs
```diff
@@ -560,22 +647,26 @@ introducing extra newtypes.
...
+newtype Health = Health { healthValue :: Int }
+
 data Player = Player
-    { playerHealth    :: Int
+    { playerHealth    :: Health
     , playerArmor     :: Int
     , playerAttack    :: Int
     , playerDexterity :: Int
     , playerStrength  :: Int
     }
```
chshersh, line 660: You can go even further and introduce the newtype for each of the fields here to gain maximum benefit from the type system :slightly_smiling_face:

You can read more about the `newtype` pattern in our blog:

* https://kowainik.github.io/posts/haskell-mini-patterns#newtype

src/Chapter3.hs
```diff
@@ -752,6 +843,31 @@ parametrise data types in places where values can be of any general type.
 🕯 HINT: 'Maybe' that some standard types we mentioned above are useful for
   maybe-treasure ;)
 -}
+data Power
+	= Invisibility
+	| Berserk
+	| Freezing -- quite weird thing for a dragon
```
chshersh, line 849: Very creative powers! :art: 

And I think `Freezing` is an okay power for a dragon :snowflake: Maybe it's an Ice Dragon or Undead Dragon? :dragon: :dragon_face:

src/Chapter3.hs
```diff
@@ -907,9 +1023,34 @@ Implement instances of "Append" for the following types:
...
 
 -}
+newtype Gold = Gold { goldValue :: Int }
 class Append a where
-    append :: a -> a -> a
+	append :: a -> a -> a
+
+instance Append Gold where
+	append :: Gold -> Gold -> Gold
+	append x1 x2 = Gold ((goldValue x1) + (goldValue x2))
```
chshersh, line 1032: Your solution is correct :+1: Alternatively, you can pattern match on the `Gold` constructor directly:
```suggestion
	append (Gold x1) (Gold x2) = Gold (x1 + x2)
```

src/Chapter3.hs
```diff
@@ -907,9 +1023,34 @@ Implement instances of "Append" for the following types:
...
 
+instance Append [a] where
+	append :: [a] -> [a] -> [a]
+	append x1 x2 = x1 ++ x2
```
chshersh, line 1036: You can eta-reduce the implementation of `append` for brevity. So you can see how easy it is to create instances :slightly_smiling_face: 
```suggestion
      append = (++)
```

src/Chapter3.hs
```diff
@@ -907,9 +1023,34 @@ Implement instances of "Append" for the following types:
...
+
+instance (Append a) => Append (Maybe a) where
+	append :: Maybe a -> Maybe a -> Maybe a
+	append (Just x1) (Just x2) = Just (append x1 x2)
+	append _ _ = Nothing
```
chshersh, line 1041: While this is valid implementation and nothing in the task description says differently, in reality it would be more beneficial is this function returned `Just` when at least one argument is `Just`.

src/Chapter3.hs
```diff
@@ -970,6 +1111,44 @@ implement the following functions:
...
+
+isWeekend :: WeekDay -> Bool
+isWeekend day = case day of
+	Saturday -> True
+	Sunday -> True
+	_ -> False
+
+nextDay :: WeekDay -> WeekDay
+nextDay day = case day of
+	Sunday -> Monday
```
chshersh, line 1131: That is right!

```suggestion
    | day == maxBound = minBound
```

src/Chapter3.hs
```diff
@@ -970,6 +1111,44 @@ implement the following functions:
...
+	_ -> succ day
+
+daysToParty :: WeekDay -> Int
+daysToParty day
+	| (day > Friday) = 7 - (fromEnum day - fromEnum Friday)
+	| otherwise = fromEnum Friday - fromEnum day
+
```
chshersh, line 1138: Nice solution! If you remember from Chapter One, how `mod` works for negative numbers, you can probably implement a shorter solution :slightly_smiling_face:

src/Chapter3.hs
```diff
@@ -1006,6 +1185,71 @@ Implement data types and typeclasses, describing such a battle between two
...
+	getDamage a d = a { warriorHealth = (warriorHealth a) - getDelta a d }
+		where
+			getDelta :: Warrior -> Int -> Int
+			getDelta w dmg
+				| dmg < (warriorDefense w) = 0
+				| otherwise = dmg - (warriorDefense w)
+
+instance Beast where
+	act :: Beast -> Beast
+	act a = a
```
chshersh, line 1233: It looks like the `act` action for `Beast` doesn't handle the `RunAway` action :thinking: 
Maybe `act` function should return `Maybe a` to handle this case? Just an idea :slightly_smiling_face:

src/Chapter3.hs
```diff
@@ -1006,6 +1185,71 @@ Implement data types and typeclasses, describing such a battle between two
...
+
+
+getAttackPoints :: Fighter -> Int
+getAttackPoints f =
+	| drop (i-1) (take i (warriorActSeq f)) == Attack = (warriorAttack f) 
+	where
+		i = warriorActIndex f
+	| otherwise = 0
+
+fightResult :: Fighter -> Fighter -> [Fighter, Fighter]
```
chshersh, line 1245: It looks like this doesn't compile. `Fighter` is not a type, it's a constraint. Also, I feel like you are trying to return two fighters, but this also can be modelled better with a custom sum type.

Something like this:

```suggestion
data FightResult a b
    = FirstWins a
     | SecondWins b
     
fightResult :: (Fighter a, Fighter b) => a -> b -> FightResult a b
```

chshersh (approved):
And well done with Advanced Tasks :medal_sports: 

I left my review and some comments to all your solutions in three chapters. Hope you find them helpful :slightly_smiling_face:

vrom911:
Hope you feel better now 😷 

PR #246, opened 2020-10-22
Chapter 1

godu:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -429,7 +429,7 @@ task is to specify the type of this function.
 49
 -}
 
-squareSum :: Num a => a -> a -> a
+squareSum :: Int -> Int -> Int
 squareSum x y = (x + y) * (x + y)
 
 
@@ -493,7 +493,7 @@ Implement a function that returns the last digit of a given number.
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 
 lastDigit :: Int -> Int
-lastDigit n = (`mod` 10) $ abs n
+lastDigit n = abs n `mod` 10
 
 
 {- |
@@ -523,7 +523,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = if x*x > y*y then y else x
+closestToZero x y = if abs x > abs y then y else x
 
 
 {- |
@@ -645,12 +645,9 @@ specifying complex expressions.
 -}
 
 sumLast2 :: Int -> Int
-sumLast2 n | n < 0 = sumLast2 (-n)
 sumLast2 n = last1 + last2
   where
-    last1 = n `mod` 10
-    last2 = (n `div` 10) `mod` 10
-
+    (last2, last1) = (`divMod` 10) $ abs n `mod` 100
 
 {- |
 =💣= Task 10*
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -429,6 +429,7 @@ task is to specify the type of this function.
 49
 -}
 
+squareSum :: Num a => a -> a -> a
```

src/Chapter1.hs
```diff
@@ -490,7 +491,9 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+
+lastDigit :: Int -> Int
+lastDigit n = (`mod` 10) $ abs n
```
vrom911, line 496: You can avoid function application and parentheses if you use the function in the infix form:
```suggestion
lastDigit n = abs n `mod` 10
```

src/Chapter1.hs
```diff
@@ -520,7 +523,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if x*x > y*y then y else x
```
vrom911, line 526: That is a clever trick, but I guess `abs` would work better here, especially on huge numbers 🙂 This solution potentially could lead to the integer overflow 💥

src/Chapter1.hs
```diff
@@ -632,7 +644,12 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n | n < 0 = sumLast2 (-n)
+sumLast2 n = last1 + last2
+  where
+    last1 = n `mod` 10
+    last2 = (n `div` 10) `mod` 10
```
vrom911, line 652: That is a nice solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

I see that you have both:

```haskell
n `mod` 10
n `div` 10
```
One hint to make it shorter: the standard library has the [`divMod`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Prelude.html#v:divMod) function, that actually combines inside both `div` and `mod`. 

```haskell
(x, y) = abs n `divMod` 10
```

src/Chapter1.hs
```diff
@@ -653,8 +670,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit n | n < 0 = firstDigit (-n)
+firstDigit n | n < 10 = n
+firstDigit n = firstDigit $ n `div` 10
```
vrom911, line 676: You can use all guards under one function name. As I see you use `firstDigit n` everywhere, so it makes sense to make it shorter and not repeat that:
```suggestion
firstDigit n 
    | n < 0 = firstDigit (-n)
    | n < 10 = n
    | otherwise = firstDigit $ n `div` 10
```

vrom911 (approved):

vrom911:
The solutions look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree:

PR #247, opened 2020-10-23
PR for Chapter 3

piushkumar:
### Solutions for Chapter 3 except the last task which is still work in progress. 

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -520,35 +520,30 @@ After defining the city, implement the following functions:
    complicated task, walls can be built only if the city has a castle
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
-data Castle = Castle
-    {
-      castleName :: String
-    } deriving (Show)
+data Castle = Castle String
+            | WallCastle String
+            | None
+            deriving (Show)
 
 data ChLib =
   Church |
   Library
+  deriving (Show)
 
 data House =
   One |
   Two |
   Three |
   Four
-
+  deriving (Show)
 
 data City =
-          CityWithCastle
+          City
           {
-            cityCastle :: Castle,
-            cityWall :: Bool,
+            castle :: Castle,
             cityChurchLib :: ChLib,
             cityHouses :: [House]
-          } |
-          CityWithOutCastle
-          {
-            cityChurchLib :: ChLib,
-            cityHouses :: [House]
-          }
+          } deriving (Show)
 
 howManyInHouse :: House -> Int
 howManyInHouse house = case house of One -> 1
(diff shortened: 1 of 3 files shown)


src/Chapter3.hs
```diff
@@ -459,6 +494,12 @@ and provide more flexibility when working with data types.
 Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
+data BreakFast =
+  Pancake |
+  Bagel |
+  Eggs |
+  Sundae |
+  Omelette
```
chshersh, line 502: Yummy :yum:

src/Chapter3.hs
```diff
@@ -479,6 +520,61 @@ After defining the city, implement the following functions:
...
+            cityCastle :: Castle,
+            cityWall :: Bool,
+            cityChurchLib :: ChLib,
+            cityHouses :: [House]
+          } |
+          CityWithOutCastle
+          {
+            cityChurchLib :: ChLib,
+            cityHouses :: [House]
+          }
```
chshersh, line 551: I feel like it would be more convenient to work `City` if it was a record type instead of a sum type. You already have `Castle` that stores walls inside. 

First, we need to decide with the castle. Let's look at the task:

1. Castle with a name
3. No castle

This looks like a sum type!:

```haskell
data Castle 
    = Castle String
    | WallCastle String
    | None
```

Note that your idea with castle suits here perfectly as well. But you need to define such type separately (as you did):

```haskell
data Castle = Castle 
    { castleName :: String
    , castleWall :: Bool 
    }
```

And in the city, we will have `Maybe Castle`.

So, instead of having two top-level constructors and duplicating all fields on both of them, you can define `City` as a record with optional castle:

After all that is done, Then let's build the city:

```haskell
data City = City
    -- OR you can do it with your type
    -- cityCastle :: Maybe Castle
    , cityChurchOrLibrary :: ChurchOrLibrary
    , cityHouse s:: [House]
    }
```

Also, in your current implementation you have records inside sum types, and this is considered an anti-pattern in Haskell, because field functions are becoming partial functions.

src/Chapter3.hs
```diff
@@ -479,6 +520,61 @@ After defining the city, implement the following functions:
...
+
+howManyInHouse :: House -> Int
+howManyInHouse house = case house of One -> 1
+                                     Two -> 2
+                                     Three -> 3
+                                     Four -> 4
+
+howManyInHouses :: [House] -> Int
+howManyInHouses houses = foldr (+) 0 (map (\x-> (howManyInHouse x)) houses)
```
chshersh, line 560: `foldr (+) 0` is the standard function called `sum` and you can eta-reduce the lambda in `map` :recycle: 
```suggestion
howManyInHouses houses = sum (map howManyInHouse houses)
```

src/Chapter3.hs
```diff
@@ -589,7 +700,7 @@ hitPlayer player1 player2 =
         newHealth = calculatePlayerHit
             damage
             defense
-            (playerHealth player1)
+            (Health (playerHealth player1))
```
chshersh, line 703: If you return `Health` instead of `Int` directly in the `calculatePlayerHit` function, you don't need to wrap it here, and this function won't require any change at all!

src/Chapter3.hs
```diff
@@ -907,9 +1035,27 @@ Implement instances of "Append" for the following types:
...
 class Append a where
     append :: a -> a -> a
 
+instance Append Gold where
+  append :: Gold -> Gold -> Gold
+  append (Gold a) (Gold b) = Gold (a + b)
+
+instance Append [a] where
+  append :: [a] -> [a] -> [a]
+  append x y = (++) x  y
```
chshersh, line 1050: You can eta-reduce the implementation of `append` for brevity. So you can see how easy it is to create instances :slightly_smiling_face: 
```suggestion
  append = (++)
```

src/Chapter3.hs
```diff
@@ -907,9 +1035,27 @@ Implement instances of "Append" for the following types:
...
+
+instance (Append a) => Append (Maybe a) where
+  append :: Maybe a -> Maybe a -> Maybe a
+  append (Just x) Nothing = Nothing
+  append Nothing (Just x) = Nothing
+  append Nothing Nothing = Nothing
```
chshersh, line 1056: This is an absolutely correct solution!

src/Chapter3.hs
```diff
@@ -971,6 +1117,29 @@ implement the following functions:
...
+                  Sunday
+                  deriving (Eq,Enum,Show)
+
+isWeekend :: DayOfWeek -> Bool
+isWeekend a
+          | a == Saturday || a == Sunday = True
+          | otherwise = False
+
+nextDay :: DayOfWeek -> DayOfWeek
+nextDay Sunday = Monday
```
chshersh, line 1135: That is right!

```suggestion
    | day == maxBound = minBound
```

src/Chapter3.hs
```diff
@@ -971,6 +1117,29 @@ implement the following functions:
...
+nextDay a = succ a
+
+daysToParty :: DayOfWeek -> Int
+daysToParty Saturday = 7
+daysToParty Sunday = 6
+daysToParty a =  (1 + (fromEnum Friday))
```
chshersh, line 1141: This looks incorrect because variable `a` is not used on the right side of `=`, which means that this function returns the same number of days for all weekdays before `Friday` :thinking:

chshersh (approved):
Nice work on Chapter Three!
Let us know if you have any questions regarding the Advanced Task :slightly_smiling_face: 

As always, provided my review with some helpful comments :speaking_head:

piushkumar:
> Let us know if you have any questions regarding the Advanced Task slightly_smiling_face
@vrom911  @chshersh  Can you please provide some pointers to start with the Advanced Task 9 ? I know I would need to define the data model, but I am getting confused on data types to use for attributes.

chshersh:
> > Let us know if you have any questions regarding the Advanced Task slightly_smiling_face
> > @vrom911  @chshersh  Can you please provide some pointers to start with the Advanced Task 9 ? I know I would need to define the data model, but I am getting confused on data types to use for attributes.

Well, for this task you need to create two data types: `Monster` and `Knight`. Each type will have the properties described in the task description.

In addition, I would recommend creating separate types for `MonsterAction` and `KnightAction` to represent possible actions for each type correspondingly.

```haskell
class Fighter a where
```

```haskell

data FightResult a b
    = FirstWins a
    | SecondWins b

```

PR #250, opened 2020-10-23
Chapter3

godu:
### Solutions for Chapter 3

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -344,6 +344,12 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+data Book = Book { name :: String
+                 , author :: String
+                 , cover :: String
+                 , pages :: Int
+                 }
+
 {- |
 =⚔️= Task 2
 
@@ -373,6 +379,20 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+
+data Monster = Monster { monsterHealth :: Int
+                       , monsterAttack :: Int
+                       , monsterGold :: Int
+                       }
+data Knight = Knight { knightHealth :: Int
+                     , knightAttack :: Int
+                     , knightGold :: Int
+                     }
+
+fight :: Knight -> Monster -> Int
+fight k m | knightAttack k >= monsterHealth m = knightGold k + monsterGold m
+fight k m | knightHealth k <= monsterAttack m = - 1
+fight k _ = knightGold k
 
 {- |
 =🛡= Sum types
@@ -460,6 +480,8 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data Meal = Breakfast | Brunch | Lunch | Dinner
+
 {- |
 =⚔️= Task 4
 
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -560,22 +605,30 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defense $ armor * dexterity
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage damage) (Defense defense) (Health health) = Health $ health + defense - damage
```
chshersh, line 631: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -910,6 +968,17 @@ Implement instances of "Append" for the following types:
...
+
+instance Append Gold where
+  append (Gold a) (Gold b) = Gold $ a + b
+
+instance Append [a] where
+  append = (++)
+
+instance Append a => Append (Maybe a) where
+  append (Just a) (Just b) = return $ append a b
+  append _ _ = Nothing
```
chshersh, line 981: While this is valid implementation and nothing in the task description says differently, in reality it would be more beneficial is this function returned `Just` when at least one argument is `Just`.

src/Chapter3.hs
```diff
@@ -971,6 +1040,20 @@ implement the following functions:
...
 
+data Weekday = Mon | Tue | Wed | Thu | Fri | Sat | Sun deriving (Enum)
+
+isWeekend :: Weekday -> Bool
+isWeekend Sat = True
+isWeekend Sun = True
+isWeekend _ = False
+
+nextDay :: Weekday -> Weekday
+nextDay = succ
```
chshersh, line 1051: Unfortunately, this implementation is not exactly correctly, because `succ` fails in runtime on the last element of enumeration instead of returning the first element. You can verify this in GHCi:

```haskell
ghci> succ Sun
*** Exception: succ: bad argument
```

src/Chapter3.hs
```diff
@@ -971,6 +1040,20 @@ implement the following functions:
...
+
+daysToParty :: Weekday -> Int
+daysToParty Fri = 0
+daysToParty day = (1 +) $ daysToParty $ nextDay day
```
chshersh, line 1055: Nice recursive implementation! Though, it's possible to provide a more efficient one by using the `fromEnum` method of the `Enum` typeclass

src/Chapter3.hs
```diff
@@ -480,6 +502,23 @@ After defining the city, implement the following functions:
...
+              , wall :: Bool
+              , building :: Building
+              , houses :: [House]
+              }
+
+buildCastle :: String -> City -> City
+buildCastle castleName city = city { castle = Castle castleName }
+
+buildHouse :: Int -> City -> City
+buildHouse people city = city { houses = houses city ++ [ House people ] }
```
chshersh, line 520: According to our task, `House` can keep only from 1 to 4 people. You can use a Haskell design pattern called "smart constructor". To avoid building houses with more than 4 people you need to restrict usage of the Constructor and allow to create `House`s only with your `mkHouse` function.

Generally, in here, you can go even further and you could restrict that on type level, by using something else than `Int` 🙂

chshersh (approved):
Nice work on Chapter Three tasks! :+1: 

I left some comments with general feedback and suggestions :slightly_smiling_face:

src/Chapter3.hs
```diff
@@ -1048,11 +1063,12 @@ isWeekend Sun = True
 isWeekend _ = False
 
 nextDay :: Weekday -> Weekday
-nextDay = succ
+nextDay Sun = Mon
```
vrom911, line 1066: That is right!

```suggestion
    if day == maxBound then minBound else ...
```

src/Chapter3.hs
```diff
@@ -1048,11 +1063,12 @@ isWeekend Sun = True
...
+nextDay day = succ day
 
 daysToParty :: Weekday -> Int
 daysToParty Fri = 0
```
vrom911, line 1070: Great! No need to have this separate case anymore 🙂

chshersh:

PR #251, opened 2020-10-23
Chapter4

godu:
### Solutions for Chapter 4

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -479,9 +479,8 @@ instance Applicative (Secret e) where
     pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) (Reward f) (Reward a) = Reward $ f a
-    (<*>) (Trap e) _ = Trap e
-    (<*>) _ (Trap e) = Trap e
+    Reward f <*> x  = fmap f x
+    Trap e <*> _ = Trap e
 
 {- |
 =⚔️= Task 5
@@ -495,6 +494,17 @@ Implement the 'Applicative' instance for our 'List' type.
   type.
 -}
 
+instance Applicative List where
+  pure :: a -> List a
+  pure x = Cons x Empty
+
+  (<*>) :: List (a -> b) -> List a -> List b
+  Empty <*> _ = Empty
+  Cons f fs <*> zs = append (fmap f zs) (fs <*> zs)
+
+append :: List a -> List a -> List a
+append Empty ys = ys
+append (Cons x xs) ys = Cons x (append xs ys)
 
 {- |
 =🛡= Monad
@@ -606,8 +616,8 @@ Implement the 'Monad' instance for our 'Secret' type.
 -}
 instance Monad (Secret e) where
     (>>=) :: Secret e a -> (a -> Secret e b) -> Secret e b
-    (>>=) (Reward a) fb = fb a
-    (>>=) (Trap e) _ = Trap e
+    Reward a >>= fb = fb a
+    Trap e >>= _ = Trap e
 
 {- |
 =⚔️= Task 7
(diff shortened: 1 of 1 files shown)


src/Chapter4.hs
```diff
@@ -472,10 +476,12 @@ Implement the Applicative instance for our 'Secret' data type from before.
...
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    (<*>) (Reward f) (Reward a) = Reward $ f a
+    (<*>) (Trap e) _ = Trap e
+    (<*>) _ (Trap e) = Trap e
```
vrom911, line 484: Note that the first and the last cases look very alike the `Functor` instance 🙂 Alternatively, you could reuse it here, and have 2 cases instead of 3 👍🏼

src/Chapter4.hs
```diff
@@ -600,7 +606,8 @@ Implement the 'Monad' instance for our 'Secret' type.
 -}
 instance Monad (Secret e) where
     (>>=) :: Secret e a -> (a -> Secret e b) -> Secret e b
-    (>>=) = error "bind Secret: Not implemented!"
+    (>>=) (Reward a) fb = fb a
+    (>>=) (Trap e) _ = Trap e
```
vrom911, line 611: Do you need help with instances for other structures?

src/Chapter4.hs
```diff
@@ -629,7 +636,7 @@ Can you implement a monad version of AND, polymorphic over any monad?
 🕯 HINT: Use "(>>=)", "pure" and anonymous function
 -}
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM ma mb = (\a -> if a then  mb else return a) =<< ma
```
vrom911, line 639: Nice lazy implementation 🙂 
`>>=` is much more canonical, and in here `=<<` doesn't seem to bring any improvements in this particular case. With `>>=` you could also omit parenthesis 🙂

vrom911 (approved):
:tada: Fantastic work!
Thank you so much for taking this course! I hope you had fun 🥳  

You can also ask any questions and we can discuss any topics you would like to dig into as well  :slightly_smiling_face: And let us know if you need help with instances and the advanced task 🙂

src/Chapter4.hs
```diff
@@ -495,6 +494,17 @@ Implement the 'Applicative' instance for our 'List' type.
...
+instance Applicative List where
+  pure :: a -> List a
+  pure x = Cons x Empty
+
+  (<*>) :: List (a -> b) -> List a -> List b
+  Empty <*> _ = Empty
+  Cons f fs <*> zs = append (fmap f zs) (fs <*> zs)
+
+append :: List a -> List a -> List a
+append Empty ys = ys
```
vrom911, line 506: Perfect!

```haskell
append xs Empty = xs
```

src/Chapter4.hs
```diff
@@ -680,6 +693,21 @@ Specifically,
...
+            | Node a (Tree a) (Tree a)
+
+instance Functor Tree where
+  fmap :: (a -> b) -> Tree a -> Tree b
+  fmap _ Leaf = Leaf
+  fmap f (Node x l r) = Node (f x) (fmap f l) (fmap f r)
+
+reverseTree :: Tree a -> Tree a
+reverseTree Leaf = Leaf
+reverseTree (Node x l r) = Node x r l
```
vrom911, line 706: You need to recursively run reverse on every subtree as well in order to get the reversed tree. Now, you only change subtrees for the root 🌴

PR #252, opened 2020-10-23
Chapter2

godu:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -337,7 +337,7 @@ ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
 subList start end | start < 0 || end < 0 || end < start = const []
-subList start end  = take (end - start + 1)  . drop start
+                  | otherwise = take (end - start + 1)  . drop start
 
 {- |
 =⚔️= Task 4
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -858,7 +875,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
vrom911, line 880: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

src/Chapter2.hs
```diff
@@ -842,7 +857,9 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate x _ | x < 0 = []
```
vrom911, line 861: Guards are better when used with several cases, that would make it much more readable, also you can reuse the variables.

src/Chapter2.hs
```diff
@@ -842,7 +857,9 @@ list.
...
+rotate x xs = take (length xs) $ drop x $ cycle xs
```
vrom911, line 862: You could have a bit more efficient solution if you would check on an empty list and if `x` is zero (by module of the length as well) 🙂

vrom911 (approved):
Nice Chapter! I added a few suggestions to make your solutions faster and more canonical 🙂 👍🏼

src/Chapter2.hs
```diff
@@ -842,7 +854,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate x l | x < 0 = []
+           | null l = []
+           | otherwise = take (length xs) $ drop x $ cycle xs
```
vrom911, line 860: I guess, it is `l` in here instead of `xs` 🙂

chshersh:

PR #253, opened 2020-10-23
Chapter 1 ❤️

will-clarke:
### Solutions for Chapter 1

Thanks so much for making this - I've been really enjoying it! :) 
cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -492,7 +492,7 @@ Implement a function that returns the last digit of a given number.
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit :: Int -> Int
-lastDigit n = (abs n) `mod` 10
+lastDigit n = abs n `mod` 10
 
 
 {- |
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -632,7 +644,11 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = lastDigit' + penultimateDigit
+ where
+  lastDigit'       = abs n `mod` 10
+  penultimateDigit = (abs n `div` 10) `mod` 10
```

One hint to make your solution even shorter, I see that you have both:

```haskell
abs n `mod` 10
abs n `div` 10
```

```haskell
(x, y) = abs n `divMod` 10
```

vrom911 (approved):

src/Chapter1.hs
will-clarke: Oh yes, thank you - `divMod`'s way better!

PR #254, opened 2020-10-23
Chapter2

will-clarke:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -349,7 +364,8 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 -- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf xs = take n xs where n = length xs `div` 2
```
chshersh, line 368: Great usage of `where` construction! Makes code easily readable 💯

src/Chapter2.hs
```diff
@@ -621,7 +640,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+takeEven []           = []
+takeEven (x : _ : xs) = x : takeEven xs
+takeEven x            = x
```
chshersh, line 646: I would recommend here to pattern match on the last remaining case explicitly to make use of GHC pattern matching checker :slightly_smiling_face: Alternatively, you can combine this case with the first line and reduce the number of lines in the implementation of this function
```suggestion
takeEven [x]          = [x]
```

src/Chapter2.hs
```diff
@@ -728,7 +750,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate xs = concatMap (\ x -> replicate x x) xs
```
chshersh, line 753: Now, after you've mastered eta-reduction, you can apply such a technique to this function as well  🙂 

src/Chapter2.hs
```diff
@@ -842,7 +865,9 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate n xs | n < 0     = []
+            | otherwise = take (length xs) $ drop n $ cycle xs
```
chshersh, line 870: That's a correct solution! Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -858,7 +883,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind []       = []
+rewind (x : xs) = rewind xs ++ [x]
```
chshersh, line 888: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
Nice work on Chapter Two tasks!

I left some comments with suggestions and general feedback :slightly_smiling_face:

vrom911:

PR #255, opened 2020-10-24
Chapter4

danielcpham:
### Solutions for Chapter 4

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -114,23 +114,23 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
-
+Char :: *
 >>> :k Bool
-
+Bool :: *
 >>> :k [Int]
-
+[Int] :: *
 >>> :k []
-
+[] :: * -> *
 >>> :k (->)
-
+(->) :: * -> * -> *
 >>> :k Either
-
+Either :: * -> * -> *
 >>> data Trinity a b c = MkTrinity a b c
 >>> :k Trinity
-
+Trinity :: * -> * -> * -> *
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
-
+IntBox :: (* -> *) -> *
 -}
 
 {- |
@@ -277,10 +277,11 @@ inside or a reward with some treasure. You never know what's inside
 until opened! But the 'Functor' instance allows changing the reward
 inside, so it is quite handy.
 -}
-data Secret e a
-    = Trap e
-    | Reward a
+data Secret err treasure
+    = Trap err
+    | Reward treasure
     deriving (Show, Eq)
+
 
 
 {- |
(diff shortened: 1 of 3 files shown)


src/Chapter4.hs
```diff
@@ -471,11 +478,12 @@ Applicatives can be found in many applications:
...
 instance Applicative (Secret e) where
-    pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure :: treasure -> Secret err treasure
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    Trap err <*> _ = Trap err
+    Reward fTreasure <*> x = fmap fTreasure x
```
vrom911, line 486: Nice work on reusing `Functor` instance for `Secret`! :world_map:

src/Chapter4.hs
```diff
@@ -489,6 +497,20 @@ Implement the 'Applicative' instance for our 'List' type.
   type.
 -}
 
+append :: List a -> List a -> List a
+append Empty list = list
```
vrom911, line 501: Perfect!

```haskell
append xs Empty = xs
```

src/Chapter4.hs
```diff
@@ -629,7 +662,7 @@ Can you implement a monad version of AND, polymorphic over any monad?
 🕯 HINT: Use "(>>=)", "pure" and anonymous function
 -}
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM monadX monadY = monadX >>= \x -> if x then monadY else (pure False)
```
vrom911, line 665: Great lazy implementation 👍🏼

src/Chapter4.hs
```diff
@@ -672,6 +705,22 @@ Specifically,
...
+  fmap _ Leaf = Leaf
+  fmap f (Node parent childLeft childRight) = Node (f parent) (fmap f childLeft) (fmap f childRight)
+
+reverseTree :: Tree a -> Tree a 
+reverseTree Leaf = Leaf -- Leaf case
+reverseTree (Node parent childLeft childRight) = Node parent (reverseTree childRight) (reverseTree childLeft) -- Tree case
+
+treeToList :: Tree a -> [a]
+treeToList Leaf = []-- Leaf case
+treeToList (Node parent childLeft childRight) = parent : (treeToList childLeft) ++ (treeToList childRight) -- Tree case
```
vrom911, line 723: Great Tree functions 🌴 🌳 

You can omit parenthesis in here with operators:
```suggestion
treeToList (Node parent childLeft childRight) = parent : treeToList childLeft ++ treeToList childRight -- Tree case
```

vrom911 (approved):
:tada: Fantastic work!
Thank you so much for taking this course! I hope you had fun :man_dancing: 💃🏼  

PR #256, opened 2020-10-24
Chapter2

GalinndanGalanodel:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +336,12 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList a b xs 
+  | a < 0 || b < 0 = []
```
vrom911, line 340: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding a separate case to your check :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -534,6 +541,7 @@ For example, we can patch the previous function to count the number of
 steps we need to take in order to reduce the number to zero.
 
 🤔 Blitz question: can you guess what this number represents?
+Answer: The number of bits needed to represent this number in binary
```
vrom911, line 544: 💡 👍🏼

src/Chapter2.hs
```diff
@@ -605,8 +613,8 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
-
+duplicate [] = []
+duplicate (x:xs) = [x,x] ++ duplicate xs
```
vrom911, line 617: It is always better to use `:` instead of `++`, as it would be much more efficient 🙂 And here it is easily possible:

```suggestion
duplicate (x:xs) = x : x : duplicate xs
```

src/Chapter2.hs
```diff
@@ -620,7 +628,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven [x] = [x]
+takeEven (x:_:xs) = [x] ++ takeEven xs
```
vrom911, line 634: Here as well, you'd better to use `:` (trains analogue) 🙂 
```suggestion
takeEven (x:_:xs) = x : takeEven xs
```

src/Chapter2.hs
```diff
@@ -727,7 +738,8 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate (x:xs) = replicate x x ++ smartReplicate xs
```
vrom911, line 742: Nice!

```haskell
[3,1,2,4,0,1]
```

src/Chapter2.hs
```diff
@@ -780,13 +792,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 803: Eta-reduction award 🏆

src/Chapter2.hs
```diff
@@ -841,7 +855,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate a l 
+  | a < 0 = []
+  | otherwise = take (length l) (drop a (cycle l))
```
vrom911, line 861: Great!
You also can implement the function more efficiently on some corner cases 🙂 For example, to check if the list is empty, and if `a` is equal to zero (by mod of the list's length) 🚀

src/Chapter2.hs
```diff
@@ -857,8 +874,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
-
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
vrom911, line 879: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Great Chapter 🥇 
I left some comments below 🙂 Let me know if you have any questions!

vrom911:

PR #257, opened 2020-10-24
Chapter 4

cherealnice:
### Solutions for Chapter 4

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -505,6 +505,10 @@ Implement the 'Applicative' instance for our 'List' type.
   type.
 -}
 
+append :: List a -> List a -> List a
+append Empty ys = ys
+append (Cons y ys) zs = Cons y (append ys zs)
+
 instance Applicative List where
   pure :: a -> List a
   pure a = Cons a Empty
@@ -512,7 +516,7 @@ instance Applicative List where
   (<*>) :: List (a -> b) -> List a -> List b
   Empty <*> _ = Empty
   _ <*> Empty = Empty
-  Cons f fs <*> Cons x xs = Cons (f x) (fs <*> xs)
+  Cons f fs <*> xs = append (fmap f xs) (fs <*> xs)
 
 {- |
 =🛡= Monad
@@ -637,10 +641,6 @@ Implement the 'Monad' instance for our lists.
 🕯 HINT: You probably will need to implement a helper function (or
   maybe a few) to flatten lists of lists to a single list.
 -}
-
-append :: List a -> List a -> List a
-append Empty ys = ys
-append (Cons y ys) zs = Cons y (append ys zs)
 
 instance Monad List where
   (>>=) :: List a -> (a -> List b) -> List b
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -114,22 +114,30 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
+Char :: *
 
 >>> :k Bool
+Bool :: *
 
 >>> :k [Int]
+[Int] :: *
 
 >>> :k []
+[] :: * -> *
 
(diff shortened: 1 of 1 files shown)


src/Chapter4.hs
```diff
@@ -629,7 +664,7 @@ Can you implement a monad version of AND, polymorphic over any monad?
 🕯 HINT: Use "(>>=)", "pure" and anonymous function
 -}
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM m1 m2 = m1 >>= (\x -> if x then m2 else pure x)
```
cherealnice, line 667: Thank you for reviewing, it's really a great project you guys have here 😄 

I originally had `andM m1 m2 = m1 >>= (\x -> m2 >>= (\y -> pure (x && y)))` but this failed the test for `Just False|Nothing`. I'm wondering if this is more accurate since `andM (Just False) Nothing` would return `Nothing` with my first implementation

src/Chapter4.hs
```diff
@@ -472,10 +487,11 @@ Implement the Applicative instance for our 'Secret' data type from before.
 -}
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    Trap x <*> _ = Trap x
+    Reward f <*> a = fmap f a
```
chshersh, line 494: Nice work on reusing `Functor` instance for `Secret`! :world_map:

src/Chapter4.hs
```diff
@@ -489,6 +505,14 @@ Implement the 'Applicative' instance for our 'List' type.
...
 -}
 
+instance Applicative List where
+  pure :: a -> List a
+  pure a = Cons a Empty
+
+  (<*>) :: List (a -> b) -> List a -> List b
+  Empty <*> _ = Empty
+  _ <*> Empty = Empty
+  Cons f fs <*> Cons x xs = Cons (f x) (fs <*> xs)
```
chshersh, line 515: Actually, you've implemented a non-standard `Applicative` instance for the list. Here you apply each function only once for the corresponding item from the second list. But in the task description we say that each function should be applied to each element from the second list.

This is still a correct `Applicative` instance. Turns out that the list has two valid `Applicative` instance. But the list type `[]` in Haskell implements a different instance, that we ask to provide in the task. With this implementation the laws of `Monad` and `Applicative` instances do not agree with each other.

Your implementation of `Applicative` for the standard list `[]` is provided via the `ZipList` newtype:

* https://hackage.haskell.org/package/base-4.14.0.0/docs/Control-Applicative.html#t:ZipList

Newtypes are the standard Haskell way to provide multiple instances for a single type. You can't have multiple instances for the same type, so you optionally choose the default (main) one, and provide others for newtypes.

src/Chapter4.hs
chshersh: Thank you for your kind words! :blush: 

The ordinary `&&` if lazy on the second argument. Meaning that if it meets `False` as the first argument, there is no need to calculate another arg. In `andM` we would like to keep the same behaviour, so the

```
Just False `andM` _
```

chshersh (approved):
:tada: Fantastic work!

src/Chapter4.hs
cherealnice: Aha, that makes sense. I misunderstood the problem. I will push an update :)

chshersh:
@cherealnice Congratulations on finishing all four Chapters! :tada: :partying_face: 
I'm going to close this PR, as always. We hope you've enjoyed learning Haskell and doing our course :hugs:

PR #260, opened 2020-10-24
Chapter3

Perceval62:
### Solutions for Chapter 3

cc @vrom911 @chshersh

Kindly asking for a review of chapter three.
I tried to do the tasks as I saw them : )

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -344,7 +344,7 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
-data Book = Book  
+data Book = Book
   { bookTitle     :: String
   , bookAuthor    :: String
   , bookISBN      :: String
@@ -384,13 +384,13 @@ after the fight. The battle has the following possible outcomes:
 
 data Knight = Knight
     { knightHealth :: Int
-    , knightAttack :: Int    
+    , knightAttack :: Int
     , knightGold   :: Int
     }
 
 data Monster = Monster
     { monsterHealth :: Int
-    , monsterAttack :: Int    
+    , monsterAttack :: Int
     , monsterGold   :: Int
     }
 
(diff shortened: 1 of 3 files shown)


src/Chapter3.hs
```diff
@@ -971,6 +1088,33 @@ implement the following functions:
...
+
+isWeekend :: DayOfWeek -> Bool
+isWeekend day = 
+  if day == Saturday || day == Sunday
+    then True
+    else False
+
+nextDay :: DayOfWeek -> DayOfWeek
+nextDay day = 
+  if day == Sunday
```
vrom911, line 1109: You could also use the instance of `Bounded` in here:

```suggestion
    if day == maxBound then minBound
```

src/Chapter3.hs
```diff
@@ -971,6 +1088,33 @@ implement the following functions:
...
+    then Monday 
+    else succ day 
+
+daysToParty :: DayOfWeek -> Int
+daysToParty day
+  | day == Friday = 0
+  | day < Friday = length (enumFromTo day Friday) - 1
+  | day > Friday = (2 - length (enumFromTo Friday day)) + 6
```
vrom911, line 1117: This could be done in a single line without all these cases with the use of `mod` function 🙂

src/Chapter3.hs
```diff
@@ -1006,6 +1150,45 @@ Implement data types and typeclasses, describing such a battle between two
...
+data Beast = Beast
+  { beastHealth :: Hlt
+  , beastAttack :: Atk
+  }
+
+flee :: Beast -> Beast
+flee beast = beast {beastHealth = 0} 
+
+class Attack target where
+  attack :: Either Hero Beast -> target -> target
```
vrom911, line 1179: Instead of `Either Hero Beast` it is better to use the polymorphic value, which should be of the class `Fighter`

vrom911 (approved):
Great Chapter! Very creative constructions you've built here 👷🏼 Very well done 👏🏼

Perceval62:
Thanks for the feedback, I took a bit more time to complete this one to really get it. :D .
I will try to learn from it as best I can for the last chapter !

PR #261, opened 2020-10-24
Attempted Chapter 1 exercises

mywyau:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,26 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: 'Char'
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
-
+42 :: Num p => p
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
-
+(True, 'x') :: (Bool, Char)
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
-
+not :: Bool -> Bool
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
-
+(&&) :: Bool -> Bool -> Bool
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
-
+(+) :: Num a => a -> a -> a
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -554,7 +546,15 @@ value after "=" where the condition is true.
...
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | x <= z && x >= y = x
+    | x <= y && x >= z = x
+    | y <= z && y >= x = y
+    | y <= x && y >= z = y
+    | z <= x && z >= y = z
+    | z <= y && z >= x = z
+    | otherwise = z
```
vrom911, line 557: It looks like you can move all these 3 cases under a single otherwise and return z in here.

As you noticed, you have covered all possible cases, so if it doesn't fit in the first 4 cases, then it will definitely satisfy one of these, but we don't care which one, we don't even need to check, as we know that the answer is going to be `z`! ⚡

src/Chapter1.hs
```diff
@@ -632,7 +640,13 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+  let
+    lastDigit2 = mod (abs n) 10
+    secondDigit = mod (div (abs n - lastDigit2) 10) 10
```
vrom911, line 647: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

Note: For `div` the last digit won't matter, so it is okay not to subtract it in here 🙂 

One hint to make your solution even shorter: you see that you use both of these:

```haskell
mod (abs n) 10
div (abs n) 10 -- div (abs n - lastDigit2) 10
```

```haskell
(a, b) = divMod (abs n) 10
```

src/Chapter1.hs
```diff
@@ -653,8 +667,15 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit n =
+  let
+    lastDigit3 x = mod (abs x) 10
+    dropLastDigit i = div (i - lastDigit3 i) 10
```
vrom911, line 674: Here you also don't need to subtract 🙂

src/Chapter1.hs
```diff
@@ -653,8 +667,15 @@ You need to use recursion in this task. Feel free to return to it later, if you
...
+    recursiveLoop j
+      | abs j <= 10 = j
+      | otherwise = recursiveLoop (dropLastDigit (abs j))
+  in recursiveLoop n
```
vrom911, line 678: You can put `abs n` in here, so you can remove doing it inside the function 🙂 
```suggestion
  in recursiveLoop (abs n)
```

vrom911 (approved):

I left some comments. Hope you find them helpful :slightly_smiling_face:

mywyau:
Hello @vrom911 thank you for the feedback 😄 really liking the short course and how the topics are explained. Thank you to everyone driving the course, appreciate the input and engagement you guys are putting in. Looking forward to finishing the other chapters.

chshersh:

PR #263, opened 2020-10-25
Chapter 2 Solutions

mywyau:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
@@ -186,31 +186,31 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
-
+[10, 2, 3, 1, 5]
 >>> [] ++ [1, 4]  -- [] is an empty list
-
+[1, 4]
 >>> 3 : [1, 2]
-
+[3, 1, 2]
 >>> 4 : 2 : [5, 10]  -- prepend multiple elements
-
+[4, 2, 5, 10]
 >>> [1 .. 10]  -- list ranges
-
+[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
 >>> [10 .. 1]
-
+[]
 >>> [10, 9 .. 1]  -- backwards list with explicit step
-
+[10,9,8,7,6,5,4,3,2,1]
 >>> length [4, 10, 5]  -- list length
-
+3
 >>> replicate 5 True
-
+[True,True,True,True,True]
 >>> take 5 "Hello, World!"
-
+"Hello"
 >>> drop 5 "Hello, World!"
-
+", World!"
 >>> zip "abc" [1, 2, 3]  -- convert two lists to a single list of pairs
-
+[('a',1),('b',2),('c',3)]
 >>> words "Hello   Haskell     World!"  -- split the string into the list of words
-
+["Hello","Haskell","World!"]
 
 
 👩‍🔬 Haskell has a lot of syntax sugar. In the case with lists, any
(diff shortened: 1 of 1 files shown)


chshersh:
Hi @mywyau! I see that you've opened a draft PR. Would you like to have feedback on already implemented solutions, or do you want to finish all tasks in Chapter Two first?

mywyau:
Hiya @chshersh I would like to finish attempting chapter 2 first.

vrom911:
Great, @mywyau 🤗 Just ping us when it is ready for review 👍🏼

mywyau:
### Solutions for Chapter 2
Updated Chapter 2 solutions thank you for the reviews
cc @vrom911 @chshersh

src/Chapter2.hs
```diff
@@ -336,8 +336,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
-
+subList i j xs
+  | i < j && i > 0 && j > 0 =  take (j - i + 1) (drop i xs)
```
vrom911, line 340: Awesome!
Though if i ==j, it could return one-element list (in case if it present):
```suggestion
  | i <= j && i > 0 && j > 0 =  take (j - i + 1) (drop i xs)
```

src/Chapter2.hs
```diff
@@ -348,9 +349,8 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
-
+firstHalf :: [a] -> [a]
+firstHalf l = take ((length l) `div` 2) l
```
vrom911, line 353: You can omit parenthesis before the operator 🙂 :
```suggestion
firstHalf l = take (length l `div` 2) l
```

src/Chapter2.hs
```diff
@@ -606,7 +607,8 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate (x:xs) = replicate 2 x ++ duplicate xs
```
vrom911, line 610: It is more efficient to construct lists using the cons operator directly without introducing intermediate values. 🚀 
```suggestion
duplicate (x:xs) = x : x : duplicate xs
```

src/Chapter2.hs
```diff
@@ -621,7 +623,9 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+takeEven (x:y:zs) = x : takeEven zs
```
vrom911, line 627: As you don't use `y` in the right part, you can use a wildcard for the pattern matching:
```suggestion
takeEven (x:_:zs) = x : takeEven zs
```

src/Chapter2.hs
```diff
@@ -621,7 +623,9 @@ Write a function that takes elements of a list only on even positions.
...
+takeEven list = if length list == 1 then list else []
```
vrom911, line 628: Actually the both cases return list itself, so no need to check 🙂 
Alternatively, you can use pattern matching to check that, it is better than using `length` generally 🚤

src/Chapter2.hs
```diff
@@ -741,8 +745,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: Int -> [[Int]] -> [[Int]]
+contains n listOfList = filter (elem n) listOfList
```
vrom911, line 749: Now, after you've mastered eta-reduction, you can apply such a technique to this function as well  🙂 

src/Chapter2.hs
```diff
@@ -781,13 +785,14 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
--- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 795: Eta-reduction award 🏆

src/Chapter2.hs
```diff
@@ -842,7 +847,8 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int] 
+rotate movementsToLeft numbers = take (length numbers) (drop movementsToLeft (cycle numbers))
```
vrom911, line 851: Unfortunately, `cycle` fails in runtime on empty lists :recycle: So you need to handle this case separately for this function to work properly.

src/Chapter2.hs
```diff
@@ -852,14 +858,13 @@ To defeat the boss, implement the reversing function that takes a list
...
-[5,4,3,2,1]
-
 ♫ NOTE: The Haskell standard library already provides the "reverse"
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
-
+rewind :: [Int] -> [Int]
+rewind (x:xs) = rewind xs ++ [x]
```
vrom911, line 866: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Wonderful work on the Chapter! 💯 Very thoughtful solutions 👏🏼

chshersh:

PR #264, opened 2020-10-25
Completed Chapter1

adrianab:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -211,7 +211,7 @@ So, the output in this example means that 'False' has type 'Bool'.
 >>> :t True
 True :: Bool
 >>> :t 'a'
-'a' :: Cha
+'a' :: Char
 >>> :t 42
 42 :: Num p => p
 
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Cha
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Cha
```
vrom911, line 214: A small typo in here, that prevents doctests from passing 🙂 

```suggestion
'a' :: Char
```

src/Chapter1.hs
```diff
@@ -631,8 +643,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = lastOne + lastTwo
+        where
+          lastOne = abs n `mod` 10
+          lastTwo = (abs n `div` 10) `mod` 10
```
vrom911, line 650: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
(a, b) = divMod (abs n) 10
```

src/Chapter1.hs
```diff
@@ -653,7 +668,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit n
+        | abs n < 10 = abs n
+        | abs n < 0 = firstDigit(abs n)
+        | otherwise = firstDigit(abs n `div` 10)
```
vrom911, line 674: That is right 💯 

Also note, that `abs n` will never be lower than 0, so this check is redundant.

And in Haskell functions and operators are always separated with spaces from the arguments, even if arguments are in parenthesis 👾 

```haskell
        | abs n < 10 = abs n
        | otherwise = firstDigit (abs n `div` 10)
```

vrom911 (approved):
Wonderful work! Congrats with finishing Chapter One 🥳 
And awesome work on the advanced task 👏🏼

vrom911:
Solutions for Chapter One look great! 🌟 I see that you opened a separate PR for Chapter Two, so I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: 

adrianab:
> Solutions for Chapter One look great! 🌟 I see that you opened a separate PR for Chapter Two, so I'm going to close this PR, because we can't merge the solutions to our repo 🙂
> But don't worry, it still counts towards Hacktoberfest contribution 👕 🌳

Sorry, still new to learning how to make pull requests, I think I did the PR for Chapter 2 correct this time, let me know if I did It right. Thank you :)

PR #265, opened 2020-10-25
Chapter 2

mridul2899:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -337,7 +337,7 @@ ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
 subList left right list
-    | left >= 0 && right >= 0 =
+    | left >= 0 && right >= left =
         let
             sub = drop left list
             final = take (right - left + 1) sub
@@ -637,7 +637,7 @@ Write a function that takes elements of a list only on even positions.
 takeEven :: [a] -> [a]
 takeEven [] = []
 takeEven (x1:_:xs) = x1 : takeEven xs
-takeEven (x1:xs) = x1 : takeEven xs
+takeEven [x] = [x]
 
 {- |
 =🛡= Higher-order functions
@@ -861,7 +861,7 @@ list.
 rotate :: Int -> [a] -> [a]
 rotate n l
     | n < 0 = []
-    | otherwise = take (length l) (drop n (cycle l))
+    | otherwise = take (length l) (drop (mod n (length l)) (cycle l))
 
 {- |
 =💣= Task 12*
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -336,8 +336,13 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
-
+subList left right list
+    | left >= 0 && right >= 0 =
```
chshersh, line 340: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding a separate case to your check :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -621,7 +634,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven (x1:_:xs) = x1 : takeEven xs
+takeEven (x1:xs) = x1 : takeEven xs
```
chshersh, line 640: Here actually only one pattern left: singleton list. So you can write it explicitly to be sure that you've covered all cases :slightly_smiling_face: 
```suggestion
takeEven [x] = [x]
```

src/Chapter2.hs
```diff
@@ -741,9 +756,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
-
+contains :: Eq a => a -> [[a]] -> [[a]]
```
chshersh, line 759: Nice work on specifying the polymorphic type signature here! :clap:

src/Chapter2.hs
```diff
@@ -842,7 +858,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate n l
+    | n < 0 = []
+    | otherwise = take (length l) (drop n (cycle l))
```
chshersh, line 864: That's a correct solution! Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -858,7 +877,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```

chshersh (approved):
And well done with the advanced task :100: 

I left some comments and suggestions regarding possible code improvements :slightly_smiling_face:

mridul2899:
Thank you very much for your suggestions. They were quite valuable. I have made the appropriate changes to ensure that implementations are now efficient.

vrom911:
The solutions look awesome 🌟 

PR #266, opened 2020-10-25
Chapter 1 done, 1 change needed in cabal and stack-yaml with my VSCode

ThreeEyedGod:
used vscode, hie, stack test

### Solutions for Chapter {1}

cc @vrom911 @chshersh


vrom911:
@ThreeEyedGod could you please add nonrelevant files into `.gitignore`? It is hard to review through 84 files 🙂

ThreeEyedGod:
So sorry; am on git learning curve as well. I think I have it fixed. Let me know if not. Appreciate you working on Sunday!

src/Chapter1.hs
```diff
@@ -554,7 +557,10 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid ::  Int -> Int -> Int -> Int
+mid x y z = 
+  (x + y + z) - (max x (max y z)) - (min x (min y z))
```
chshersh, line 562: Interesting solution! Though, not sure if it works for all numbers because of integer overflow :thinking:

src/Chapter1.hs
```diff
@@ -568,7 +574,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c =
+  elem c ['a', 'e', 'i', 'o', 'u']
```
chshersh, line 579: Absolutely correct! And you use eta-reduction 👍🏼 

src/Chapter1.hs
```diff
@@ -631,8 +640,12 @@ Implement a function that returns the sum of the last two digits of a number.
...
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = 
+  let 
+    onesplacedigit = abs n `mod` 10
+    tensplacedigit = abs n `mod` 100 `div` 10
+    in (onesplacedigit+tensplacedigit)
```
chshersh, line 648: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
```

src/Chapter1.hs
```diff
@@ -652,8 +665,16 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
+firstDigit :: Int -> Int
+firstDigit n 
+  | acc >= 10 = firstDigit acc
+  | acc == 0 = abs n
+  | otherwise = acc
+ where acc = ((abs n - lastDigit n) `div` 10)
```
chshersh, line 673: When you do `div 10`, you don't need to subtract `lastDigit`, `div` already takes care of this, and it doesn't matter for this function :slightly_smiling_face:

chshersh (approved):

We have time to review on Sunday exactly because we don't work on Sundays :laughing: 

Don't worry about git troubles, its UI/UX can be confusing sometimes.

Also, I've noticed that you've started doing Chapter Two directly in this PR. I would encourage you to open a separate PR for each own chapter. It's easier to review for us, and you also will get 4 PRs contribution towards the Hacktoberfest goal :shirt: :palm_tree:

src/Chapter2.hs
```diff
@@ -727,8 +739,11 @@ value of the element itself
 
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
+
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate xs = foldl (++) [] (map (\x -> replicate x x) xs)
```
chshersh, line 745: Nice solution!

You correctly noticed that after using `map` you will get a list of lists. To flatten such a list, you can use the `concat` function instead of `foldl`:

```haskell
[3,1,2,4,0,1]
```

src/Chapter2.hs
```diff
@@ -741,8 +756,9 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: Int -> [[Int]] -> [[Int]]
+contains _ [] = []
+contains n xs = filter (elem n) xs
```
chshersh, line 761: You don't need to handle empty lists separately, `filter` perfectly works for them :ok_hand: 

```suggestion
contains n xs = filter (elem n) xs
```

After you remove a separate case, you can eta-reduce this function even further :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -781,13 +797,19 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+--listElementsLessThan x l = filter (< x) l
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+--pairMul xs ys = zipWith (*) xs ys
+pairMul = zipWith (*)
```
chshersh, line 812: Eta-reduction award! :trophy:

chshersh (approved):
Hi @ThreeEyedGod! I've provided my review for the second chapter here is well. Looks wonderful! And great job on the Advanced Task :top: 

As I mentioned previously, I would encourage you to open separate PRs for each Chapter, so you can complete the Hacktoberfest contribution checklist. Otherwise, all changes within a single PR won't count as multiple PRs :disappointed:

vrom911:

PR #267, opened 2020-10-25
Update Chapter1.hs

callumfrance:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -554,7 +557,11 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | (y <= x && x <= z) || (z <= x && x <= y) = x
+  | (x <= y && y <= z) || (z <= y && y <= x) = y
+  | (x <= z && z <= y) || (y <= z && z <= x) = z
```

Because of that, you will need to add another guard – `| otherwise = ...`, to tell the compiler, that your pattern matching is exhaustive 🙂 In this case, as you truly covered all cases, you can skip checking the conditions on the last case, and write:

```suggestion
  | otherwise = z
```

as if other cases won't fit, then it is definitely `z` 👍🏼

src/Chapter1.hs
```diff
@@ -632,7 +646,12 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+    let n2 = abs n
+        ones = mod n2 10
+        tens = div (mod n2 100 - ones) 10
```
vrom911, line 653: Great!

vrom911 (approved):

PR #270, opened 2020-10-25
Chapter 3

will-clarke:
### Solutions for Chapter 3

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -342,6 +342,12 @@ afterwards.
 Define the Book product data type. You can take inspiration from our description
 of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
+
+data Book = MkBook {
+  pages :: Int
+  , title :: String
+} deriving (Show)
+
 -}
 
 {- |
@@ -371,8 +377,27 @@ after the fight. The battle has the following possible outcomes:
  ⊛ Monster defeats the knight. In that case return -1
  ⊛ Neither the knight nor the monster wins. On such an occasion, the knight
    doesn't earn any money and keeps what they had before.
-
 -}
+
+data Entity = MkEntity {
+  health :: Int
+  , attack :: Int
+  , gold :: Int
+}
+
+type Monster = Entity
+type Knight = Entity
+
+fight :: Monster -> Knight -> Int
+fight m k | monsterSurvivesHit && not knightSurvivesHit = -1
+          | monsterSurvivesHit && knightSurvivesHit = gold m
+          | otherwise                               = gold k
+ where
+  monsterSurvivesHit = hit k m > 0
+  knightSurvivesHit  = hit m k > 0
+
+hit :: Entity -> Entity -> Int
+hit (MkEntity _ attk _) (MkEntity hlth _ _) = hlth - attk
 
 {- |
 =🛡= Sum types
@@ -460,6 +485,8 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data Food = Breakfast | Lunch | Dinner | ForbiddenSnacks
+
 {- |
 =⚔️= Task 4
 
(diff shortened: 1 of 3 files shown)


src/Chapter3.hs
```diff
@@ -342,6 +342,12 @@ afterwards.
 Define the Book product data type. You can take inspiration from our description
 of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
+
+data Book = MkBook {
+  pages :: Int
+  , title :: String
+} deriving (Show)
```
chshersh, line 349: You can write this data type outside comments to make sure that it actually compiles :heavy_check_mark:

src/Chapter3.hs
```diff
@@ -371,9 +377,28 @@ after the fight. The battle has the following possible outcomes:
...
+data Entity = MkEntity {
+  health :: Int
+  , attack :: Int
+  , gold :: Int
+}
+
+type Monster = Entity
+type Knight = Entity
+
+fight :: Monster -> Knight -> Int
```
chshersh, line 391: Nice usage of type aliases!

```haskell
> fight knight monster
> fight monster knight
```

```haskell
```

src/Chapter3.hs
```diff
@@ -478,8 +505,45 @@ After defining the city, implement the following functions:
...
+
+peopleInHouse :: House -> Int
+peopleInHouse house = case house of
+  One   -> 1
+  Two   -> 2
+  Three -> 3
+  Four  -> 4
+
+peopleCount :: City -> Int
+peopleCount city = sum $ map peopleInHouse $ houses city
```
chshersh, line 545: Great solution for this task! Nice types and functions :gem:

src/Chapter3.hs
```diff
@@ -560,37 +624,41 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor a) (Dexterity d) = Defense (a * d)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Int
+calculatePlayerHit (Damage d) (Defense dfns) (Health h) = h + dfns - d
```
chshersh, line 651: Nice work on newtypes! :medal_sports: This function can also return new `Health` as the result :heavy_plus_sign: 


Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -907,9 +980,21 @@ Implement instances of "Append" for the following types:
...
 
+instance Append Gold where
+  append (Gold a) (Gold b) = Gold (a + b)
+
+instance Append [a] where
+  append = (++)
+
+instance Append a => Append (Maybe a) where
+  append (Just x) (Just y) = Just (append x y)
+  append _        _        = Nothing
```
chshersh, line 996: While this is valid implementation and nothing in the task description says differently, in reality it would be more beneficial is this function returned `Just` when at least one argument is `Just`.

src/Chapter3.hs
```diff
@@ -971,6 +1056,25 @@ implement the following functions:
...
+data Day = Monday | Tuesday | Wednesday | Thursday | Friday | Saturday | Sunday deriving (Enum)
+
+isWeekend :: Day -> Bool
+isWeekend Sunday   = True
+isWeekend Saturday = True
+isWeekend _        = False
+
+-- maybe there's a better way to make `succ` wrap here??
+nextDay :: Day -> Day
+nextDay Sunday = Monday
```
chshersh, line 1068: That is right!

```suggestion
    | day == maxBound = minBound
```

src/Chapter3.hs
```diff
@@ -971,6 +1056,25 @@ implement the following functions:
...
+nextDay day    = succ day
+
+daysToParty :: Day -> Int
+daysToParty day = go day 0
+ where
+  go :: Day -> Int -> Int
+  go Friday n = n
+  go day'   n = go (nextDay day') (n + 1)
```
chshersh, line 1076: Nice recursive implementation! But you can implement it more efficiently by using the `fromEnum` method of the `Enum` typeclass

chshersh (approved):
Nice job on Chapter Three :top: 
Keep up the good work :muscle:

src/Chapter3.hs
will-clarke: Thanks so much for the comment - I din't know about `Bounded` instances! 🎉

src/Chapter3.hs
will-clarke: Thanks so much for the explanation - having the compiler type-check the types, rather than just aliasing, sounds like a sensible idea :+1:

PR #271, opened 2020-10-25
Chapter 1

Gregdams:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-True :: Bool
+<INSERT THE RESULT INSTEAD OF THE TEXT>
 >>> :t 'a'
-'a' :: Char
+<INSERT THE RESULT INSTEAD OF THE TEXT>
 >>> :t 42
-42 :: Num p => p
+<INSERT THE RESULT INSTEAD OF THE TEXT>
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-(True, 'x') :: (Bool, Char)
+<INSERT THE RESULT INSTEAD OF THE TEXT>
 
 Boolean negation:
 >>> :t not
-not :: Bool -> Bool
+<INSERT THE RESULT INSTEAD OF THE TEXT>
 
 Boolean 'and' operator:
 >>> :t (&&)
-(&&) :: Bool -> Bool -> Bool
+<INSERT THE RESULT INSTEAD OF THE TEXT>
 
 Addition of two numbers:
 >>> :t (+)
-(+) :: Num a => a -> a -> a
+<INSERT THE RESULT INSTEAD OF THE TEXT>
 
 Maximum of two values:
 >>> :t max
-max :: Ord a => a -> a -> a
+<INSERT THE RESULT INSTEAD OF THE TEXT>
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +490,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = abs n `mod` 10
```
chshersh, line 494: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1:

src/Chapter1.hs
```diff
@@ -553,8 +554,12 @@ value after "=" where the condition is true.
...
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | (x > y && x < z) || (x < y && x > z) = x
+  | (y > x && y < z) || (y < x && y > z) = y
+  | (z > x && z < y) || (z < x && z > y) = z
+  | otherwise = x
```
chshersh, line 562: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

Because of that, you will need to add another guard – `| otherwise = x`, to tell the compiler, that your pattern matching is exhaustive 🙂


It looks like you can move the last two cases under a single otherwise and return `z` there, if you also use `>=` and `<=` instead of `>` and `<` in other guards :slightly_smiling_face: 

src/Chapter1.hs
```diff
@@ -568,9 +573,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
-
-
+isVowel :: Char -> Bool
+isVowel c
+  | c `elem` ['a', 'e', 'i', 'o', 'u', 'y'] = True
```
chshersh, line 578: Nice one! 👍🏼 

```haskell
```

src/Chapter1.hs
```diff
@@ -632,8 +638,10 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n = last2 `div` 10 + last2 `mod` 10
+    where
+      last2 = abs n `mod` 100
```
chshersh, line 644: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
```

src/Chapter1.hs
```diff
@@ -653,7 +661,12 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+  | x >= 10 = firstDigit (x `div` 10)
+  | otherwise = x
+  where
+    x = abs n
```
chshersh, line 669: Nice usage of `where` to reduce some code duplication :clap:

chshersh (approved):

Gregdams:
Thank you for your feedback @chshersh  :D

chshersh:
@Gregdams The solutions look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to open PRs with solutions to other chapters!

PR #272, opened 2020-10-25
Solutions for Chapter {2}

pislas:
This chapter was difficult for me, but I enjoyed solving the tasks with haskell. Thanks!!
cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -755,10 +755,10 @@ the list with only those lists that contain a passed element.
 -}
 contains :: Int -> [[Int]] -> [[Int]]
 contains n l = go [] l
-  where 
+  where
     go :: [[Int]] -> [[Int]] -> [[Int]]
     go acc [] = reverse acc
-    go acc (x:xs) = if (elem n x) then go (x:acc) xs else go acc xs
+    go acc (x:xs) = if n `elem` x then go (x:acc) xs else go acc xs
 
 {- |
 =🛡= Eta-reduction
@@ -880,7 +880,7 @@ and reverses it.
   cheating!
 -}
 rewind :: [a] -> [a]
-rewind l = go [] l
+rewind = go []
   where
     go :: [a] -> [a] -> [a]
     go acc [] = acc
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True,False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +336,10 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList x y l
+  | x < 0 || y <0 = []
+  | x > y = []
+  | otherwise = drop x (take (y+1) l)
```
chshersh, line 342: It is recommended to add spaces on both sides of binary operators in Haskell :arrow_left: :space_invader: :arrow_right: 
```suggestion
  | otherwise = drop x (take (y + 1) l)
```

src/Chapter2.hs
```diff
@@ -620,8 +627,14 @@ Write a function that takes elements of a list only on even positions.
...
-takeEven = error "takeEven: Not implemented!"
-
+takeEven :: [a] -> [a]
+takeEven n = go 0 [] n
+  where
+    go :: Int -> [a] -> [a] -> [a]
+    go _ acc [] = reverse acc
+    go i acc (x:xs)
+      |mod i 2 == 0 = go (i +1) (x:acc) xs
+      |otherwise = go (i+1) acc xs
```
chshersh, line 637: Nice idea on the helper functions :+1:

Also, calling `reverse` in the end is a bit slow, as `reverse` traverses the whole list. You can avoid calling `reverse` by adding new elements directly to the result:

```haskell
| mod i 2 == 0 = x : go (i +1) acc xs
```

After that refactoring you can notice that the `acc` variable becomes redundant :recycle:

src/Chapter2.hs
```diff
@@ -727,7 +740,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concat (map (\x -> replicate x x) l)
```
chshersh, line 743: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -740,8 +753,12 @@ the list with only those lists that contain a passed element.
...
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: Int -> [[Int]] -> [[Int]]
+contains n l = go [] l
+  where
+    go :: [[Int]] -> [[Int]] -> [[Int]]
+    go acc [] = reverse acc
+    go acc (x:xs) = if n `elem` x then go (x:acc) xs else go acc xs
```
chshersh, line 761: Nice implementation with the Recursive go pattern!

In this particular function, you can use standard function `filter` to keep only those elements, satisfying the given check

src/Chapter2.hs
```diff
@@ -780,13 +797,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
chshersh, line 808: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -841,7 +860,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate _ [] = []
+rotate n l = if n>=0 then subList n (length l + (n -1)) (cycle l) else []
```
chshersh, line 865: That's a correct solution! Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -857,7 +879,12 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind = go []
+  where
+    go :: [a] -> [a] -> [a]
+    go acc [] = acc
+    go acc (x:xs) = go (x:acc) xs
```
chshersh, line 887: Nice and idiomatic solution! :gem:

chshersh (approved):
Nice work on Chapter Two!
And great job on the Advanced Task :medal_sports: 

I understand that new concepts can be difficult and challenging indeed. But don't worry, you're doing great :muscle: I left some comments with suggestions and possible improvements, but your solutions are already nice :slightly_smiling_face:

PR #274, opened 2020-10-26
Chapter4

Perceval62:
### Solutions for Chapter 4

cc @vrom911 @chshersh

I'll definitely be coming back to this chapter in the future for help and reference !

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -114,22 +114,24 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
-
+Char :: *
 >>> :k Bool
-
+Bool :: *
 >>> :k [Int]
-
+[Int] :: *
 >>> :k []
-
+[] :: * -> *
 >>> :k (->)
-
+(->) :: * -> * -> *
 >>> :k Either
-
+Either :: * -> * -> *
 >>> data Trinity a b c = MkTrinity a b c
 >>> :k Trinity
+Trinity :: * -> * -> * -> *
 
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
+IntBox :: (* -> *) -> *
 
 -}
 
@@ -291,9 +293,11 @@ method. Yes, similar to how we can partially apply functions. See, how
 we can reuse already known concepts (e.g. partial application) from
 values and apply them to the type level?
 -}
+
 instance Functor (Secret e) where
     fmap :: (a -> b) -> Secret e a -> Secret e b
-    fmap = error "fmap for Box: not implemented!"
+    fmap _ (Trap e) = Trap e
+    fmap f (Reward a) = Reward (f a) 
 
 {- |
 =⚔️= Task 3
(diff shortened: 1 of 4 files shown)


src/Chapter4.hs
```diff
@@ -472,10 +481,11 @@ Implement the Applicative instance for our 'Secret' data type from before.
 -}
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    Trap e <*> _ = Trap e
+    Reward f <*> x = fmap f x
```
vrom911, line 488: Nice work on reusing `Functor` instance for `Secret`! :world_map:

src/Chapter4.hs
```diff
@@ -611,6 +622,18 @@ Implement the 'Monad' instance for our lists.
   maybe a few) to flatten lists of lists to a single list.
 -}
 
+-- I couldn't complete this.
+-- I looked at the solutions and it wouldn't compile either.
+
+{-
+instance Monad List where
+  (>>=) :: List a -> (a -> List b) -> List b
+  list >>= f = flatHelper (fmap f l)
```
vrom911, line 631: Here you named the list as `list` on the left, but use `l` on the right 🤔

src/Chapter4.hs
```diff
@@ -611,6 +622,18 @@ Implement the 'Monad' instance for our lists.
...
+
+flatHelper :: List (List a) -> List a
+flatHelper Empty = Empty
+flatHelper (Cons x xs) =  x : (flatHelper xs)
```
vrom911, line 635: Here `x` is list itself as this is the list of lists inside 🙂 
So `:` won't work, you need to use some `append` function to get 2 lists together 🙂

src/Chapter4.hs
```diff
@@ -629,7 +652,7 @@ Can you implement a monad version of AND, polymorphic over any monad?
 🕯 HINT: Use "(>>=)", "pure" and anonymous function
 -}
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM m1 m2 = m1 >>= \x -> if x then m2 else pure False
```
vrom911, line 655: Yes, this is an absolutely correct lazy implementation 👍🏼

src/Chapter4.hs
```diff
@@ -678,3 +701,23 @@ Specifically,
...
+
+reverseBinaryTree :: BinaryTree a -> BinaryTree a
+reverseBinaryTree Val = Val
+reverseBinaryTree (Node x left right) = 
+  Node x (reverseBinaryTree right) (reverseBinaryTree left)
+
+treeToList :: BinaryTree a -> [a]
+treeToList Val = []
+treeToList (Node x left right) = 
+  x : treeToList left ++ treeToList right
```
vrom911, line 723: Great work with the trees 🎄 🌴 🔝

vrom911 (approved):
:tada: Fantastic work! If you have any questions at the moment, don't hesitate to ask! Even about the task that you implemented, but feel unsure about 👍🏼 

Thank you so much for taking this course! I hope you had fun :man_dancing: 💃🏼 

PR #276, opened 2020-10-26
Chapter 2

Gregdams:
### Solutions for Chapter {2}

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -337,7 +337,7 @@ ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
 subList x y l
-  | x > y || x > length l || x < 0 || y < 0 = []
+  | x > y || x < 0 || y < 0 = []
   | otherwise = drop x (take (y + 1) l)
 
 {- |
@@ -352,7 +352,6 @@ Implement a function that returns only the first half of a given list.
 -}
 -- PUT THE FUNCTION TYPE IN HERE
 firstHalf :: [a] -> [a]
-firstHalf [] = []
 firstHalf l = take (length l `div` 2) l
 
 {- |
@@ -505,7 +504,7 @@ True
 False
 -}
 isThird42 :: [Int] -> Bool
-isThird42 (_:_:z:_) = z == 42
+isThird42 (_:_:42:_) = True
 isThird42 _ = False
 
 {- |
@@ -628,12 +627,9 @@ Write a function that takes elements of a list only on even positions.
 -}
 takeEven :: [Int] -> [Int]
 takeEven [] = []
-takeEven l = go [] l
-  where
-    go :: [Int] -> [Int] -> [Int]
-    go acc [] = acc
-    go acc (x:_:xs) = go (acc ++ [x]) xs
-    go acc (x:_) = go (acc ++ [x]) []
+takeEven (x:_:xs) = x : takeEven xs
+takeEven (x:_) = x : takeEven []
+
 {- |
 =🛡= Higher-order functions
 
@@ -739,7 +735,6 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate [] = []
 smartReplicate l = concatMap (\x -> replicate x x) l
 
 {- |
@@ -754,7 +749,6 @@ the list with only those lists that contain a passed element.
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
 contains :: Int -> [[Int]] -> [[Int]]
-contains _ [] = []
 contains nb l = filter ((nb `elem`)) l
 
 {- |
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +336,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList x y l
+  | x > y || x > length l || x < 0 || y < 0 = []
```
vrom911, line 340: In lists, to get `length` is very expensive process. You need to walk each element to count that, so I would not advice to have this check in here 🙂 You actually don't need to check `x > length l` at all, as `take` and `drop` handle these cases okay 👍🏼

src/Chapter2.hs
```diff
@@ -349,8 +351,9 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 -- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
-
+firstHalf :: [a] -> [a]
+firstHalf [] = []
+firstHalf l = take (length l `div` 2) l
```
vrom911, line 356: Great! Note that `take` covers case with the empty list, so no need to have this separately 🤠

src/Chapter2.hs
```diff
@@ -501,8 +504,9 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
-
+isThird42 :: [Int] -> Bool
+isThird42 (_:_:z:_) = z == 42
```
vrom911, line 508: You don't need to compare `z` with 42 separately, you can pattern match on `42` directly in the pattern 🎨

src/Chapter2.hs
```diff
@@ -621,8 +626,14 @@ Write a function that takes elements of a list only on even positions.
...
-takeEven = error "takeEven: Not implemented!"
-
+takeEven :: [Int] -> [Int]
+takeEven [] = []
+takeEven l = go [] l
+  where
+    go :: [Int] -> [Int] -> [Int]
+    go acc [] = acc
+    go acc (x:_:xs) = go (acc ++ [x]) xs
+    go acc (x:_) = go (acc ++ [x]) []
```
vrom911, line 636: Usually, adding elements to the end of the list is costly. Your implementation is correct :+1: But Alternatively, you could use pattern matching in `takeEven` instead. As you need to take every second element, it is possible to represent with a few pattern matching clauses 🙂

src/Chapter2.hs
```diff
@@ -728,7 +739,8 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
```
vrom911, line 742: You don't need this check separately 🙂

src/Chapter2.hs
```diff
@@ -741,8 +753,9 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: Int -> [[Int]] -> [[Int]]
+contains _ [] = []
```
vrom911, line 757: And here as well. All of the standard list functions usually cover cases with the empty list, so you don't need to worry about it 😉

src/Chapter2.hs
```diff
@@ -781,13 +794,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 805: Eta-reduction award 🏆

src/Chapter2.hs
```diff
@@ -842,7 +857,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate _ [] = []
+rotate 0 l = l
+rotate nb l = drop nb (take (length l + nb) (cycle l))
```
vrom911, line 863: Nice catch on the case when `nb` is zero 🔥 
There is a way to optimise it even more! The current solution does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. you can check this case and reduce `nb` somehow 👍🏼

src/Chapter2.hs
```diff
@@ -858,9 +876,13 @@ and reverses it.
...
-rewind = error "rewind: Not Implemented!"
-
-
+rewind :: [a] -> [a]
+rewind [] = []
+rewind l = go [] l
+  where
+    go :: [a] -> [a] -> [a]
+    go acc [] = acc
+    go acc xs = foldl (flip (:)) acc xs
```
vrom911, line 885: Wow, that's a clever solution using standard functions! 🤯

Btw, you don't need to handle the case with `[]` in the `go` function, since `foldl` already takes care of this. And, after you do this, you can notice, that there is no need in the `go` function with such implementation at all, and you can implement `rewind` in a single line 🙂

vrom911 (approved):
Great Chapter 👏🏼 Very interesting and creative solutions, bravo 🎉

Gregdams:
Thank you for your feedback @vrom911 :D

vrom911:
No worries, @Gregdams 😊  

PR #282, opened 2020-10-26
Chapter 2 done

karlcoelho:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -1,21 +1,14 @@
 {- 👋 Welcome to Chapter Two, Traveller!
-
 If you haven't finished Chapter One yet, we encourage you to check it
 out, as the knowledge from the previous section is required in order
 to pass this training.
-
 If you already finished Chapter One — congratulations and huge
 respect! We are happy to see you with us at the next step.
-
 A casual reminder of how it works.
-
 == You are on __Chapter Two__.
-
 Here we will explore more interesting functional concepts. We are
 going to practice our skills on the list type in Haskell.
-
 In this chapter, you are going to master:
-
  ✧ The list data type
  ✧ Immutability
  ✧ Pattern-matching
@@ -25,16 +18,13 @@ In this chapter, you are going to master:
  ✧ Higher-order functions
  ✧ Partial function application
  ✧ Eta-reduction
-
 As usual, the explanations are in the Haskell comments of this
 module. We are leaving a number of tasks on our path. Your goal is to
 solve them all and make the tests for Chapter Two green.
-
 After finishing the PR, you can choose to summon us, @vrom911 and @chshersh,
 to look at your solution in order to give some advice on your
 code. This is optional; however, you can ask us for review only if you
 want some feedback on your solutions.
-
 Now, if you are ready, bring it on!
 -}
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -25,203 +18,164 @@ In this chapter, you are going to master:
...
+(:) :: a -> [a] -> [a]
 Reverse a list:
 >>> :t reverse
-
-
+reverse :: [a] -> [a]
 Take first N elements of a list:
 >>> :t take
-
-
```
chshersh, line 164: I see that empty lines were removed here. Unfortunately, because of this, Doctests on CI will fail :disappointed: They need an empty line after the output and before the next textual block

src/Chapter2.hs
```diff
@@ -492,50 +395,42 @@ the same as writing "(x:(y:xs))".
...
 is the number 42.
-
 >>> isThird42 [1, 2, 42, 10]
 True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird42 :: [a] -> Bool
+isThird42 (_:_:42:_) = True
```
chshersh, line 406: I see that this code doesn't compile. In the type signature, you write that the function `isThird42` works with any type `a`, but in the implementation, in pattern, you match on number 42. Not every type is a number. To fix this problem, specify the type of a function to accept a list of integers:

```haskell
isThird42 :: [Int] -> Bool
```

src/Chapter2.hs
```diff
@@ -683,139 +551,110 @@ partially, we will get a new function that takes only one number and
...
 value of the element itself
-
 >>> smartReplicate [3, 1, 2]
 [3,3,3,1,2,2]
-
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concat(map(\n -> replicate n n) l)
```
chshersh, line 587: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -683,139 +551,110 @@ partially, we will get a new function that takes only one number and
...
 
--- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: a -> [a] -> [a]
+listElementsLessThan x = filter (< x)
 
--- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [a] -> [a] -> [a]
+pairMul = zipWith (*)
```
chshersh, line 636: Eta-reduction award! :trophy:

src/Chapter2.hs
```diff
@@ -825,42 +664,41 @@ Infinity Stone!
...
 >>> rotate 3 [1,2,3,4]
 [4,1,2,3]
-
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: a -> [a] -> [a]
+rotate n l
+  | n < 0 || null l = []
+  | otherwise = drop n (take (length 1+n) (cycle l))
```
chshersh, line 682: That's a correct solution! Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

chshersh (approved):
Neat work on Chapter Two tasks!

PR #283, opened 2020-10-26
Chapter 4

will-clarke:
### Solutions for Chapter 4

cc @vrom911 @chshersh

:D

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -114,22 +114,30 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
+Char :: *
 
 >>> :k Bool
+Bool :: *
 
 >>> :k [Int]
+[Int] :: *
 
 >>> :k []
+[] :: * -> *
 
 >>> :k (->)
+(->) :: * -> * -> *
 
 >>> :k Either
+Either :: * -> * -> *
 
 >>> data Trinity a b c = MkTrinity a b c
 >>> :k Trinity
+Trinity :: * -> * -> * -> *
 
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
+IntBox :: (* -> *) -> *
 
 -}
 
@@ -292,8 +300,9 @@ we can reuse already known concepts (e.g. partial application) from
 values and apply them to the type level?
 -}
 instance Functor (Secret e) where
-    fmap :: (a -> b) -> Secret e a -> Secret e b
-    fmap = error "fmap for Box: not implemented!"
+  fmap :: (a -> b) -> Secret e a -> Secret e b
+  fmap f (Reward a) = Reward (f a)
+  fmap _ (Trap   e) = Trap e
 
 {- |
 =⚔️= Task 3
@@ -305,7 +314,12 @@ typeclasses for standard data types.
 -}
 data List a
     = Empty
-    | Cons a (List a)
+    | Cons a (List a) deriving (Show)
+
+instance Functor List where
+  fmap :: (a -> b) -> List a -> List b
+  fmap _ Empty         = Empty
+  fmap f (Cons a rest) = Cons (f a) $ fmap f rest
 
 {- |
 =🛡= Applicative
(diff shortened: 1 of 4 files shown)


src/Chapter4.hs
```diff
@@ -673,6 +722,26 @@ Specifically,
...
+  fmap f (Node x l r) = Node (f x) (fmap f l) (fmap f r)
+
+reverseTree :: BinaryTree a -> BinaryTree a
+reverseTree EmptyNode    = EmptyNode
+reverseTree (Node x l r) = Node x (reverseTree r) (reverseTree l)
+
+treeToList :: BinaryTree a -> List a
+treeToList EmptyNode = Empty
+treeToList (Node x l r) =
+  append (append (treeToList l) (Cons x Empty)) (treeToList r)
```
will-clarke, line 744: Is it possible to use fmap in any of these tree functions? I was trying to use it... but couldn't find a decent way to. Perhaps with a decent monad implementation we'd be able to use >>= to convert Tree to List? 🤔

src/Chapter4.hs
```diff
@@ -470,12 +484,20 @@ Applicatives can be found in many applications:
...
-    pure = error "pure Secret: Not implemented!"
+  pure :: a -> Secret e a
+  pure = Reward
 
-    (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+  (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
+  (Reward f) <*> a        = fmap f a
+  (Trap   e) <*> _        = Trap e
+  _          <*> (Trap e) = Trap e
```
chshersh, line 500: This case is actually redundant! See, if the left argument is `Trap`, this is covered by the previous line. And if it's `Reward`, the `Functor` instance for `Secret` already handles it, and it will return `Trap e` :slightly_smiling_face:

src/Chapter4.hs
```diff
@@ -611,6 +651,15 @@ Implement the 'Monad' instance for our lists.
...
 
+listConcat :: List (List a) -> List a
+listConcat Empty         = Empty
+listConcat (Cons xs xss) = append xs (listConcat xss)
+
+
+instance Monad List where
+  (>>=) :: List a -> (a -> List b) -> List b
+  Empty >>= _ = Empty
+  l     >>= f = listConcat $ fmap f l
```
chshersh, line 662: The first pattern with `Empty` is redundant. `fmap` for `Empty` returns `Empty`, and `listConcat` of `Empty` is also `Empty`, so no need to handle separate cases :slightly_smiling_face: Notice, how it's enough to cover such bases cases only in basic fundamental functions, so later they compose nicely and you don't need to worry about missing patterns in such scenarios :earth_africa:

src/Chapter4.hs
```diff
@@ -629,7 +678,7 @@ Can you implement a monad version of AND, polymorphic over any monad?
 🕯 HINT: Use "(>>=)", "pure" and anonymous function
 -}
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM a b = a >>= (\a' -> b >>= (\b' -> pure (a' && b')))
```
chshersh, line 681: The ordinary `&&` if lazy on the second argument. Meaning that if it meets `False` as the first argument, there is no need to calculate another arg. In `andM` we would like to keep the same behaviour, so the

```
Just False `andM` _
```

And your current implementation returns `Nothing` in this case.

src/Chapter4.hs
```diff
@@ -673,6 +722,26 @@ Specifically,
  ❃ Implement the function to convert Tree to list
 -}
 
+data BinaryTree a = EmptyNode | Node a (BinaryTree a) (BinaryTree a)
+  deriving (Show)
+
+insert :: Ord a => a -> BinaryTree a -> BinaryTree a
+insert x EmptyNode = Node x EmptyNode EmptyNode
+insert x (Node val l r) | x > val   = Node val l (insert x r)
+                        | otherwise = Node val (insert x l) r
```
chshersh, line 731: Nice implementation of `insert` to Binary Search Tree! :top: :100:

src/Chapter4.hs
chshersh: No, it's not possible to use `fmap` :disappointed: And the `Tree` data type is also not a `Monad`. This final boss combines tasks from different chapters: pattern-matching, custom data types, Functor from this chapter. If you want to play with functors, you can implement a function that increments each value in each tree node, but that's not so interesting :slightly_smiling_face: 

Here you can simplify the implementation a bit by using fewer appends and writing `append` in infix form:

```suggestion
  treeToList l `append` Cons x (treeToList r)
```

chshersh (approved):
:tada: Fantastic work!

src/Chapter4.hs
will-clarke: 💡 That's really nice! 😃

will-clarke:
Thanks so much for all your help! I've really enjoyed the course & have given some feedback at that URL.
😃

PR #284, opened 2020-10-27
Chapter 2 Solutions

leyohla:
### Solutions for Chapter 2

cc @vrom911 @chshersh

Opened this new PR rather than my old PR just to fix up some things

Thanks!

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -39,6 +39,8 @@ Now, if you are ready, bring it on!
 -}
 
 module Chapter2 where
+import Data.List
+import Data.Map (Map)
 
 {-
 =🛡= Imports
@@ -136,43 +138,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+[Char]
 
 Empty list:
 >>> :t []
-
+[a]
 
 Append two lists:
 >>> :t (++)
-
+[a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+[a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+[String] -> String
 
 -}
 
@@ -186,30 +188,43 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
+[10,2,3,1,5]
 
 >>> [] ++ [1, 4]  -- [] is an empty list
+[1, 4]
 
 >>> 3 : [1, 2]
+[3, 1, 2]
 
 >>> 4 : 2 : [5, 10]  -- prepend multiple elements
+[4,2,5,10]
 
 >>> [1 .. 10]  -- list ranges
+[1,2,3,4,5,6,7,8,9,10]
 
 >>> [10 .. 1]
+[]
 
 >>> [10, 9 .. 1]  -- backwards list with explicit step
+[10,9,8,7,6,5,4,3,2,1]
 
 >>> length [4, 10, 5]  -- list length
+3
 
 >>> replicate 5 True
+[True,True,True,True,True]
 
 >>> take 5 "Hello, World!"
+"Hello"
 
 >>> drop 5 "Hello, World!"
+", World!"
 
 >>> zip "abc" [1, 2, 3]  -- convert two lists to a single list of pairs
+[('a',1),('b',2),('c',3)]
 
 >>> words "Hello   Haskell     World!"  -- split the string into the list of words
+["Hello","Haskell","World!"]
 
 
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -39,6 +39,8 @@ Now, if you are ready, bring it on!
 -}
 
 module Chapter2 where
+import Data.List
+import Data.Map (Map)
```
chshersh, line 43: I wonder if this import is needed :thinking: 
Also, I'm not sure that it compiles... The `Data.Map` module is in the `containers` library, but we depend only on `base` (the standard library). Unfortunately, container types are not in the standard Haskell library :disappointed:

src/Chapter2.hs
```diff
@@ -336,7 +351,11 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList c d ef
+    | c < 0 = []
+    | d < 0 = []
+    | otherwise = (take ((d - c) + 1) (drop c ef))
```
chshersh, line 357: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding a separate case to your check :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -500,7 +519,11 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+
+isThird42 :: [Int] -> Bool
+isThird42 (x:(y:42:xs)) = True
+isThird42 (_:xs) = False
```
chshersh, line 525: That's the correct implementation! :gem: 

Usually, when variable names don't matter, it's recommended to use underscores `_` instead of them to indicate in the code, that those variables are not used. And, you also can remove a few extra parenthesis. So, something like this should work and look a bit more cleaner :slightly_smiling_face: 

```suggestion
isThird42 (_ : _ : 42 : _) = True
isThird42 (_ : _) = False
```

src/Chapter2.hs
```diff
@@ -620,7 +644,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+takeEven [] = []
+takeEven (x:_:xs) = x : takeEven xs 
+takeEven [x] = [x]
```
chshersh, line 650: Great pattern-matching skills! :partying_face:

src/Chapter2.hs
```diff
@@ -727,7 +754,9 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate (x:xs) = replicate x x ++ smartReplicate xs
```
chshersh, line 758: This is right! 💯 

```haskell
[3,1,2,4,0,1]
```

src/Chapter2.hs
```diff
@@ -740,7 +769,12 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains p [] = []
+contains p (x:xs)
+  | p `elem` x = x : contains p xs
+  | otherwise = contains p xs
```
chshersh, line 776: Absolutely correct implementation! :+1: 

Additionally, you can use the `filter` function instead of manual pattern matching and recursive. `filter` takes a function `(a -> Bool)`, a list of a's (`[a]`) and returns the list where all elements satisfy the given boolean function :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -780,13 +814,13 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
--- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
--- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
chshersh, line 823: Eta-reduction award! :trophy:

src/Chapter2.hs
```diff
@@ -841,7 +875,11 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate p [] = []
+rotate p xs
+  | p <= 0 = []
+  | otherwise = zipWith const (drop (abs (p)) (cycle xs)) xs
```
chshersh, line 882: That's a very clever implementation :brain:

src/Chapter2.hs
```diff
@@ -857,7 +895,11 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [Int] -> [Int]
+rewind [] = []
+rewind (x:xs)
+  | x <= 0 = []
+  | otherwise = rewind xs ++ [x]
```
chshersh, line 902: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

src/Chapter2.hs
```diff
@@ -857,7 +895,11 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [Int] -> [Int]
+rewind [] = []
+rewind (x:xs)
+  | x <= 0 = []
```
chshersh, line 901: I see, that your implementation works with lists of integer, and you check that elements of the list shouldn't be negative. I guess, this is not necessary, and you make this function work with any list :slightly_smiling_face: 

```haskell
rewind :: [a] -> [a]
```

chshersh (approved):
Wonderful solutions for Chapter Two! :clap: 
And well done with the Advanced Task :+1: 

I left some comments. The second chapter indeed contains a lot more new concepts, and can be challenging :jigsaw: But your solutions are great! :medal_sports:

vrom911:

PR #285, opened 2020-10-27
chap 1 done

sxxrfx:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -575,7 +575,7 @@ isVowel c
     | ((c == 'e') || c == 'E')  = True
     | ((c == 'i') || c == 'I')  = True
     | (c == 'o') || c == 'O'  = True
-    | ((c == 'u') || c == 'U')  = True
+    | (c == 'u') || c == 'U'  = True
     | otherwise = False
 
 {- |
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -518,7 +518,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = if (abs x) > abs y then y else x
+closestToZero x y = if abs x > abs y then y else x
 
 
 {- |
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -574,7 +574,7 @@ isVowel c
     | ((c == 'a') || c == 'A')  = True
     | ((c == 'e') || c == 'E')  = True
     | ((c == 'i') || c == 'I')  = True
-    | ((c == 'o') || c == 'O')  = True
+    | (c == 'o') || c == 'O'  = True
     | ((c == 'u') || c == 'U')  = True
     | otherwise = False
 
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -488,7 +488,7 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 lastDigit :: Int -> Int
-lastDigit n = (n) `mod` 10
+lastDigit n = n `mod` 10
 
 
 {- |
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,29 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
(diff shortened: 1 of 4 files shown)


src/Chapter1.hs
```diff
@@ -209,31 +209,29 @@ So, the output in this example means that 'False' has type 'Bool'.
...
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
-
```
vrom911, line 221: This line is intentional because it is needed for the doctests to pass 🙂

src/Chapter1.hs
```diff
@@ -489,8 +487,8 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = n `mod` 10
```
vrom911, line 491: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1: 

Unfortunately, this implementation returns negative numbers on negative inputs because of how `mod` works. Sometimes corner cases can be tricky to spot and fix...

src/Chapter1.hs
```diff
@@ -520,7 +518,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if (abs x) > abs y then y else x
```
vrom911, line 521: Parenthesis are redundant here

```suggestion
closestToZero x y = if abs x > abs y then y else x
```

src/Chapter1.hs
```diff
@@ -553,8 +551,11 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | ((x > y) && (y > z)) || ((x < y) && (y < z)) = y
+    | ((y > x) && (x > z)) || ((y < x) && (x < z)) = x
+    | ((y > x) && (z > x)) || ((y < x) && (z < x)) = z
```
vrom911, line 558: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

But as you noticed, these cases cover all possible cases, so if you didn't fit the first two patterns, you know for sure that the answer is `z`, you can even skip checking it:

```suggestion
    | otherwise = z
```

src/Chapter1.hs
```diff
@@ -568,8 +569,14 @@ True
...
 -}
-isVowel c = error "isVowel: not implemented!"
-
+isVowel :: Char -> Bool
+isVowel c
+    | ((c == 'a') || c == 'A')  = True
+    | ((c == 'e') || c == 'E')  = True
+    | ((c == 'i') || c == 'I')  = True
+    | (c == 'o') || c == 'O'  = True
+    | (c == 'u') || c == 'U'  = True
```
vrom911, line 578: You can skip all the parenthesises in this function! :)

src/Chapter1.hs
```diff
@@ -631,8 +638,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = div (last2 n) 10 + mod (last2 n) 10
```
vrom911, line 642: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (last2 n) 10
div (last n) 10
```
The standard library has the [`divMod`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Prelude.html#v:divMod) function, that actually combines inside both `div` and `mod`. And this is exactly what you use!

Also, you need to consider negative numbers in this task as well 🙂

src/Chapter1.hs
```diff
@@ -652,8 +662,8 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n = if (div n 10) > 9 then (firstDigit (div n 10)) else div n 10
```
vrom911, line 666: I see, that for a negative number like -19, this implementation will not work correctly in here 🤔 


Also, You can see that `div n 10` is reused several times in this function. This usually means that you can move it into the `where` case and reuse everywhere 👍🏼

vrom911 (approved):
Great stuff! Congrats with finishing Chapter One 🎉

chshersh:

PR #286, opened 2020-10-27
Solutions to Chapter 1.

howey-aus:
### Solutions for Chapter 1

cc @vrom911  @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -646,7 +646,7 @@ sumLast2 :: Int -> Int
 sumLast2 n = (last2 n - last1 n) `div` 10 + last1 n
     where
         last2 :: Int -> Int
-        last2 n = (abs n) `mod` 100
+        last2 n = abs n `mod` 100
         last1 :: Int -> Int
         last1 n = (abs n) `mod` 10
 
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -631,8 +642,13 @@ Implement a function that returns the sum of the last two digits of a number.
...
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = (last2 n - last1 n) `div` 10 + last1 n
+    where
+        last2 :: Int -> Int
+        last2 n = abs n `mod` 100
+        last1 :: Int -> Int
+        last1 n = abs n `mod` 10
```
chshersh, line 651: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
```

Also, you don't need to subtract the last digit when doing `div 10`. The `div` function already handles this :slightly_smiling_face:

src/Chapter1.hs
```diff
@@ -631,8 +642,13 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = (last2 n - last1 n) `div` 10 + last1 n
+    where
+        last2 :: Int -> Int
+        last2 n = abs n `mod` 100
```
chshersh, line 649: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1:

chshersh (approved):
Perfect solutions for Chapter One! :gem: 

I left a few general notes. But the solutions are already great :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face: But don't worry, it still counts towards Hacktoberfest contribution :shirt: :deciduous_tree: Feel free to open PRs with solutions to other chapters!

PR #287, opened 2020-10-27
 Chapter 3 done

karlcoelho:
### Solutions for Chapter 3

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -343,7 +343,13 @@ Define the Book product data type. You can take inspiration from our description
 of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
-
+data Book = Book
+  {
+    bookAuthor :: String
+  , bookTitle  :: String
+  , bookYear   :: Int
+  , bookPages  :: Int
+  }
 {- |
 =⚔️= Task 2
 
@@ -373,6 +379,26 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+
+data Knight = Knight
+  {
+    kHealth :: Int
+  , kAttack :: Int
+  , kGold   :: Int
+  } deriving(Show)
+
+data Monster = Monster
+  {
+    mHealth :: Int
+  , mAttack :: Int
+  , mGold   :: Int
+  } deriving(Show)
+
+fight :: Knight ->  Monster -> Int
+fight k m
+  | mHealth m <= kAttack k = kGold k + mGold magic
+  | kHealth k <= mAttack m = (-1)
+  | otherwise = kGold k
 
 {- |
 =🛡= Sum types
@@ -460,6 +486,8 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data Meal = Breakfast | Lunch | Teatime | Dinner | Supper
+
 {- |
 =⚔️= Task 4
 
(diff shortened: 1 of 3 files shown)


src/Chapter3.hs
```diff
@@ -479,6 +507,41 @@ After defining the city, implement the following functions:
    complicated task, walls can be built only if the city has a castle
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
+data peopleInHouse = One | Two | Three | Four
+countPeople :: peopleInHouse -> Int
```
chshersh, line 511: I see that this code doesn't compile :thinking: Data type names should start with an uppercase letter according to Haskell naming rules
```suggestion
data PeopleInHouse = One | Two | Three | Four
countPeople :: PeopleInHouse -> Int
```

src/Chapter3.hs
```diff
@@ -560,22 +623,31 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defense (armor * dexterity)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage damage) (Defense defense) (Health health) = Health (health + defense - damage)
```
chshersh, line 650: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -971,6 +1069,24 @@ implement the following functions:
...
+
+nextDay :: day -> day
+nextDay d
+  | d == maxBound = minBound
+  | otherwise = succ d
+
+daysToParty :: day -> Int
+daysToParty day
+  | day > Friday = (fromEnum day - fromEnum Friday) + fromEnum Friday
+  | otherwise = fromEnum Friday - fromEnum day
```
chshersh, line 1088: Nice solution! You can implement it even shorter, if you remember from Chapter One, how `mod` works for negative numbers :slightly_smiling_face:

src/Chapter3.hs
```diff
@@ -1006,7 +1122,78 @@ Implement data types and typeclasses, describing such a battle between two
...
 -}
 
-
+--Assuming data for each player (monster/knight)
+newtype PlayerHealth = PlayerHealth Int
+newtype PlayerAttack = PlayerAttack Int
+newtype PlayerDefense = PlayerDefense Int
+
+--All possible actions
+data BattleAction = Attack | DrinkPotion Int | CastSpell Int | MonsterAttack | MonsterRun
```
chshersh, line 1131: It would be better to create two separate data types for `MonsterAction` and `FighterAction`, so it won't be possible to mix actions of different types :slightly_smiling_face:

src/Chapter3.hs
```diff
@@ -1006,7 +1122,78 @@ Implement data types and typeclasses, describing such a battle between two
...
+  getAttack = monsterAttack
+
+--Defensing
+class PlayerDefensing a where
+  getDefense :: a -> PlayerDefense
+
+instance PlayerDefensing FighterDefense where
+  getDefense = fighterDefense
+
+roundFighting :: BattleAction -> FighterAttack -> MonsterAttack -> BattleResult
```
chshersh, line 1169: It looks like data types `FitherData` and `MonsterData` are not used in this function :thinking: 
Also, this function works with specific types, not polymorphic. Probably you need to use typeclasses defined before

chshersh (approved):
Nice work on Chapter Three!
And great start on the Andvanced Task :+1: 

I see that this module doesn't compile. I would recommend utilise the Haskell compiler to make sure that implementations are maximally corrent :slightly_smiling_face:

PR #288, opened 2020-10-27
resolve Chapter 3

krivtsov:
### Solutions for Chapter 3
cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -344,6 +344,12 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+data MyBook = MyBook
+  { title   :: String
+  , author  :: String
+  , year    :: Int
+  }
+
 {- |
 =⚔️= Task 2
 
@@ -373,6 +379,19 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+
+
+data Properties = P Int Int Int
+
+type Knight = Properties
+type Monster = Properties
+
+fight :: Knight -> Monster -> Int
+fight (P kh ka kg) (P mh ma mg)
+  | mh < 0 = kg + mg
+  | kh < 0 = -1
+  | kh == mh = kg
+  | otherwise = fight (P (kh - ma) ka kg) (P (mh - ka) ma mg)
 
 {- |
 =🛡= Sum types
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -479,6 +479,13 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data MealTypes
+    = Breakfast
+    | TeaAfterMeeting
+    | Tea5Clock
+    | Dinner
+    | NightSnack
+
 {- |
 =⚔️= Task 4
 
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -374,6 +380,19 @@ after the fight. The battle has the following possible outcomes:
 
 -}
 
+
+data Properties = P Int Int Int
+
+type Knight = Properties
+type Monster = Properties
+
+fight :: Knight -> Monster -> Int
```
chshersh, line 389: Nice usage of type aliases!

```haskell
> fight knight monster
> fight monster knight
```

```haskell
```

src/Chapter3.hs
```diff
@@ -374,6 +380,19 @@ after the fight. The battle has the following possible outcomes:
 
 -}
 
+
+data Properties = P Int Int Int
```
chshersh, line 384: It's generally recommended to use records instead of unnamed fields. When I see three `Int`s, it's not clear from the data type definition what they mean. But if you implement this data type as a record, its definition already becomes clearer, even if you don't use record fields as getters or setters :slightly_smiling_face: 

Alternatively, you can write documentation to each field, but in such cases it almost always makes sense to implement data types with a single constructor as records.

src/Chapter3.hs
```diff
@@ -480,6 +506,42 @@ After defining the city, implement the following functions:
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
 
+data City = City
+  { castle :: Castle
+  , wall :: Wall
```
chshersh, line 511: This is the correct data type, that can represent what we have in requirements 👍🏼 

src/Chapter3.hs
```diff
@@ -561,21 +623,31 @@ introducing extra newtypes.
...
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerDamage :: Attack -> Strength -> Damage
+calculatePlayerDamage (Attack attack) (Strength strength) = Damage (attack + strength)
+
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defense (armor * dexterity)
+
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage damage) (Defense defense) (Health health) = Health (health + defense - damage)
```
chshersh, line 650: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -910,6 +989,18 @@ Implement instances of "Append" for the following types:
...
+
+data List = List [String]
+
+instance Append Gold where
+  append :: Gold -> Gold -> Gold
+  append (Gold x) (Gold y) = Gold (x + y)
+
+instance Append List where
+  append :: List -> List -> List
+  append (List x) (List y) = List (x ++ y)
```
chshersh, line 1002: In this task it wasn't required to implement a separate `List` data type as a wrapper around single list. So you could implement the `Append` instance directly for `[a]`. But if you want to have a wrapper around any list, you can create a polymorphic data type:
```haskell
newtype List a = List [a]
```

src/Chapter3.hs
```diff
@@ -971,6 +1062,23 @@ implement the following functions:
...
 
+data WeekDay = Mon | Tue | Wed | Thu | Fri | Sat | Sun deriving (Enum)
+
+isWeekend :: WeekDay -> Bool
+isWeekend Sat = True
+isWeekend Sun = True
+isWeekend _ = False
+
+nextDay :: WeekDay -> WeekDay
+nextDay Sun = Mon
```
chshersh, line 1073: That is right!

```suggestion
    | day == maxBound = minBound
```

src/Chapter3.hs
```diff
@@ -971,6 +1062,23 @@ implement the following functions:
...
+nextDay d = succ d
+
+daysToParty :: WeekDay -> Int
+daysToParty day =  iter day 0 
+  where
+    iter Fri x = x
+    iter d x = iter (nextDay d) (x + 1)
```
chshersh, line 1080: Great implementation! Alternatively, you can use the `fromEnum` method of the `Enum` typeclass to implement this function more efficiently :slightly_smiling_face:

chshersh (approved):
Wonderful work on Chapter Three! :clap: 
Let us know if you have any questions regarding the Advanced Task :slightly_smiling_face: 

I left some comments with suggestions and general feedback :+1:

chshersh:

PR #290, opened 2020-10-28
Done chapter 1

poyea:
### Solutions for Chapter {1}

Thanks for this amazing project!

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -553,8 +542,11 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | x > y = if y > z then y else if x > z then z else x
+    | otherwise = if x > z then x else if y > z then z else y
```
chshersh, line 548: That's a nice solution!
I guess, you can use only guards here instead of mixing `guards` and `if-then-else` expressions, but that's just a stylistic preference :slightly_smiling_face:

src/Chapter1.hs
```diff
@@ -631,8 +626,14 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
+sumLast2 :: Int -> Int
+sumLast2 n = lastOne n + secToLast n
+  where
+    lastOne :: Int -> Int
+    lastOne n = mod (abs n) 10
+    secToLast :: Int -> Int
+    secToLast n = let nd = mod (div (abs n) 10) 10 in nd
```
chshersh, line 635: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
```

src/Chapter1.hs
```diff
@@ -652,8 +653,10 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+    | div n 10 == 0 = abs n
+    | otherwise = firstDigit (abs (div n 10))
```
chshersh, line 659: That is right 💯 

chshersh (approved):
Neat work on solving our challenges!

vrom911:

PR #294, opened 2020-10-28
Chapter 3

mridul2899:
### Solutions for Chapter 3

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -343,6 +343,15 @@ Define the Book product data type. You can take inspiration from our description
 of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
+data Book = Book
+    { bookName :: String
+    , author :: String
+    , publisher :: String
+    , edition :: Int
+    , publishingYear :: Int
+    , pages :: Int
+    , rating :: Double
+    } deriving (Show)
 
 {- |
 =⚔️= Task 2
@@ -373,7 +382,29 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+data Knight = Knight
+    { kHealth :: Int
+    , kAttack :: Int
+    , kGold :: Int
+    } deriving (Show)
 
+data Monster = Monster
+    { mHealth :: Int
+    , mAttack :: Int
+    , mGold :: Int
+    } deriving (Show)
+
+fight :: Knight -> Monster -> Int
+fight k m
+    | kAttack k <= 0 && mAttack m <= 0 = kGold k
+    | kAttack k >= mHealth m = kGold k + mGold m
+    | mAttack m >= kHealth k = -1
+    | otherwise = fight kNew mNew
+      where
+        kNew :: Knight
+        kNew = k {kHealth = kHealth k - mAttack m}
+        mNew :: Monster
+        mNew = m {mHealth = mHealth m - kAttack k}
 {- |
 =🛡= Sum types
 
@@ -459,7 +490,7 @@ and provide more flexibility when working with data types.
 Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
-
+data Meals = Breakfast | Brunch | Lunch | HighTea | Snacks | Dinner
 {- |
 =⚔️= Task 4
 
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -479,7 +510,54 @@ After defining the city, implement the following functions:
...
+
+buildHouse :: City -> House -> City
+buildHouse city house =
+    city {cityHouses = cityHouses city ++ [house]}
+
+convert :: House -> Int
+convert One = 1
+convert Two = 2
+convert Three = 3
+convert _ = 4
```
chshersh, line 551: For the sum types it's usually recommended to handle all constructors explicitly instead of using wildcard. Let's say that tomorrow you need to add the constructor `Five` to the `House` data type. If you write all constructors explicitly, the compiler warns you when you forget to pattern match, so it becomes much easier to refactor and add new features :slightly_smiling_face: 

Also, here, to avoid some code duplication, you can use the `case-of` expression:

```haskell
convert house = case house of
    One -> 1
    ... -> ...
```

src/Chapter3.hs
```diff
@@ -479,7 +510,54 @@ After defining the city, implement the following functions:
...
+
+countPeople :: [House] -> Int
+countPeople = foldr ((+) . convert) 0
```
chshersh, line 554: There's a standard function `sum` for this. If you're already familiar with the composition operator `.` then you can write this in a slightly cleaner way :slightly_smiling_face: 
```suggestion
countPeople = sum . map convert
```

src/Chapter3.hs
```diff
@@ -560,22 +638,46 @@ introducing extra newtypes.
...
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense armor dexterity =
+    Defense (getArmor armor * getDexterity dexterity)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit damage defense health =
+    Health (getHealth health + getDefense defense - getDamage damage)
```
chshersh, line 680: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -907,9 +1015,18 @@ Implement instances of "Append" for the following types:
...
 class Append a where
     append :: a -> a -> a
 
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append (Gold a) (Gold b) = Gold (a + b)
+
+instance Append [x] where
+    append :: [x] -> [x] -> [x]
+    append a b = a ++ b
```
chshersh, line 1029: You can eta-reduce the implementation of `append` for brevity. So you can see how easy it is to create instances :slightly_smiling_face: 
```suggestion
    append = (++)
```

src/Chapter3.hs
```diff
@@ -970,6 +1087,28 @@ implement the following functions:
...
+    | Sunday
+    deriving (Enum, Show)
+
+isWeekend :: WeekDay -> Bool
+isWeekend Sunday = True
+isWeekend Saturday = True
+isWeekend _ = False
+
+nextDay :: WeekDay -> WeekDay
+nextDay Sunday = Monday
```
chshersh, line 1106: That is right!

```suggestion
    | day == maxBound = minBound
```

src/Chapter3.hs
```diff
@@ -970,6 +1087,28 @@ implement the following functions:
...
+nextDay day = succ day
+
+daysToParty :: WeekDay -> Int
+daysToParty Friday = 0
+daysToParty x = 1 + daysToParty (nextDay x)
```
chshersh, line 1111: That's the correct solution :+1: But you make it slightly more efficient by using the `fromEnum` method of the `Enum` typeclass :slightly_smiling_face:

chshersh (approved):
Wonderful solutions to Chapter Three tasks!
I left some feedback with general notes and suggestions :slightly_smiling_face:

mridul2899:
Thank you for the review. The feedback was constructive and helped me to have more clarity on some of these topics. I have made the suggested changes.

chshersh:
@mridul2899 The solutions look good :+1: I'm going to close this PR, because we can't merge the solutions to our repo :slightly_smiling_face:

PR #296, opened 2020-10-28
Chapter 4

mridul2899:
### Solutions for Chapter 4

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -633,7 +633,6 @@ flatten (Cons x list) = concatenate x (flatten list)
 
 instance Monad List where
     (>>=) :: List a -> (a -> List b) -> List b
-    Empty >>= _ = Empty
     list >>= f = flatten (fmap f list)
 
 
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -114,23 +114,23 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
-
+Char :: *
 >>> :k Bool
-
+Bool :: *
 >>> :k [Int]
-
+[Int] :: *
 >>> :k []
-
+[] :: * -> *
 >>> :k (->)
-
+(->) :: * -> * -> *
 >>> :k Either
-
+Either :: * -> * -> *
 >>> data Trinity a b c = MkTrinity a b c
 >>> :k Trinity
-
+Trinity :: * -> * -> * -> *
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
-
+IntBox :: (* -> *) -> *
 -}
 
 {- |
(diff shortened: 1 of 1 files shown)


src/Chapter4.hs
```diff
@@ -472,10 +476,11 @@ Implement the Applicative instance for our 'Secret' data type from before.
 -}
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    Trap e <*> _ = Trap e
+    (Reward f) <*> a = fmap f a
```
vrom911, line 483: Nice work on reusing `Functor` instance for `Secret`! :world_map:

src/Chapter4.hs
```diff
@@ -488,7 +493,18 @@ Implement the 'Applicative' instance for our 'List' type.
   may also need to implement a few useful helper functions for our List
   type.
 -}
+concatenate :: List a -> List a -> List a
+concatenate x Empty = x
+concatenate Empty x = x
```
vrom911, line 498: Nice catch!

You won't need to walk through each element of the first list when the second one is empty and return the list instantly 🚤

src/Chapter4.hs
```diff
@@ -610,6 +627,14 @@ Implement the 'Monad' instance for our lists.
 🕯 HINT: You probably will need to implement a helper function (or
   maybe a few) to flatten lists of lists to a single list.
 -}
+flatten :: List (List a) -> List a
+flatten Empty = Empty
+flatten (Cons x list) = concatenate x (flatten list)
+
+instance Monad List where
+    (>>=) :: List a -> (a -> List b) -> List b
+    Empty >>= _ = Empty
```
vrom911, line 636: This case is redundant as both `flatten` and `fmap` covers the empty case 🙂

src/Chapter4.hs
```diff
@@ -629,7 +654,7 @@ Can you implement a monad version of AND, polymorphic over any monad?
 🕯 HINT: Use "(>>=)", "pure" and anonymous function
 -}
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM m1 m2 = m1 >>= (\x -> if x then m2 else pure False)
```
vrom911, line 657: Great lazy implementation 👍🏼

vrom911 (approved):
:tada: Fantastic work! Let me know if you have any questions about Tree task 🎄 🌴 
Thank you so much for taking this course! I hope you had fun :man_dancing: 🦖  

vrom911:
Brilliant! Thanks again for participating and congrats with finishing Hacktoberfest2020 🎉

PR #299, opened 2020-10-28
Chapter 4

ufocoder:
### Solutions for Chapter 4

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -114,23 +114,23 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
-
+Char :: *
 >>> :k Bool
-
+Bool :: *
 >>> :k [Int]
-
+[Int] :: *
 >>> :k []
-
+[] :: * -> *
 >>> :k (->)
-
+(->) :: * -> * -> *
 >>> :k Either
-
+Either :: * -> * -> *
 >>> data Trinity a b c = MkTrinity a b c
 >>> :k Trinity
-
+Trinity :: * -> * -> * -> *
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
-
+IntBox :: (* -> *) -> *
 -}
 
 {- |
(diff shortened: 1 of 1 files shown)


src/Chapter4.hs
```diff
@@ -472,10 +478,11 @@ Implement the Applicative instance for our 'Secret' data type from before.
 -}
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    (Trap e) <*> _ = Trap e
+    (Reward f) <*> x = fmap f x
```
vrom911, line 485: Nice work on reusing `Functor` instance for `Secret`! :world_map:

src/Chapter4.hs
```diff
@@ -489,6 +496,18 @@ Implement the 'Applicative' instance for our 'List' type.
...
 -}
 
+instance Applicative List where
+    pure :: a -> List a
+    pure a = Cons a Empty
+
+    (<*>) :: List (a -> b) -> List a -> List b
+    Empty <*> _ = Empty
+    _ <*> Empty = Empty
+    Cons f fs <*> zs = append (fmap f zs) (fs <*> zs)
```
vrom911, line 506: This is correct! Also, you could reuse `fmap` in these two cases as well 🙂

src/Chapter4.hs
```diff
@@ -489,6 +496,18 @@ Implement the 'Applicative' instance for our 'List' type.
...
+
+append :: List a -> List a -> List a
+append Empty ys = ys
```
vrom911, line 509: Perfect!

```haskell
append xs Empty = xs
```

src/Chapter4.hs
```diff
@@ -629,7 +652,7 @@ Can you implement a monad version of AND, polymorphic over any monad?
 🕯 HINT: Use "(>>=)", "pure" and anonymous function
 -}
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM ma mb = ma >>= \a -> if a then  mb else pure a
```
vrom911, line 655: Neat Lazy implementation 🎉

vrom911 (approved):
:tada: Brilliant work!
Thank you so much for taking this course! I hope you had fun 🎈 🌮  

Also, let me know if you need any consultation on the final task 👍🏼

PR #300, opened 2020-10-28
Chap3

Gregdams:
### Solutions for Chapter 3

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -545,7 +545,8 @@ data MagicalCity = MagicalCity
     }
 
 buildCastle :: String -> Maybe MagicalCity -> MagicalCity
-buildCastle newname _ = createCity
+buildCastle newname (Just city) = city { castle = Just Castle { __name__ = newname, wall = False } }
+buildCastle newname Nothing = createCity
   where
     newCastle = Castle { __name__ = newname, wall = False }
     newHouses = [Houses { __people__ = 1 }] 
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -534,7 +534,7 @@ data Castle = Castle
       wall :: Wall
     }
 
-data Houses = Houses
+newtype Houses = Houses
     {
       __people__ :: Int
     }
@@ -563,12 +563,12 @@ buildHouse np city = city { houses = newHouse:hcity }
 
 buildWalls :: MagicalCity -> MagicalCity
 buildWalls city
-  | length castleName > 0 && totalPeoples >= 10 = city { castle = ccity { wall = True } }
+  | not (null castleName) && totalPeoples >= 10 = city { castle = ccity { wall = True } }
   | otherwise = city
   where
     ccity = castle city
     castleName = __name__ ccity
-    totalPeoples = sum (map ( __people__ ) (houses city))
+    totalPeoples = sum (map __people__ (houses city))
 
 {-
 =🛡= Newtypes
@@ -681,7 +681,7 @@ calculatePlayerDefense :: Int -> Int -> Defense
 calculatePlayerDefense armor dexterity = Defense (armor * dexterity)
 
 calculatePlayerHit :: Damage -> Defense -> HitHealth -> HitHealth
-calculatePlayerHit damage defense hithealth = HitHealth ((hh hithealth) + (def defense) - (dmg damage))
+calculatePlayerHit damage defense hithealth = HitHealth (hh hithealth + def defense - dmg damage)
 
 -- The second player hits first player and the new first player is returned
 hitPlayer :: Player -> Player -> Player
@@ -859,7 +859,7 @@ parametrise data types in places where values can be of any general type.
   maybe-treasure ;)
 -}
 
-data Dragon a = Dragon
+newtype Dragon a = Dragon
     {
       magic_power :: a
     }
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -494,13 +494,13 @@ comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
 data Meal =
-  Breakfast Bool |
-  Brunch Bool |
-  Elevenses Bool |
-  Lunch Bool |
-  Tea Bool |
-  Supper Bool |
-  Dinner Bool
+  Breakfast |
+  Brunch    |
+  Elevenses |
+  Lunch     |
+  Tea       |
+  Supper    |
+  Dinner
 
 {- |
 =⚔️= Task 4
@@ -524,8 +524,6 @@ After defining the city, implement the following functions:
 
 type Wall = Bool
 
-type Church = Bool
-type Library = Bool
 data Building = Library | Church
 
 data Castle = Castle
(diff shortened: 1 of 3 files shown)


src/Chapter3.hs
```diff
@@ -460,6 +493,15 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
...
 -}
 
+data Meal =
+  Breakfast Bool |
+  Brunch Bool |
+  Elevenses Bool |
+  Lunch Bool |
+  Tea Bool |
+  Supper Bool |
+  Dinner Bool
```
vrom911, line 503: Great! 😋 
What `Bool` represents in the constructors?

src/Chapter3.hs
```diff
@@ -480,6 +522,54 @@ After defining the city, implement the following functions:
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
 
+type Wall = Bool
+
+type Church = Bool
+type Library = Bool
+data Building = Library | Church
```
vrom911, line 529: You would have name clashes. It is better not to name the constructors as already defined types 🙂 

Btw, I don't see where you use those type aliases?

src/Chapter3.hs
```diff
@@ -480,6 +522,54 @@ After defining the city, implement the following functions:
...
+  where
+    ccity = castle city
+    newCastle = ccity { __name__ = newname }
+
+
+buildHouse :: Int -> MagicalCity -> MagicalCity
+buildHouse np city = city { houses = newHouse:hcity }
+  where
+    hcity = houses city
+    newHouse = Houses { __people__ = np }
```
vrom911, line 561: Interesting approach. It is usually called "smart constructor". However, to avoid building houses with more than 4 people you need to check it here, and also restrict usage of the Constructor and allow to create `House`s only with your `buildHouse` function.

Generally, in here, it is possible and easy restrict that on type level, by using something else than `Int` 🙂

src/Chapter3.hs
```diff
@@ -480,6 +522,54 @@ After defining the city, implement the following functions:
...
+    }
+
+newtype Houses = Houses
+    {
+      __people__ :: Int
+    }
+
+data MagicalCity = MagicalCity
+    {
+      castle :: Castle,
```
vrom911, line 544: Actually, this is optional, so the type could be:
```suggestion
      castle :: Maybe Castle,
```

vrom911: Oh, I see that you indecate castle availability by the empty/non-empty string. 
In this case,this also allows having some states that should not ideally exist: like having walls without any castle at all. With this representation, this is possible. However, we actually can restrict this on type-level by providing more representative types 🙂 This is actually the greatest power of Haskell and type system 😍 

src/Chapter3.hs
```diff
@@ -560,22 +650,38 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Int -> Int -> Defense
+calculatePlayerDefense armor dexterity = Defense (armor * dexterity)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> HitHealth -> HitHealth
+calculatePlayerHit damage defense hithealth = HitHealth (hh hithealth + def defense - dmg damage)
```
vrom911, line 684: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -753,6 +859,17 @@ parametrise data types in places where values can be of any general type.
...
 
+newtype Dragon a = Dragon
+    {
+      magic_power :: a
+    }
+
+data Lair a = Lair
+    {
+      dragon :: Dragon a,
+      treasure_chest :: Maybe Bool
```
vrom911, line 870: Interesting. What states did you want to represent with `Maybe Bool`.

* `Just True` - this is when the treasure is present?
* `Just False` - 🤔 
* `Nothing` - no treasures?

I think you could use a simpler type here 🙂

src/Chapter3.hs
```diff
@@ -907,9 +1024,25 @@ Implement instances of "Append" for the following types:
...
 class Append a where
     append :: a -> a -> a
 
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append (Gold a) (Gold b) = Gold(a + b)
+
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append x y = x ++ y
```
vrom911, line 1038: You can eta-reduce the implementation of `append` for brevity. So you can see how easy it is to create instances :slightly_smiling_face: 
```suggestion
    append = (++)
```

src/Chapter3.hs
```diff
@@ -907,9 +1024,25 @@ Implement instances of "Append" for the following types:
...
 
 -}
+newtype Gold = Gold Int
+
 class Append a where
     append :: a -> a -> a
 
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append (Gold a) (Gold b) = Gold(a + b)
```
vrom911, line 1034: In Haskell arguments are always separated from functions with spaces 👾 
```suggestion
    append (Gold a) (Gold b) = Gold (a + b)
```

src/Chapter3.hs
```diff
@@ -907,9 +1024,25 @@ Implement instances of "Append" for the following types:
...
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append x y = x ++ y
+
+instance (Append a) => Append (Maybe a) where
+    append :: Maybe a -> Maybe a -> Maybe a
+    append (Just x) (Just y) = Just (append x y)
+    append (Just x) Nothing = Just x
+    append Nothing (Just y) = Just y
+    append Nothing Nothing = Nothing
```
vrom911, line 1045: This is an absolutely correct solution!

src/Chapter3.hs
```diff
@@ -971,6 +1104,28 @@ implement the following functions:
...
+  deriving (Eq,Ord,Enum,Show)
+
+isWeekend :: Week -> Bool
+isWeekend Saturday = True
+isWeekend Sunday = True
+isWeekend _ = False
+
+nextDay :: Week -> Week
+nextDay Sunday = Monday
+nextDay w = toEnum (fromEnum w + 1)
```
vrom911, line 1124: Unfortunately, `toEnum` will fail in the runtime when you go out the number of constructors. You would need to use `mod` here as well 🙂 
Alternatively, you could use the `Bounded` instance.

src/Chapter3.hs
```diff
@@ -1006,6 +1161,43 @@ Implement data types and typeclasses, describing such a battle between two
...
+class Fighter a where
+  attacked :: Int -> a -> a
+
+instance Fighter FKnight where
+  attacked :: Int -> FKnight -> FKnight
+  attacked mDmg k = k { fkHealth = fkHealth k + fkDefense k - mDmg }
+
+instance Fighter FMonster where
+  attacked :: Int -> FMonster -> FMonster
+  attacked kDmg m = m { fmHealth = fmHealth m - kDmg }
```
vrom911, line 1200: Nice!
You can also write `fight` function to imitate the real battle ⚔️ 
For example:
```haskell
fight :: [FKActions] -> FKnight -> [FMActions] -> FMonster -> Winner
```

vrom911 (approved):
Great Chapter! You constructed nice types 👷🏼

Gregdams:
Thank you for your feedback :D I hope I solve all cases correctly ^^

src/Chapter3.hs
```diff
@@ -541,34 +539,53 @@ newtype Houses = Houses
...
   where
-    ccity = castle city
-    newCastle = ccity { __name__ = newname }
-
+    newCastle = Castle { __name__ = newname, wall = False }
+    newHouses = [Houses { __people__ = 1 }] 
+    createCity = MagicalCity { 
+      castle = Just newCastle, 
+      building = Library, 
+      houses = newHouses
```
vrom911, line 555: This is strange that you erase all previous city data and write new default 🤔

Gregdams: buildCastle — build a castle in the city. If the city `already has a castle, the old castle is destroyed`, and the new castle with the __new name__ is built

I misunderstanding the exercise ?

Gregdams: Ohh I got it ! Sorry I destroy everything x) 
Ok I change that

vrom911:

PR #301, opened 2020-10-29
Chapter 2 Solutions

siv12345:
### Solutions for Chapter {2}

Thanks a lot 

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -628,7 +628,7 @@ Write a function that takes elements of a list only on even positions.
 takeEven :: [a] -> [a]
 takeEven [] = []
 takeEven (x:_:xs) = x : takeEven xs
-takeEven (x:_) = x : takeEven []
+takeEven [x] = [x]
 
 {- |
 =🛡= Higher-order functions
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ Then, do [the workaround](https://www.stackage.org/blog/2020/08/ghc-8-10-2-windo
 If you're on Linux or macOS, then the process is easy:
 
 1. Install [ghcup](https://www.haskell.org/ghcup/) and follow `ghcup`
-   instructions for successful installation.
+   instructions for successful installation (remember to restart your terminal afterwards to avoid an `unknown ghcup command` error on the next step).
 2. Install the latest version of the Haskell compiler — GHC — and the
    [Cabal](https://www.haskell.org/cabal/) build tool. After you install
    `ghcup`, it is easy to install the rest with a few commands from your
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -621,9 +621,11 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 firstDigit :: Int -> Int
-firstDigit n
-  | div n 10 == 0 = n
-  | otherwise = firstDigit (div n 10)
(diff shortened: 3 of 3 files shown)


siv12345:
Not sure why but the CI doesn't seem to be running, but on my local machine all the test cases passed

src/Chapter2.hs
```diff
@@ -336,7 +336,8 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList x y a = if x < 0 || y < 0 || x > y then []
+else take (y - x + 1) (drop x a)
```
chshersh, line 340: Does it compile with `else` starting from the beginning of a line? :thinking:

src/Chapter2.hs
```diff
@@ -621,7 +625,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven (x:_:xs) = x : takeEven xs
+takeEven (x:_) = x : takeEven []
```
chshersh, line 631: I would write this case more explicitly to be sure that the compiler can check all patterns :slightly_smiling_face: Also, since `takeEven []` returns `[]`, you can simplify this as well
```suggestion
takeEven [x] = [x]
```

src/Chapter2.hs
```diff
@@ -728,7 +735,8 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate l = concatMap (\x -> replicate x x) l
```
chshersh, line 739: Great implementation! You don't need to handle an empty list as a separate pattern. Also, you can eta-reduce the implementation a bit :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -858,7 +874,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [Int] -> [Int]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
chshersh, line 879: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
And well done on the Advanced Task :+1: 

I haven't noticed this PR initially, so I've commented in a PR in your fork. Too many PRs :slightly_smiling_face: But I'm providing my review here, since this PR is open, and some reviewing features don't work on closed/merged PRs.

src/Chapter2.hs
siv12345: Yes the code seems to be running fine

src/Chapter2.hs
siv12345: Yeah, I saw this seems to be a less than idea implementation, Will try to implement it later

siv12345:
Thanks again @chshersh I am closing this PR

PR #303, opened 2020-10-29
Done Chapter2

poyea:
### Solutions for Chapter {2}

cc @vrom911 @chshersh

```
divToZero :: Int -> Int
divToZero n = go 0 n
  where
    go :: Int -> Int -> Int
    go acc 0 = acc
    go acc n = go (acc + 1) (div n 2)
```
can probably be represented by `floor(log2(n)) + 1`. Correct me if I'm wrong.

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -861,7 +861,7 @@ list.
 -}
 rotate :: Int -> [Int] -> [Int]
 rotate a x
-    | length x == 0 = []
+    | null x = []
     | otherwise = drop a (take (a + length x) (cycle x))
 
 {- |
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -880,7 +880,7 @@ and reverses it.
 -}
 rewind :: [Int] -> [Int]
 rewind [] = []
-rewind (x:xs) = (rewind xs) ++ [x]
+rewind (x:xs) = rewind xs ++ [x]
 
 {-
 You did it! Now it is time to open pull request with your changes
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -621,7 +627,12 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven [y] = [y]
+takeEven (y : _ : xs) = y : takeEven xs
```
chshersh, line 633: Great pattern-matching skills! :tada:

src/Chapter2.hs
```diff
@@ -741,7 +752,11 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains a x
+    | null x = []
+    | a `elem` ((take 1 x) !! 0) = ((take 1 x) !! 0) : contains a (drop 1 x)
+    | otherwise = contains a (drop 1 x)
```
chshersh, line 759: That's the correct implementation :+1: But it can be simplified by using the `filter` function. You're correctly using `elem` to check if the element is in the list. Now, you can combine it with the `filter` that already keeps only those elements satisfying the given boolean function :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -842,7 +859,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate a x
+    | null x = []
+    | otherwise = drop a (take (a + length x) (cycle x))
```
chshersh, line 865: That's a correct solution! Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -858,8 +878,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
-
+rewind :: [Int] -> [Int]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
chshersh, line 883: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
Great work on Chapter Two!
And well done on the Advanced task :+1: 

I left some comments and suggestions.

And yes, you answer is one possible and correct way to interpret the meaning of this recursive function :slightly_smiling_face:

PR #306, opened 2020-10-29
Chapter 3 solutions

callumfrance:
### Solutions for Chapter 3

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -344,6 +344,13 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+data Book = Book
+    { name      :: String
+    , author    :: String
+    , cover     :: String
+    , pages     :: Int
+    }
+
 {- |
 =⚔️= Task 2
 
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -752,6 +841,19 @@ parametrise data types in places where values can be of any general type.
...
+    }
+
+data TreasureChest c = TreasureChest
+    { treasureGold :: Int
+    , treasureLoot :: c
+    }
+
+data Dragon mp = Dragon
+    { magicPower :: mp
+    }
```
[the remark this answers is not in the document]
callumfrance: oh man thats a good idea thanks bot

src/Chapter3.hs
```diff
@@ -1006,6 +1149,50 @@ Implement data types and typeclasses, describing such a battle between two
 contestants, and write a function that decides the outcome of a fight!
 -}
 
+data BattleResult = Victory | Defeat
+
+class Fighter x y where
+    attackEm :: x -> y -> y
+    runAway :: x -> y -> BattleResult
```
vrom911, line 1156: It would be enough to make the instance for each creature separately:
For that you can use constraints inside the typeclass as well:

```suggestion
class Fighter x where
    attackEm :: (Fighter y) => x -> y -> y
    runAway :: (Fighter y) => x -> y -> BattleResult
```

src/Chapter3.hs
```diff
@@ -1006,6 +1149,50 @@ Implement data types and typeclasses, describing such a battle between two
...
+
+instance Fighter Monster2 Knight2 where
+    attackEm :: Monster2 -> Knight2 -> Knight2
+    attackEm (Monster2 ma mh) (Knight2 kd ka kh) = Knight2 kd ka (kd + kh - ma)
+    runAway :: Monster2 -> Knight2 -> BattleResult
+    runAway (Monster2 ma mh) (Knight2 kd ka kh) = Victory
+
+-- Idk how to do this man
+
+-- roundOfFight :: BattleAction -> Monster2 -> Knight2 -> BattleResult
```
vrom911, line 1183: Don't worry! 👍🏼 

You can add the constraint about fighters and then use the methods from the typeclass.
For example:

```suggestion
roundOfFight :: (Fighter m, Fighter k) => BattleAction -> m -> k -> BattleResult
```

vrom911 (approved):
Great Chapter! And very nice types you constructed here 👷🏼 👍🏼

PR #307, opened 2020-10-29
First exercise Solutions

pastoreerrante:
### Solutions for Chapter {1}

cc @vrom911 @chshersh

Thanks for the awesome work, guys!

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -70,8 +70,10 @@ Each Haskell module starts with the "module <MODULE_NAME> where" line.
 Modules should have the same name as the corresponding file with
 the `.hs` extension.
 -}
+
 module Chapter1 where
 
+import Data.List
 {- |
 In Haskell, we have __expressions__. Expressions can be represented by some
 primitive values (numbers: 1, 100; characters: 'a', 'z'; booleans: True, False;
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -428,7 +430,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Num a => a -> a -> a
```
chshersh, line 433: We explain `Num` in the following chapters, so you are one step ahead 😸

src/Chapter1.hs
```diff
@@ -519,8 +522,15 @@ branches because it is an expression and it must always return some value.
...
   satisfying the check will be returned and, therefore, evaluated.
 -}
+
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = closest
+    where
+        absClosest :: Int
+        absClosest = min (abs x) (abs y)
+        closest    = if absClosest == abs x then x else y
```
chshersh, line 531: Very interesting solution! 💡 

src/Chapter1.hs
```diff
@@ -554,7 +564,9 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Integral a => a -> a -> a -> a
+mid x y z = (sort [x, y, z]) !! 1
```
chshersh, line 568: Clever solution! :+1:

src/Chapter1.hs
```diff
@@ -631,8 +652,12 @@ Implement a function that returns the sum of the last two digits of a number.
...
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = firstDigit + secondDigit
+  where
+    absN = abs n -- handle negative numbers
+    firstDigit  = mod absN  10
+    secondDigit = mod (div (absN   - firstDigit) 10) 10
```
chshersh, line 660: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
```

src/Chapter1.hs
```diff
@@ -653,7 +678,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Integral a => a -> a
+firstDigit n
+  | abs n `div` 10 == 0 = abs n
+  | otherwise           = firstDigit $ abs(n) `div` 10
```
chshersh, line 684: That is right 💯 

chshersh (approved):
Neat work on solving our challenges!

pastoreerrante:
Thank you, Dmitrii.

I know it's hard to do this kind of review work and you do it for free.

You (and Veronika) are truly awesome, thanks guys.

Daniele

Il giorno ven 30 ott 2020 alle ore 16:57 Dmitrii Kovanikov <
notifications@github.com> ha scritto:

chshersh:

PR #310, opened 2020-10-29
Chapter 1 completed

burdiuz:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -554,7 +556,12 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  |  y <= x && x <= z || y >= x && x >= z = x
+  |  x <= y && y <= z || x >= y && y >= z = y
+  |  x <= z && z <= y || x >= z && z >= y = z
+  | otherwise = x
```
vrom911, line 564: Yeah.
As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

Because of that, you needed to add another guard – `| otherwise = x`, to tell the compiler, that your pattern matching is exhaustive 🙂
However, you have already covered all possible option, so it is not necessary to check the conditions for `z` and just return `z`

```suggestion
  | otherwise = z
```

src/Chapter1.hs
```diff
@@ -632,7 +643,13 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n
+  | n >= 0 && n < 10 = n
+  | otherwise =
+    let one = mod (abs n) 10
+        two = div (mod (abs n) 100) 10
```
vrom911, line 651: This case actually covers the previous one, but yours would be faster on small numbers 👍🏼

src/Chapter1.hs
```diff
@@ -653,7 +670,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+  | n > 10 = firstDigit (div n 10)
+  | otherwise = n
```
vrom911, line 676: Unfortunately, this will return negative numbers on negative input...
Also, I think you could include `10` in the previous line comparison 👍🏼

vrom911 (approved):
Amazing! Congrats with finishing Chapter One 🎉

chshersh:

src/Chapter1.hs
burdiuz: thanks for the reply, on second attempt I've implemented this function using let a = abs n
```
firstDigit :: Int -> Int
firstDigit n =
  let a = abs n
      b = if a >= 10 then firstDigit (div a 10) else a
  in b
```

src/Chapter1.hs
burdiuz: thank you, I've updated it to
```
sumLast2 :: Int -> Int
sumLast2 n = 
  let a = abs n
      b = mod (div a 10) 10
      c = mod a 10
      d = b + c
  in d
```

PR #317, opened 2020-10-30
Solutions for Chapter {3}, without final task

pislas:
### Solutions for Chapter {3}
I'm a bit lost with the last task. But I am very interested in knowing how to solve it.
Coming from Scala( basic FP), I'm enjoying haskell. Thank you..
cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -343,7 +343,12 @@ Define the Book product data type. You can take inspiration from our description
 of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
-
+data Book = MkBook
+    { bookName   :: String
+    , bookCover  :: String
+    , bookAuthor :: String
+    , bookPages  :: Int
+    } deriving (Show)
 {- |
 =⚔️= Task 2
 
@@ -373,7 +378,21 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
-
+data Knight = MkKnight
+    { knightHealth :: Int
+    , knightAttack :: Int
+    , knightGold   :: Int
+    } deriving (Show)
+data Monster = MkMonster
+    { monsterHealth :: Int
+    , monsterAttack  :: Int
+    , monsterGold   :: Int
+    }
+fight:: Monster -> Knight -> Int
+fight m n
+  | knightAttack n >= monsterHealth m = knightGold n + monsterGold m
+  | monsterAttack m >= knightHealth n = (-1)
+  | otherwise = knightGold n
 {- |
 =🛡= Sum types
 
@@ -459,6 +478,12 @@ and provide more flexibility when working with data types.
 Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
+data Meal
+    = Breakfast
+    | Brunch
+    | Lunch
+    | Snack
+    | Dinner 
 
 {- |
 =⚔️= Task 4
@@ -479,7 +504,38 @@ After defining the city, implement the following functions:
    complicated task, walls can be built only if the city has a castle
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
+data Castle = Castle String | NoneCastle
+data Building = Church | Library
+data House = One | Two | Three | Four
+data Wall = Wall | NoWall
+data City = City{
+  cityCastle :: Castle,
+  cityWall :: Wall,
+  cityBuilding :: Building,
+  cityHouses :: [House]
+}
+buildCastle :: City -> String -> City
+buildCastle c n = c {cityCastle = Castle n}
 
+buildHouse :: City -> House -> City
+buildHouse c h = case c of
+  City _ _ _ hs -> c {cityHouses = h:hs}
+
+
+
+amountOfPeople :: [House] -> Int
+amountOfPeople hs = sum (map people hs)
+  where
+    people :: House -> Int
+    people One = 1
+    people Two = 2
+    people Three = 3
+    people Four = 4
+
+buildWalls :: City -> City
+buildWalls c = case (cityCastle c,amountOfPeople (cityHouses c)) of
+  (Castle _, i) -> if(i>=10) then c {cityWall = Wall} else c
+  _ -> c
 {-
 =🛡= Newtypes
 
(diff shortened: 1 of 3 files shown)


src/Chapter3.hs
```diff
@@ -479,7 +504,38 @@ After defining the city, implement the following functions:
...
    and at least 10 living __people__ inside in all houses of the city totally.
 -}
-
+data Castle = Castle String | NoneCastle
+data Building = Church | Library
+data House = One | Two | Three | Four
+data Wall = Wall | NoWall
+data City = City{
+  cityCastle :: Castle,
+  cityWall :: Wall,
```
vrom911, line 513: This is the correct data type, that can represent what we have in requirements 👍🏼 

src/Chapter3.hs
```diff
@@ -479,7 +504,38 @@ After defining the city, implement the following functions:
...
+  cityBuilding :: Building,
+  cityHouses :: [House]
+}
+buildCastle :: City -> String -> City
+buildCastle c n = c {cityCastle = Castle n}
+
+buildHouse :: City -> House -> City
+buildHouse c h = case c of
+  City _ _ _ hs -> c {cityHouses = h:hs}
```
vrom911, line 522: No need to use `case-of` here, as there is only one constructor, so you have a few options here:

1. 
    ```haskel
    buildHouse c h = c {cityHouses = h : cityHouses c}
    ```
    or
    ```haskel
    buildHouse c h = c {cityHouses = h : hs}
      where
        hs :: [House]
        hs = cityHouses c
    ```
2. `buildHouse (City _ _ _ hs) h = c {cityHouses = h:hs}` (but I won't recommend this one)
3. There is a `RecordWildCards` extension in Haskell, that allows you to use record field names as values:
    ```haskell
    buildHouse City{..} h = c {cityHouses = h : cityHouse}
    ```
You can learn more about the extension here: https://kodimensional.dev/recordwildcards

src/Chapter3.hs
```diff
@@ -479,7 +504,38 @@ After defining the city, implement the following functions:
...
+  where
+    people :: House -> Int
+    people One = 1
+    people Two = 2
+    people Three = 3
+    people Four = 4
+
+buildWalls :: City -> City
+buildWalls c = case (cityCastle c,amountOfPeople (cityHouses c)) of
+  (Castle _, i) -> if(i>=10) then c {cityWall = Wall} else c
```
vrom911, line 537: In Haskell functions and operators are always separated by spaces 👾 

```suggestion
  (Castle _, i) -> if i >= 10 then c {cityWall = Wall} else c
```

Also note, that no need to pattern match on the pair, as you will use amount of people only in the first case. 

```haskell
buildWalls c = case cityCastle c of
    Castle _ -> if people >= 10 then c {cityWall = Wall} else c
    _ -> c
  where
     people :: Int
     people = amountOfPeople (cityHouses c)
```

src/Chapter3.hs
```diff
@@ -560,22 +616,27 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Int
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = armor * dexterity
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Int -> Int -> Health -> Health
+calculatePlayerHit damage defense (Health health) = Health (health + defense - damage)
```
vrom911, line 639: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -909,7 +980,16 @@ Implement instances of "Append" for the following types:
 -}
 class Append a where
     append :: a -> a -> a
-
+newtype Gold = Gold Int
+instance Append Gold where
+  append (Gold i) (Gold j) = Gold (i+j)
```
vrom911, line 985: spaces 🙂 
```suggestion
  append (Gold i) (Gold j) = Gold (i + j)
```

src/Chapter3.hs
```diff
@@ -909,7 +980,16 @@ Implement instances of "Append" for the following types:
...
+instance Append [a] where
+  append i j = i++j
+instance (Append a) => Append (Maybe a) where
+  append (Just i) (Just j) = Just (append i j)
+  append (Just i) Nothing = Just i
+  append Nothing (Just j) = Just j
+  append _ _ = Nothing
```
vrom911, line 992: This is an absolutely correct solution!

src/Chapter3.hs
```diff
@@ -970,6 +1050,26 @@ implement the following functions:
...
+  | Friday
+  | Saturday
+  | Sunday deriving (Enum)
+isWeekend:: Day -> Bool
+isWeekend Saturday = True
+isWeekend Sunday = True
+isWeekend _ = False
+
+nextDay :: Day -> Day
+nextDay Sunday = Monday
```
vrom911, line 1067: That is right!
You could also use the instance of `Bounded` here:

```suggestion
    | day == maxBound = minBound
```

src/Chapter3.hs
```diff
@@ -970,6 +1050,26 @@ implement the following functions:
...
+nextDay d = succ d
+
+daysToParty :: Day -> Int
+daysToParty Friday = 0
+daysToParty d = 1 + daysToParty (nextDay d)
```
vrom911, line 1072: Interesting recursive solution!
Though you can try to implement it with the `Enum` instance 🙂

src/Chapter3.hs
```diff
@@ -1005,6 +1105,13 @@ properties using typeclasses, but they are different data types in the end.
 Implement data types and typeclasses, describing such a battle between two
 contestants, and write a function that decides the outcome of a fight!
 -}
+data Defence = Defence Int
+data KnightF  = KnightF Health Attack Defence
+data MonsterF = MonsterF Health Attack
+{-
+  I need to learn more to finish these task, i tried but i dont have clear ideas for now. 
+  I will come back after chapter 4.
```
vrom911, line 1113: No worries! This is quite big tasks with lots of concepts together 🙂 

In addition to `Knight` and `Monster` types, I would recommend creating separate types for `MonsterAction` and `KnightAction` to represent possible actions for each type correspondingly.

```haskell
class Fighter a where
```

```haskell

data FightResult a b
    = FirstWins a
    | SecondWins b

```

vrom911 (approved):
Wonderful Chapter ⭐ 
Lots of interesting types you constructed here 👷🏼

vrom911:

PR #320, opened 2020-10-31
Solutions for Chapter 1

recluse27:
### Solutions for Chapter 1

For task 6 a possible solution could be without using if-else-then:
`closestToZero x y = min (abs x) (abs y)`

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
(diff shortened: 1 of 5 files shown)


src/Chapter1.hs
```diff
@@ -490,8 +490,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
-
+lastDigit :: Int -> Int
+lastDigit n = n `mod` 10
```

src/Chapter1.hs
```diff
@@ -520,8 +520,8 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
-
+closestToZero x y = if abs x < abs y then x else y
+--closestToZero x y = min (abs x) (abs y)
```
vrom911, line 524: Unfortunately, this solution will return the absolute value instead of the value itself.
So the above one is the way to go 😉

src/Chapter1.hs
```diff
@@ -554,7 +554,14 @@ value after "=" where the condition is true.
...
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | x > y && y > z = y
+    | z > y && y > x = y
+    | y > x && x > z = x
+    | z > x && x > z = x
+    | x > z && z > y = z
+    | y > z && z > y = z
```

Here you covered all cases for sure. So if pattern matching didn't work on previous 4 cases, then you can say even without condition checking that the answer would be `z`:

```suggestion
    | otherwise = z
```

src/Chapter1.hs
```diff
@@ -631,9 +644,13 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n =
+    let last = n `mod` 10
+        prelast = (n `mod` 100)
```
vrom911, line 650: This is the right idea 💡  Though you need to consider the case of the negative numbers 🙂

src/Chapter1.hs
```diff
@@ -652,8 +669,8 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n = if n > 10 then firstDigit(n `div` 10) else n
```
vrom911, line 673: I see, that for a negative number like -19, this implementation will return the negative number itself instead of the last digit because the first check n <= 10 will always succeed for negative numbers.

vrom911 (approved):
Great one! Congrats with finishing Chapter One 🥳 👏🏼

chshersh:

PR #324, opened 2020-10-31
Chapter3

gantu:

cc @vrom911 @chshersh
Thanks again. It was really fun, I learned lot of things.

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -616,22 +616,30 @@ introducing extra newtypes.
 🕯 HINT: if you complete this task properly, you don't need to change the
     implementation of the "hitPlayer" function at all!
 -}
+
+newtype Health = Health Int
+newtype Armor = Armor Int
+newtype Attack = Attack Int
+newtype Dexterity = Dexterity Int
+newtype Strength = Strength Int
+newtype Demage = Demage Int
+newtype Defence = Defence Int
 data Player = Player
-    { playerHealth    :: Int
-    , playerArmor     :: Int
-    , playerAttack    :: Int
-    , playerDexterity :: Int
-    , playerStrength  :: Int
+    { playerHealth    :: Health
+    , playerArmor     :: Armor
+    , playerAttack    :: Attack
+    , playerDexterity :: Dexterity
+    , playerStrength  :: Strength
     }
 
-calculatePlayerDamage :: Int -> Int -> Int
-calculatePlayerDamage attack strength = attack + strength
+calculatePlayerDamage :: Attack -> Strength -> Demage
+calculatePlayerDamage (Attack attack) (Strength strength) = Demage $ attack + strength
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defence
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defence $ armor * dexterity
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Demage -> Defence -> Health -> Health
+calculatePlayerHit (Demage damage) (Defence defense) (Health health) = Health $ health + defense - damage
 
 -- The second player hits first player and the new first player is returned
 hitPlayer :: Player -> Player -> Player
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -480,6 +504,38 @@ After defining the city, implement the following functions:
...
+data House = House {people :: People}
+data People = One | Two | Three | Four
+
+buildCastle :: City -> String -> City
+buildCastle city name = case castle city of
+  CastleWithWall _ -> city {castle = CastleWithWall name}
+  _ -> city {castle = Castle name}
+
+buildHouse :: City -> House -> City
+buildHouse city house = city {houses = house : houses city}
```
chshersh, line 524: According to our task, each house can contain from 1 to 4 people, But your `House` data type is a wrapper around `Int`, which means that users of this function can put any number of people inside the `House` (even the negative number). You can restrict this by introducing more checks.

This approach is usually called "smart constructor". To avoid building houses with more than 4 people you need to restrict usage of the Constructor and allow to create `House`s only with your `mkHouse` function.

src/Chapter3.hs
```diff
@@ -560,22 +616,30 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defence
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defence $ armor * dexterity
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Demage -> Defence -> Health -> Health
+calculatePlayerHit (Demage damage) (Defence defense) (Health health) = Health $ health + defense - damage
```
chshersh, line 642: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -910,6 +985,22 @@ Implement instances of "Append" for the following types:
 class Append a where
     append :: a -> a -> a
 
+newtype Gold = Gold Int
+
+instance Append Gold where
+  append :: Gold -> Gold -> Gold
+  append (Gold a)(Gold b) = Gold $ a + b
```
chshersh, line 992: It's usually a good idea to separate patterns by space :arrow_left: :space_invader: :arrow_right: 
```suggestion
  append (Gold a) (Gold b) = Gold $ a + b
```

src/Chapter3.hs
```diff
@@ -910,6 +985,22 @@ Implement instances of "Append" for the following types:
...
+
+instance Append [a] where
+  append :: [a] -> [a] -> [a]
+  append = (++)
+
+instance Append a => Append (Maybe a) where
+  append (Just a) (Just b) = Just (append a b)
+  append Nothing (Just b) = Just b
+  append (Just a) Nothing = Just a
+  append Nothing Nothing = Nothing
```
chshersh, line 1002: This is an absolutely correct solution!

src/Chapter3.hs
```diff
@@ -970,7 +1061,25 @@ implement the following functions:
 
 🕯 HINT: to implement this task, derive some standard typeclasses
 -}
-
+data Weekday = MON | TUE | WED | THU | FRI | SAT | SUN deriving (Enum, Show, Eq)
+instance Bounded Weekday where
+  minBound = MON
+  maxBound = SUN
```
chshersh, line 1067: The `Bounded` typeclass is one of the standard typeclasses and it can be derived as well :slightly_smiling_face: Moreover, it would be preferable to derive it, because if you change the data type definition (to the "American" system, for instance), you can forget to update the manual instance, but the derived instance always remains up-to-date.

src/Chapter3.hs
```diff
@@ -970,7 +1061,25 @@ implement the following functions:
...
+nextDay :: Weekday -> Weekday
+nextDay a | a == maxBound = minBound
+          | otherwise = succ a
+
+daysToParty :: Weekday -> Int
+daysToParty a = go a 0
+  where
+    go :: Weekday -> Int -> Int
+    go a b | a == FRI = b
+           | otherwise = go (nextDay a) (b + 1)
```
chshersh, line 1082: That's the correct implementation! Though, you can implement slightly more efficiently by using the `fromEnum` method of the `Enum` typeclass

chshersh (approved):
Neat work on solving Chapter Three tasks!

vrom911:

PR #325, opened 2020-10-31
Chapter 1 finished

bendo:

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -428,7 +428,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Num a => a -> a -> a
```

src/Chapter1.hs
```diff
@@ -554,7 +558,8 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: (Ord a) => a -> a -> a -> a
+mid x y z = maximum [min x y, min x z, min y z]
```
vrom911, line 562: Very creative solution 🧠 💡 🎨

src/Chapter1.hs
```diff
@@ -568,7 +573,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c = elem c ['a','e','i','o','u']
```
vrom911, line 577: Nice one! 👍🏼 

vrom911 (approved):

vrom911:

PR #333, opened 2020-11-02
complete the Chapter1, passed all tests.

luigidisotto:

cc @vrom911 @chshersh

Enjoyed the Chapter 1! It would be nice to discuss ways to improve my primeval Haskell code! Thank you guys!

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -523,7 +523,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = if abs (x) < abs (y) then x else y 
+closestToZero x y = if abs x < abs y then x else y 
 
 
 {- |
@@ -563,8 +563,7 @@ mid x y z
     | (z <= y) && (y <= x) = y
     | (y <= x) && (x <= z) = x
     | (z <= x) && (x <= y) = x
-    | (x <= z) && (z <= y) = z
-    | (y <= z) && (z <= x) = z
+    | otherwise = z
 
 {- |
 =⚔️= Task 8
@@ -651,12 +650,12 @@ specifying complex expressions.
 
 sumLast2 :: Int -> Int
 sumLast2 n = 
-  let last = lastDigit n
-      sec2last = secondToLastDigit n
+  let lastit = lastDigit n
+      sec2lastit = secondToLastDigit n
         where
           secondToLastDigit :: Int -> Int
           secondToLastDigit n = lastDigit (div (abs n) 10)
-  in last + sec2last
+  in lastit + sec2lastit
 
 
 {- |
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -520,7 +523,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if abs (x) < abs (y) then x else y
```
vrom911, line 526: Nice one 👌🏼 
Note that these parentheses are not required and you can omit it in this case 🙂

src/Chapter1.hs
```diff
@@ -554,7 +557,14 @@ value after "=" where the condition is true.
...
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | (x <= y) && (y <= z) = y
+    | (z <= y) && (y <= x) = y
+    | (y <= x) && (x <= z) = x
+    | (z <= x) && (x <= y) = x
+    | (x <= z) && (z <= y) = z
+    | (y <= z) && (z <= x) = z
```
vrom911, line 567: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

However, as you rightly noticed, you covered __all__ possible situations with these checks, so if the pattern matching won't work on the first 4 cases, then you can say for sure, that the answer is `z` even without additional checks (as they will definitely succeed):

```haskell
mid x y z
    | (x <= y) && (y <= z) = y
    | (z <= y) && (y <= x) = y
    | (y <= x) && (x <= z) = x
    | (z <= x) && (x <= y) = x
    | otherwise = z
```

src/Chapter1.hs
```diff
@@ -632,7 +649,14 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = 
+  let last = lastDigit n
```
vrom911, line 654: `last` is the name of a standard function. Haskell allows name shadowing, that's why it works. Though, in practice, it's usually better to avoid name clashes with other functions cause it could be confusing 🙂

src/Chapter1.hs
```diff
@@ -653,8 +677,13 @@ You need to use recursion in this task. Feel free to return to it later, if you
...
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+isDigit :: Int -> Bool
+isDigit n = n >= 0 && n <= 9
 
+firstDigit :: Int -> Int
+firstDigit n
+    | isDigit n = n
+    | otherwise = firstDigit (div n 10)
```
vrom911, line 686: That is nice 💯 
Although, you might want to use `abs n` here instead of just `n` to handle negative cases 🙂

vrom911 (approved):
Awesome work! Congrats with finishing chapter One 🎉 
Very nice solutions 🚀

vrom911:
The solutions look awesome :+1: I'm going to close this PR because we can't merge the solutions to our repo :slightly_smiling_face: 
But feel free to open PRs with solutions to other chapters!

PR #334, opened 2020-11-02
Second exercise solutions

pastoreerrante:
### Solutions for Chapter 2
cc @vrom911 @chshersh

Thanks for the awesome work!

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,44 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
+reverse :: [a] -> [a]
 
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
@@ -186,31 +187,31 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
-
+[10,2,3,1,5]
 >>> [] ++ [1, 4]  -- [] is an empty list
-
+[1,4]
 >>> 3 : [1, 2]
-
+[3,1,2]
 >>> 4 : 2 : [5, 10]  -- prepend multiple elements
-
+[4,2,5,10]
 >>> [1 .. 10]  -- list ranges
-
+[1,2,3,4,5,6,7,8,9,10]
 >>> [10 .. 1]
-
+[]
 >>> [10, 9 .. 1]  -- backwards list with explicit step
-
+[10,9,8,7,6,5,4,3,2,1]
 >>> length [4, 10, 5]  -- list length
-
+3
 >>> replicate 5 True
-
+[True,True,True,True,True]
 >>> take 5 "Hello, World!"
-
+"Hello"
 >>> drop 5 "Hello, World!"
-
+", World!"
 >>> zip "abc" [1, 2, 3]  -- convert two lists to a single list of pairs
-
+[('a',1),('b',2),('c',3)]
 >>> words "Hello   Haskell     World!"  -- split the string into the list of words
-
+["Hello","Haskell","World!"]
 
 
 👩‍🔬 Haskell has a lot of syntax sugar. In the case with lists, any
@@ -336,7 +337,11 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList start stop xs
+  | start < 0 || stop < 0 = [] -- return an empty list when given numbers are negative.
+  | start > stop          = [] -- >>> subList 3 0 [True, False, False, True, False] = []
+  | otherwise             = drop start $ take (stop + 1 ) xs
+
 
 {- |
 =⚔️= Task 4
@@ -349,8 +354,8 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 -- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
-
+firstHalf :: [a] -> [a]
+firstHalf l = fst $ splitAt (length l `div` 2) l
 
 {- |
 =🛡= Pattern matching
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +337,11 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList start stop xs
+  | start < 0 || stop < 0 = [] -- return an empty list when given numbers are negative.
+  | start > stop          = [] -- >>> subList 3 0 [True, False, False, True, False] = []
```
vrom911, line 342: Nice catch!
This could be spatted in the next case, but it is better to have this check separately for performance!
Very cool 😎

src/Chapter2.hs
```diff
@@ -349,8 +354,8 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 -- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
-
+firstHalf :: [a] -> [a]
+firstHalf l = fst $ splitAt (length l `div` 2) l
```
vrom911, line 358: Interesting solution 🎨 

You also can use `take` to take only `length l `div` 2` elements 👍🏼

src/Chapter2.hs
```diff
@@ -621,8 +627,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
-
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven (x:_:xs) = x : takeEven xs -- base pattern
+takeEven (x:_)    = x : takeEven []
```
vrom911, line 633: This case is actually is needed to cover `[x]` case, as anything else will pattern match earlier. 🙂 
So, you can avoid additional function call:

```suggestion
takeEven [x]    = [x]
```

src/Chapter2.hs
```diff
@@ -728,7 +736,11 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate []     = []
+smartReplicate (x:xs) = replicate x x ++ smartReplicate xs
```
vrom911, line 740: This is right! 💯 

```haskell
[3,1,2,4,0,1]
```

src/Chapter2.hs
```diff
@@ -728,7 +736,11 @@ value of the element itself
...
+
+smartReplicate2 :: [Int] -> [Int]
+smartReplicate2 xs = concat $ map (\x -> replicate x x) xs
```
vrom911, line 743: Oh, sorry, I haven't noticed this one 😄 
Yes, this is correct as well, you can also use `concatMap` 👍🏼

src/Chapter2.hs
```diff
@@ -741,8 +753,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: Int -> [[Int]] -> [[Int]]
+contains item xxs = filter (\xs -> elem item xs) xxs
```
vrom911, line 757: Now, after you've mastered eta-reduction, you can apply such a technique to this and lambda function both as well  🙂 

src/Chapter2.hs
```diff
@@ -781,13 +793,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
-listElementsLessThan x l = filter (< x) l
-
+listElementsLessThan :: Int -> [Int] -> [Int]
+-- listElementsLessThan x l = filter (< x) l
+listElementsLessThan x = filter (< x)
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+-- pairMul xs ys = zipWith (*) xs ys
+pairMul :: Num a => [a] -> [a] -> [a] 
+pairMul = zipWith (*)
```
vrom911, line 804: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -842,7 +856,11 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate x xs
+  | x < 0     = []
+  | null xs   = []
+  | otherwise = drop x $ take ((length xs) + x) $ cycle xs
```
vrom911, line 863: Great!
Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. By using `mod x len` you could optimize this 🚤

src/Chapter2.hs
```diff
@@ -858,7 +876,11 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind xs = lastElement : rewind (init xs)
+    where
+       lastElement = xs !! ((length xs) - 1)
```
vrom911, line 883: Your solution is correct! However, it is slow 🐌 
 In lists, it is quite slow to add anything at the end of the list, take at the position, or take the length. That is why it is always better to, for example, use `:` cons instead of `++`.  Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Great Chapter! very neat and thoughtful solutions 👏🏼

src/Chapter2.hs
pastoreerrante: Thank you for the insights! my solutions tend to be a little bit complex than they need. 

Regarding the recursive-go pattern, I already saw that but it was called tail recursion, like this: https://youtu.be/_JtPhF8MshA

 Anyway, I fail to see how to apply tail-recursion/recursive go pattern to the rewind function. I mean, in the end I'll need to access to the last element anyway. I don't understand how tail-recursion can mitigate/avoid this issue.

Can you kindly show me that? Thanks for your awesome effort!!!

vrom911: Sure! I will try to explain with the reasoning at each step 🙂 

Taking element by index in lists in Haskell is not a cheap operation, due to the way lists are implemented. You need to go through each element until you meet the required position. That is why the solution with `!!` would not be efficient as well.

In Haskell, all such problems in lists are solved with the `:` cons and pattern matching.
Instead of the index, let's have as an accumulator the new list that we will construct on each recursive step.

```haskell
rewind :: [Int] -> [Int]
rewind l = go [] l
  where
    go :: [Int] -> [Int] -> [Int]
    go acc l
```

and now the most interesting begins.
Let's start our recursive function with the stop case. As we will go element by element, our final stop will be the empty list. In that case, we assume that the accumulator already has all elements reversed, so we return it:

```haskell
go acc [] = acc
```

Let's now look at the general case. Now it would be the mind-blowing moment (at least it was for me 🤯 ):

```haskell
go acc (x:xs) = go (x:acc) xs
```

So what we are doing? We are adding the elements to the start of the accumulator.

Let's prove that it will be reversed in the accumulator (even though we add to the beginning) on a small example:

| acc | list |
|-----|----|
| [] | [1,2,3,4]|
| 1: [] | [2,3,4] |
| 2 : 1 : [] | [3,4] |
| 3 : 2 : 1 : [] | [4] |
| 4 : 3 : 2 : 1 : [] | [] |

So at the last step, the list is reversed, and the function works only by one walk-through the list!

Hope it makes sense, let me know if you have questions 🙂

pastoreerrante: Ah yes! Now I see how it works, brilliant. Thank you for the crystal-clear explanation!

vrom911:
Feel free to open PRs with solutions to other chapters! 🎈

PR #342, opened 2020-11-11
third exercise solutions

pastoreerrante:
### Solutions for Chapter 3

cc @vrom911 @chshersh

I wasn't able to defeat the final boss :-(

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -50,6 +50,7 @@ signatures in places where you can't by default. We believe it's helpful to
 provide more top-level type signatures, especially when learning Haskell.
 -}
 {-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE RecordWildCards #-}
 
 module Chapter3 where
 
@@ -344,6 +345,12 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+data Book = MkBook
+    { bookName      :: String
+    , bookAuthor    :: String
+    , bookPages     :: Int
+    } deriving (Show)
+
 {- |
 =⚔️= Task 2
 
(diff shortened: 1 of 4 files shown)


src/Chapter3.hs
```diff
@@ -1005,8 +1178,10 @@ properties using typeclasses, but they are different data types in the end.
 Implement data types and typeclasses, describing such a battle between two
 contestants, and write a function that decides the outcome of a fight!
 -}
-
-
+-- sorry, a bit too foggy for me :-( having hard time translating this "knight and monster domain" into typeclasses.
+-- doubts: 1) knights and monsters do their activities *in turns* but then each fighter has a list of actions
+-- to be performed *in sequence* and then again from the beginning...it seems a bit contradicting to me.
+-- 2) what's the meaning of action like "drink Health potion" or "cast a spell"?
```
chshersh, line 1184: I can give a few hints, so you can crack the advanced task 🙂 

For this task you need to create two data types: `Monster` and `Knight`. Each type will have the properties described in the task description.

In addition, I would recommend creating separate types for `MonsterAction` and `KnightAction` to represent possible actions for each type correspondingly.

Regarding the list of actions, let's look at the example. A `Knight` can have initially 3 actions:

`Attack, DrinkPotion, Attack`

So, if the `Knight` survives 8 turns, their actions will be: 

`Attack, DrinkPotion, Attack, Attack, DrinkPotion, Attack, Attack, DrinkPotion`

`DrinkHealthPotion` should increase the health of a `Knight` and `Cast a spell` should increase defence of a knight, as written in the task description: `cast a spell to increase their defence`

```haskell
class Fighter a where
```

```haskell

data FightResult a b
    = FirstWins a
    | SecondWins b

```

chshersh (approved):
Nice work! 👍🏻 I left some comments with suggestions and a few hints on how to improve solutions and move forward 🙂

vrom911:
@pastoreerrante I think this is ready one 👍🏼 I am going to close this PR, but feel free to reach out to us if you have any questions 😊 Don't be a stranger 🤗 

And Merry Christmas 🎄

PR #343, opened 2020-11-13
Typos Chapters 1, 2, 3 & 4

ARPL-c-63h:
### Pull request on main. No solutions

cc @vrom911

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -22,7 +22,7 @@ Here is how it works:
    task proposed to consolidate the results.
  ★ Create a PR with solutions to as many tasks as you can after each Chapter is
    finished.
- ★ Make sure that tests for the Chapter you are working at are passing.
+ ★ Make sure that the tests for the Chapter you are working on are passing.
  ★ Receive some feedback and suggestions from us.
  ★ Merge the PR.
  ★ Go to the next Chapter.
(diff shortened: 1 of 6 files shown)


src/Chapter1.hs
```diff
@@ -284,7 +284,7 @@ ghci> 1 + 2
 
 > It may look scary to you, but we will cover all this 'Num' and "=>" later. For
   now, you can think of this as a polymorphic function — in this case, the
-  operator, that can work with any numeric types, including 'Int's, 'Double's,
```
vrom911, line 287: This is the haddock syntax to have a link on `Int` documentation. As these are Haskell comment format of info we decided to go with it 🙂

vrom911 (approved):
Thanks, @ARPL-c-63h! 👍🏼 
Could you please remove the additional file, so we can merge the improvements right away? 🙂

src/Chapter1.hs
ARPL-c-63h: Makes sense. Didn't know that. Not much experience with Haddock. Better look into that. : )

ARPL-c-63h (commented):
File deleted.

chshersh (approved):
Thanks! 👍🏻

ARPL-c-63h:
@chshersh: You're welcome. I was a bit hesitant since we got off on the wrong foot (because of [this](https://github.com/kowainik/relude/pull/300)). Glad it worked out this time.

chshersh:
@ARPL-c-63h No worries, no hard feelings left! I appreciate you contributing improvements to our projects 🙂

ARPL-c-63h:
@chshersh And you, more than anyone else, know that it is almost impossible to resist the cuteness that is @vrom911 ; )
And now I've probably overstepped a whole different set of boundaries! Ah well : )
Until next time!

PR #344, opened 2020-11-16
Chapter1

marcellourbani:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -518,11 +518,7 @@ mid x y z
 -- False
 isVowel :: Char -> Bool
 isVowel c
-  | c == 'a' = True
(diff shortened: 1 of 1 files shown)


vrom911:
Hey @marcellourbani ! Happy to have you with us on this course 🤗 

Would it be possible for you to configure your autoformatter not to change the style of comments?
It's easier for us when the diff is only for solutions, so we won't miss any questions and changes while the review :slightly_smiling_face:

marcellourbani:
Hi @vrom911 Yes sure, will open another one with the right formatting.
Already had half an idea to do it.

Thanks
Marcello

PR #353, opened 2020-11-19
chapter 1 and 2 solutions

minimario:
### Solutions for Chapter 1

Hi @vrom911 @chshersh, thanks for putting this together! I'd love to you to take a look at my solutions and suggest some improvements! I've been playing around with Haskell for a few months, but I'm not too familiar with best practices/coding idiomatically :)

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -653,7 +667,11 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+  | n < 0 = firstDigit (-n)
+  | n < 10 = n
+  | otherwise = firstDigit $ div n 10
```
vrom911, line 674: Neat advanced task solution 👏🏼

src/Chapter2.hs
```diff
@@ -336,7 +347,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList l r xs =
+  if l<0 || r<0 || l>r then []
```
vrom911, line 351: In Haskell, functions and operators are always separated from arguments with spaces 👾 

```suggestion
  if l < 0 || r < 0 || l > r then []
```

src/Chapter2.hs
```diff
@@ -336,7 +347,9 @@ from it!
...
+  else take (r-l+1) $ drop l xs
```
vrom911, line 352: And here as well 🙂 
```suggestion
  else take (r - l + 1) $ drop l xs
```

src/Chapter2.hs
```diff
@@ -348,8 +361,10 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+
+firstHalf :: [a] -> [a]
+firstHalf l =
+  take len l where len = div (length l) 2
```
vrom911, line 367: Nice usage of `where` construction 👌🏼

src/Chapter2.hs
```diff
@@ -501,8 +516,9 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
-
+isThird42 :: [Int] -> Bool
+isThird42 (_:_:a:_) = (a == 42)
```
vrom911, line 520: You don't need to compare `z` with 42 separately, you can pattern match on `42` directly in the pattern.

```suggestion
isThird42 (_:_:42:_) = True
```

src/Chapter2.hs
```diff
@@ -621,7 +637,11 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+
+takeEven :: [Int] -> [Int]
+takeEven [] = []
+takeEven [x] = [x]
+takeEven (x:_:xs) = x:takeEven xs
```
vrom911, line 644: Great pattern matching skills 💪🏼

src/Chapter2.hs
```diff
@@ -728,7 +748,8 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = 
+  concat $ map (\x->replicate x x) l
```
vrom911, line 752: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -741,7 +762,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains e xs = filter (\x -> elem e x) xs
```
vrom911, line 766: Now, after you've mastered eta-reduction, you can apply such a technique to this function and the lambda function inside as well  🙂 

src/Chapter2.hs
```diff
@@ -781,13 +803,13 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
-divideTenBy x = div 10 x
+divideTenBy = div 10
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul = zipWith (*)
```
vrom911, line 812: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -858,7 +883,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = (rewind xs) ++ [x]
```
vrom911, line 888: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Fantastic work! 👏🏼 
Congrats with finishing two Chapters 🎉

chshersh:
@minimario I'm going to close this PR, as we can't merge it to our repo 🙂

PR #355, opened 2020-11-20
Chapter3

minimario:
### Solutions for Chapter 3

cc @vrom911 @chshersh

A little stuck on task 9 though, not sure what goes inside the typeclass instance as opposed to the individual instances of the knight and monster!

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -68,6 +68,7 @@ Haskell has several different ways to create entirely new data types. Let's talk
 about them all and master our skill of data types construction.
 -}
 
+import qualified Data.Maybe
 {- |
 =🛡= Type aliases
 
@@ -344,6 +345,12 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+data Book = MkBook {
+  bookAuthor :: String,
+  bookTitle :: String
+}
+
+
 {- |
 =⚔️= Task 2
 
@@ -373,6 +380,25 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+
+data Monster = MkMonster {
+  monsterHealth :: Int,
+  monsterAttack :: Int,
+  monsterGold :: Int
+}
+
+data Knight = MkKnight {
+  knightHealth :: Int,
+  knightAttack :: Int,
+  knightGold :: Int
+}
+
+fight :: Knight -> Monster -> Int
+fight MkKnight { knightHealth = kh, knightAttack = ka, knightGold = kg} 
+      MkMonster { monsterHealth = mh, monsterAttack = ma, monsterGold = mg}
+        | ka >= mh = mh + mg
+        | ma >= kh = -1
+        | otherwise = kg
 
 {- |
 =🛡= Sum types
@@ -460,6 +486,11 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data Meal 
+  = Breakfast Int -- Breakfast calorie amount
+  | Lunch Int
+  | Dinner Int
+
 {- |
 =⚔️= Task 4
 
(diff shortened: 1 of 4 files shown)


src/Chapter3.hs
```diff
@@ -374,6 +381,25 @@ after the fight. The battle has the following possible outcomes:
...
+  knightAttack :: Int,
+  knightGold :: Int
+}
+
+fight :: Knight -> Monster -> Int
+fight MkKnight { knightHealth = kh, knightAttack = ka, knightGold = kg} 
+      MkMonster { monsterHealth = mh, monsterAttack = ma, monsterGold = mg}
+        | ka >= mh = mh + mg
+        | ma >= kh = -1
+        | otherwise = kg
```
chshersh, line 401: That's a great and correct solution! 👍🏻 You can also use a Haskell feature called `RecordWildCards` to reduce some code, as the field names are unambiguous. I wrote a blog post about feature, if you are interested 🙂 

* https://kodimensional.dev/recordwildcards

src/Chapter3.hs
```diff
@@ -480,6 +511,47 @@ After defining the city, implement the following functions:
    and at least 10 living __people__ inside in all houses of the city in total.
 -}
 
+data City = City {
+  castle :: Maybe (Castle, Maybe Wall),
```
chshersh, line 515: That's a correct representation! 👍🏻 Though, I feel like it's a bit awkward to work with this type. You can create a custom data type to model the same structure:

```haskell
data Castle
    = NoCastle
    | Castle String
    | CastleWithWalls String
```

Creating custom types in Haskell is cheap, so no harm in introducing custom domain data types 🙂

src/Chapter3.hs
```diff
@@ -560,22 +632,31 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defense (armor * dexterity)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage damage) (Defense defense) (Health health) = Health (health + defense - damage)
```
chshersh, line 659: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -907,9 +1001,23 @@ Implement instances of "Append" for the following types:
...
+    append (Gold a) (Gold b) = Gold (a+b)
+
+instance Append [a] where 
+    append :: [a] -> [a] -> [a]
+    append = (++)
+
+instance (Append a) => Append (Maybe a) where
+    append :: Maybe a -> Maybe a -> Maybe a
+    append (Just x) (Just y) = Just (append x y)
+    append _ _ = Nothing
```
chshersh, line 1020: While this is valid implementation and nothing in the task description says differently, in reality it would be more beneficial is this function returned `Just` when at least one argument is `Just`.

src/Chapter3.hs
```diff
@@ -971,6 +1079,23 @@ implement the following functions:
...
+data Day = Monday | Tuesday | Wednesday | Thursday | Friday | Saturday | Sunday
+  deriving (Show, Eq, Enum, Bounded)
+
+isWeekend :: Day -> Bool
+isWeekend Saturday = True
+isWeekend Sunday = True
+isWeekend _ = False
+
+nextDay :: Day -> Day
+nextDay day = if day == maxBound then minBound else succ day
```
chshersh, line 1091: That is right!

src/Chapter3.hs
```diff
@@ -971,6 +1079,23 @@ implement the following functions:
...
+
+daysToParty :: Day -> Int
+daysToParty Sunday = 5
+daysToParty Saturday = 6
+daysToParty day = fromEnum Friday - fromEnum day
```
chshersh, line 1096: Nice solution with pattern-matching and usage of `fromEnum`! 👍🏻 You can write a solution even shorter if you use `mod` (and the fact of how it works for negative numbers)

src/Chapter3.hs
```diff
@@ -1006,6 +1131,33 @@ Implement data types and typeclasses, describing such a battle between two
...
+  knightDefense :: Int
+} deriving (Show)
+
+data T9_Monster = Monster {
+  monsteHealth :: Int,
+  monsterAttack :: Int
+}
+class Fighter a where
+  attack :: Int
+  health :: Int
```
chshersh, line 1158: That's a good start! I can provide some hints for this task to help you move forward 🙂 

For this task you need to create two data types: `Monster` and `Knight` (as you already correctly did). Each type will have the properties described in the task description.

In addition, I would recommend keeping separate types for `MonsterAction` and `KnightAction` to represent possible actions for each type correspondingly.

```haskell
class Fighter a where
```

```haskell

data FightResult
    = FirstWins
    | SecondWins

fight :: (Fighter a, Fighter b) => a -> b -> FightResult
```

Your current typeclass methods don't take values of type `a` as arguments. Which means that they are constant for all values. To patch the typeclass definition, you need to write it like this:

```haskell
class Fighter a where
  attack :: a -> Int
  health :: a -> Int
```

See, how methods are functions from values to fields you want to get? You can try implementing a generic `Fight` function, so you will be able to see, what methods you need to add to the `Fighter` typeclass.

chshersh (approved):
Great solutions for Chapter Three! 👏🏻 

I left some feedback with helpful information and general suggestions 🙂

vrom911:
@minimario I am going to close this PR, but feel free to reach out to us if you have any questions 😊 Don't be a stranger 🤗 

And Merry Christmas 🎄

PR #360, opened 2020-11-21
Chapter4

pastoreerrante:
### Solutions for Chapter 4

cc @vrom911 @chshersh

Thanks guys, somehow I made it, even if one month late...

You are awesome!!!

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -114,23 +114,23 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
-
+Char :: *
 >>> :k Bool
-
+Bool :: *
 >>> :k [Int]
-
+[Int] :: *
 >>> :k []
-
+[] :: * -> *
 >>> :k (->)
-
+(->) :: * -> * -> *
 >>> :k Either
-
+Either :: * -> * -> *
 >>> data Trinity a b c = MkTrinity a b c
 >>> :k Trinity
-
+Trinity :: * -> * -> * -> *
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
-
+IntBox :: (* -> *) -> *
 -}
 
 {- |
@@ -293,7 +293,8 @@ values and apply them to the type level?
 -}
 instance Functor (Secret e) where
     fmap :: (a -> b) -> Secret e a -> Secret e b
-    fmap = error "fmap for Box: not implemented!"
+    fmap f (Reward x) = Reward $ f x
+    fmap _ (Trap x) = Trap x
 
 {- |
 =⚔️= Task 3
@@ -305,7 +306,14 @@ typeclasses for standard data types.
 -}
 data List a
     = Empty
-    | Cons a (List a)
+    | Cons a (List a) deriving (Show)
+
+instance Functor List where
+    fmap :: (a -> b) -> List a -> List b
+    fmap f (Cons x a) = Cons (f x) (fmap f a)
+    fmap _ Empty = Empty
+
+
 
 {- |
 =🛡= Applicative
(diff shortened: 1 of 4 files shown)


src/Chapter4.hs
```diff
@@ -472,10 +480,11 @@ Implement the Applicative instance for our 'Secret' data type from before.
 -}
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure x = Reward x
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    Trap x <*> _ = Trap x
+    Reward f <*> x = fmap f x
```
chshersh, line 487: Nice work on reusing `Functor` instance for `Secret`! :world_map:

src/Chapter4.hs
```diff
@@ -488,6 +497,27 @@ Implement the 'Applicative' instance for our 'List' type.
   may also need to implement a few useful helper functions for our List
   type.
 -}
+-- helper to join two List together
+join :: List a -> List a -> List a
+join l1 Empty = l1
+join Empty l2 = l2
+join (Cons x xs) l2 = Cons x $ join xs l2
```
chshersh, line 504: Usually this function is called `append`, and `join` in Haskell is a function of type `List (List a)` or similar 🙂

src/Chapter4.hs
```diff
@@ -488,6 +497,27 @@ Implement the 'Applicative' instance for our 'List' type.
...
+
+instance Applicative List where
+    pure :: a -> List a
+    pure x = Cons x Empty
+
+    (<*>) :: List (a -> b) -> List a -> List b
+    Empty <*> _ = Empty
+    (Cons f fs) <*> xs = join (fmap f xs) (fs <*> xs)
```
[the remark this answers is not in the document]
chshersh: That's actually an HLint bug, because your `join` is not the standard `join` and HLint doesn't see it 🙂

src/Chapter4.hs
```diff
@@ -629,7 +666,19 @@ Can you implement a monad version of AND, polymorphic over any monad?
...
+
+andM2 :: (Monad m) => m Bool -> m Bool -> m Bool
+andM2 m1 m2 =
+    m1 >>= (\x -> m2 >>= (\y -> return $ x && y))
+
+andM3 :: (Monad m) => m Bool -> m Bool -> m Bool
+andM3 m1 m2 = do
+    x <- m1
+    y <- m2
+    return $ x && y
```
chshersh, line 680: Great implementations! All these functions are semantically equivalent ↔️ 

Though, your current implementation doesn't pass all tests. The ordinary `&&` if lazy on the second argument. Meaning that if it meets `False` as the first argument, there is no need to calculate another arg. In `andM` we would like to keep the same behaviour, so the

```
Just False `andM` _
```

src/Chapter4.hs
```diff
@@ -673,6 +722,29 @@ Specifically,
...
+    fmap f (Leaf a) = Leaf $ f a
+    fmap f (Node t1 a t2) = Node (fmap f t1) (f a) (fmap f t2)
+
+-- alg:
+-- 1. Call reverseTree for left-subtree.
+-- 2. Call reverseTree for right-subtree.
+-- 3. Swap left and right subtrees.
+reverseTree :: Tree a -> Tree a
+reverseTree (Leaf a) = Leaf a
+reverseTree (Node (Leaf x) a (Leaf y)) = (Node (Leaf y) a (Leaf x))
```
chshersh, line 742: This case is actually redundant, because it's a special case of your next line. So you can safely remove it 🧹

src/Chapter4.hs
```diff
@@ -673,6 +722,29 @@ Specifically,
  ❃ Implement the function to convert Tree to list
 -}
 
+data Tree a = Leaf a
+            | Node (Tree a) a (Tree a) deriving (Show, Eq)
```
chshersh, line 726: That's one possible implementation of `Tree` and it's valid 👍🏻 But it would be more convenient to not have any fields in the `Leaf` constructor, so you can represent empty trees and generally define more flexible structures
```suggestion
data Tree a = Leaf
            | Node (Tree a) a (Tree a) deriving (Show, Eq)
```

chshersh (approved):
:tada: Fantastic work!

vrom911:
@pastoreerrante I am going to close this PR, but feel free to reach out to us if you have any questions 😊 Don't be a stranger 🤗

PR #361, opened 2020-11-22
Solved. Up for review

lautaropaske:
### Solutions for Chapter 1

Nice sunday activity! Can't wait for Chapter 2 (in fact, I won't 👨🏻‍💻)

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -490,7 +490,7 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 lastDigit :: Int -> Int
-lastDigit n = mod n 10
+lastDigit n = mod (abs n) 10
 
 
 {- |
@@ -559,9 +559,12 @@ Casual reminder about adding top-level type signatures for all functions :)
 
 mid :: Int -> Int -> Int -> Int
 mid x y z
-  | y < x && x < z = x
-  | x < y && y < z = y
-  | otherwise = z
+  | y <= x && x <= z = x
+  | y >= x && x >= z = x
+  | x <= y && y <= z = y
+  | x >= y && y >= z = y
+  | x <= z && z <= y = z
+  | x >= z && z >= y = z
 
 {- |
 =⚔️= Task 8
@@ -645,9 +648,9 @@ sumLast2 n = sum (lastTwo n)
     lastTwo :: Int -> [Int]
     lastTwo n =
       let
-        snd = mod n 10
-        fst = div (mod n 100 - snd) 10
-      in [fst, snd]
+        second = mod (abs n) 10
+        first  = div (mod (abs n) 100 - second) 10
+      in [first, second]
 
 {- |
 =💣= Task 10*
@@ -668,8 +671,9 @@ aren't ready for this boss yet!
 -}
 firstDigit :: Int -> Int
 firstDigit n
-  | n < 10 = n
-  | otherwise = firstDigit (div n 10)
+  | absN < 10 = absN
+  | otherwise = firstDigit (div (absN) 10)
+  where absN = abs n
 
 
 {-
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -568,8 +575,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
-
+isVowel :: Char -> Bool
+isVowel c = elem c ['a', 'e', 'i', 'o', 'u']
```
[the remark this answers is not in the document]
lautaropaske: Why is this the recommended approach?

src/Chapter1.hs
```diff
@@ -554,7 +557,14 @@ value after "=" where the condition is true.
...
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | y <= x && x <= z = x
+  | y >= x && x >= z = x
+  | x <= y && y <= z = y
+  | x >= y && y >= z = y
+  | x <= z && z <= y = z
+  | x >= z && z >= y = z
```
chshersh, line 567: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

However, It looks like you can move the last two cases under a single otherwise and return z in here instead.

src/Chapter1.hs
```diff
@@ -568,8 +578,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
-
+isVowel :: Char -> Bool
+isVowel c = c `elem` ['a', 'e', 'i', 'o', 'u']
```
chshersh, line 582: Absolutely correct! 👍🏼 

src/Chapter1.hs
```diff
@@ -632,8 +642,15 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
-
+sumLast2 :: Int -> Int
+sumLast2 n = sum (lastTwo n)
+  where
+    lastTwo :: Int -> [Int]
+    lastTwo n =
+      let
+        second = mod (abs n) 10
+        first  = div (mod (abs n) 100 - second) 10
+      in [first, second]
```
chshersh, line 653: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
```

src/Chapter1.hs
```diff
@@ -652,8 +669,11 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+  | absN < 10 = absN
+  | otherwise = firstDigit (div (absN) 10)
+  where absN = abs n
```
chshersh, line 676: Great idea to move `absN` into a separate `where` variable 🔝

chshersh (approved):
Great work on solving our challenges for Chapter One!
And you nailed the Advanced Task 🔨 

src/Chapter1.hs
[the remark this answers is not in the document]
chshersh: @lautaropaske this is a just default HLint suggestion. I guess

```haskell
c `elem` someList
```

reads more naturally like _c is in some list_. But feel free to ignore this suggestion and use your own judgement 👍🏻

chshersh:
@lautaropaske Great work! I'm going to close this PR as we can't merge it to our repo. But feel free to open PRs with solutions to other chapters 🙂

PR #362, opened 2020-11-23
Chapter2

lautaropaske:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -39,6 +39,7 @@ Now, if you are ready, bring it on!
 -}
 
 module Chapter2 where
+import Data.List (take, drop)
 
 {-
 =🛡= Imports
@@ -136,43 +137,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create a list from N same elements:
 >>> :t replicate
-
+take :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+lines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -39,6 +39,7 @@ Now, if you are ready, bring it on!
 -}
 
 module Chapter2 where
+import Data.List (take, drop)
```
vrom911, line 42: These functions come with `Prelude` module, so no need to import them manually from `Data.List` 🙂

src/Chapter2.hs
```diff
@@ -621,7 +626,14 @@ Write a function that takes elements of a list only in even positions.
...
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven = go 0 []
+  where
+    go :: Int -> [a] -> [a] -> [a]
+    go _ acc [] = acc
+    go i acc (x:xs)
+      | (mod i 2) == 0 = go (i+1) (acc ++ [x]) xs
```
vrom911, line 635: Nice idea on the helper functions :+1:

src/Chapter2.hs
```diff
@@ -728,7 +740,8 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l =
+  concat $ map (\n -> replicate n n) l
```
vrom911, line 744: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -781,13 +794,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: (Ord a) => a -> [a] -> [a]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: Num a => [a] -> [a] -> [a]
+pairMul = zipWith (*)
```
vrom911, line 805: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -842,7 +857,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate n xs
+  | n < 0     = []
+  | otherwise = take (length xs) $ drop n (cycle xs)
```
vrom911, line 863: That's a correct solution! Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. There is a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -858,8 +876,11 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
-
+rewind :: [a] -> [a]
+rewind = go []
+  where
+    go acc []     = acc
+    go acc (x:xs) = (rewind xs) ++ (x:acc)
```
vrom911, line 883: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Great Chapter! Very interesting and thoughtful solutions 👏🏼

vrom911:
@lautaropaske I'm going to close this PR, as we can't merge it to our repo 🙂 But feel free to ask any questions in the [course discussions](https://github.com/kowainik/learn4haskell/discussions), don't be a stranger!

And Merry Christmas 🎄

PR #364, opened 2020-11-25
Complete Chapter 2

luigidisotto:
### Solutions for Chapter 2

cc @vrom911 @chshersh

I complete Chapter 2 (with some delay due to the lack of free time), and passed the tests.

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -621,7 +644,16 @@ Write a function that takes elements of a list only on even positions.
...
+
+takeEven :: [a] -> [a]
+takeEven xs = go 0 xs
+  where
+    go :: Int -> [a] -> [a]
+    go _ [] = []
+    go i (x:xs) = if (isEven i) then x : (go (i + 1) xs) else (go (i + 1) xs)
+                  where
+                    isEven :: Int -> Bool
+                    isEven n = (mod n 2) == 0
```
vrom911, line 656: there is a standard function – `even` that does exactly this 🙂

src/Chapter2.hs
```diff
@@ -621,7 +644,16 @@ Write a function that takes elements of a list only on even positions.
...
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+
+takeEven :: [a] -> [a]
+takeEven xs = go 0 xs
+  where
+    go :: Int -> [a] -> [a]
+    go _ [] = []
+    go i (x:xs) = if (isEven i) then x : (go (i + 1) xs) else (go (i + 1) xs)
```
vrom911, line 653: Nice idea on the helper functions :+1:

A few additional thoughts on your idea: as you have only two states of the operation (odd or even), instead of accumulators of the current position number, you can just use `Bool`. And on each next recursion call, you just give it `not acc`. This will remove the necessity to check `isEven` each time, which would save a couple of functions call on each interaction.

src/Chapter2.hs
```diff
@@ -728,7 +760,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate xs = concat (map (\x -> replicate x x) xs)
```
vrom911, line 763: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -741,8 +773,9 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
 
+contains :: Int -> [[Int]] -> [[Int]]
+contains n xs = filter (\x -> elem n x) xs
```
vrom911, line 778: Now, after you've mastered eta-reduction, you can apply such a technique to this function and the lambda function as well  🙂 

src/Chapter2.hs
```diff
@@ -781,13 +814,14 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
+divideTenBy = div 10
 
--- TODO: type ;)
+listElementsLessThan :: Int -> [Int] -> [Int]
 listElementsLessThan x l = filter (< x) l
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 824: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -858,7 +896,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = (rewind xs) ++ [x]
```
vrom911, line 901: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Great Chapter! 👍🏼 Nice and thoughtful solutions 👏🏼

vrom911:
@luigidisotto I'm going to close this PR, as we can't merge it to our repo 🙂 But feel free to ask any questions in the [course discussions](https://github.com/kowainik/learn4haskell/discussions), don't be a stranger!

And Merry Christmas 🎄

PR #365, opened 2020-11-29
solutions for chapter 1

Jeffrey04:
### Solutions for Chapter 1

cc @vrom911 @chshersh

Thanks for the tutorial, this looks like an interesting approach to teach a language (: :+1:

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -5,7 +5,6 @@ Happy to see you here, on the way to the wonderful Functional Programming land
 with Haskell! Fight the fierce Monad Dragon and save the globe from despicable
 runtime exceptions!
 
-
 We appreciate your curiosity and will try to provide you with all the necessary
 equipment for your training before the battle in the real FP world. Learning
 Functional Programming can be challenging. But we designed this training to be
@@ -43,7 +42,6 @@ concepts on the way. In this chapter, you are going to learn:
    variables
  ✧ How to write your function from scratch
  ✧ Some standard Haskell functions
-
 
 We are leaving a number of tasks on our path. Your goal is to solve them all and
 make the test for Chapter One green.
@@ -115,7 +113,6 @@ signatures__ and provide types in different situations where you don't
 immediately see what types will be inferred.
 -}
 
-
  {-
 Haskell is a __compiled__ language. In the illustration below, you can see the
 overall picture of the process from your code to the binary of the written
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -429,9 +430,9 @@ task is to specify the type of this function.
 49
 -}
 
+squareSum :: Num a => a -> a -> a
```
vrom911, line 433: We explain `Num` in the following chapters, so you are one step ahead 😸 👏🏼 

src/Chapter1.hs
```diff
@@ -520,8 +521,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
-
+closestToZero x y = if abs x == min (abs $ negate x) (abs $ negate y) then x else y
```
vrom911, line 524: For absolute value, it doesn't matter what sign the number has, so `negate` is redundant in this case 🙂 

Very interesting solution! 💡 
Though it could be a bit simplified. You see that all you need is to compare the absolute value of both `x` and `y`, and based on that return the result as you do in the `then` and `else` clauses 👍🏼

src/Chapter1.hs
```diff
@@ -568,8 +572,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
-
+isVowel :: Char -> Bool
+isVowel = (`elem` "aeiou")
```
vrom911, line 576: Nice one! 👍🏼 

vrom911 (approved):
Amazing job! Congrats with finishing Chapter One 🎉 
Very thoughtful solutions, and great advanced task solution 👌🏼 🧑‍🍳

src/Chapter1.hs
Jeffrey04: yea, briefly read the learn haskell for great good book before starting this
also having the language server set up properly makes it easier :P

src/Chapter1.hs
Jeffrey04: originally it was `(abs (0 - x))`, and it became this after applying the fix hinted by the linter
didn't really give it another thought after seeing it passing the unit test :P

src/Chapter1.hs
Jeffrey04: that makes sense, thanks for the information

Jeffrey04:
just out of curiosity, is there anything i still miss in this?

vrom911:
Oh, sorry! It looks awesome 🌟 I think this Chapter is completely finished, great job 💪🏼

PR #366, opened 2020-11-29
chapter 2 solutions

Jeffrey04:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -756,7 +756,7 @@ the list with only those lists that contain a passed element.
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
 contains :: Eq a => a -> [[a]] -> [[a]]
-contains x = filter (\l-> x `elem` l)
+contains x = filter (elem x)
 
 {- |
 =🛡= Eta-reduction
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -514,7 +514,7 @@ True
 False
 -}
 isThird42 :: (Eq a, Num a) => [a] -> Bool
-isThird42 (_:_:subject:_) = subject == 42
+isThird42 (_:_:42:_) = True
 isThird42 _ = False
 
 {- |
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -861,7 +861,7 @@ list.
 rotate :: Int -> [a] -> [a]
 rotate n xs
   | n < 0 = []
-  | otherwise = take (length xs) $ drop n (cycle xs)
+  | otherwise = let count = length xs in take count $ drop (n `mod` count) (cycle xs)
 
 {- |
 =💣= Task 12*
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -878,8 +878,14 @@ and reverses it.
   cheating!
 -}
 rewind :: [a] -> [a]
+-- alternatively
+-- rewind = foldl (flip (:)) []
 rewind [] = []
-rewind (x:xs) = rewind xs ++ [x]
+rewind xs = go [] xs
+  where
+    go :: [a] -> [a] -> [a]
+    go acc [] = acc
+    go acc (y:ys) = go (y : acc) ys
 
 {-
 You did it! Now it is time to open pull request with your changes
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -348,7 +348,7 @@ ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
 subList n m list
-  | n < 0 = []
+  | n < 0 || m < n = []
   | otherwise = take (m - n + 1) $ drop n list
 
 {- |
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -1,3 +1,4 @@
+{- ORMOLU_DISABLE -}
 {- 👋 Welcome to Chapter Two, Traveller!
 
 If you haven't finished Chapter One yet, we encourage you to check it
@@ -75,7 +76,6 @@ When working with lists, the most practical module will be "Data.List":
    * http://hackage.haskell.org/package/base-4.14.0.0/docs/Data-List.html
 -}
 
-
 {- |
 =🛡= Lists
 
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -1,3 +1,4 @@
+{- ORMOLU_DISABLE -}
```
chshersh, line 1: 🙏🏻

src/Chapter2.hs
```diff
@@ -336,7 +347,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList n m list
+  | n < 0 = []
```
chshersh, line 351: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding a separate case to your check :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -501,8 +513,9 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
-
+isThird42 :: (Eq a, Num a) => [a] -> Bool
+isThird42 (_:_:subject:_) = subject == 42
```
chshersh, line 517: You don't need to compare `subject` with 42 separately, you can pattern match on `42` directly in the pattern.
```suggestion
isThird42 (_:_:42:_) = True
```

src/Chapter2.hs
```diff
@@ -621,7 +633,10 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven [x] = [x]
+takeEven (x:_:xs) = x : takeEven xs
```
chshersh, line 639: Great pattern matching skills 👏🏻

src/Chapter2.hs
```diff
@@ -728,7 +742,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate = concatMap (\x->replicate x x)
```
chshersh, line 745: Great usage of standard functions! 👍🏻

src/Chapter2.hs
```diff
@@ -741,8 +755,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: Eq a => a -> [[a]] -> [[a]]
+contains x = filter (\l-> x `elem` l)
```
chshersh, line 759: If you use the prefix notation for the `elem` function, you can notice an opportunity for eta-reduction here 🙂 
```suggestion
contains x = filter (elem x)
```

src/Chapter2.hs
```diff
@@ -741,8 +755,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: Eq a => a -> [[a]] -> [[a]]
```
chshersh, line 758: Btw, great work on writing polymorphic type signatures! 👏🏻

src/Chapter2.hs
```diff
@@ -842,7 +858,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate n xs
+  | n < 0 = []
+  | otherwise = take (length xs) $ drop n (cycle xs)
```
chshersh, line 864: That's a correct solution! Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -858,8 +877,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
-
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```

chshersh (approved):
Wonderful job in Chapter Two! 💎 
And great work on the Advanced Task 👍🏻 

I left some comments with suggestions and some ideas 🙂

src/Chapter2.hs
Jeffrey04: yea, otherwise the whole pull request will be full of whitespace changes LOL

src/Chapter2.hs
Jeffrey04: O.O totally forgot about that, thanks

src/Chapter2.hs
Jeffrey04: thanks (:

src/Chapter2.hs
Jeffrey04: thanks to the linter (it was `concat $ map .....`) :P didn't know flatmap is called a concatMap in haskell

src/Chapter2.hs
Jeffrey04: thanks for the tip (:

src/Chapter2.hs
Jeffrey04: (:

src/Chapter2.hs
Jeffrey04: oooo ok

src/Chapter2.hs
Jeffrey04: am aware of that, just thought it would be easier to understand to write it in this way lol
will attempt a rewrite

btw, am kinda stuck with Chapter3 as it is a lot more free-form
can i submit incomplete pull request?

Jeffrey04: I suppose this is something I am more familiar with but prolly not something you are looking for `rewind xs = foldl (flip (:)) [] xs
`

src/Chapter2.hs
Jeffrey04: not too sure if I get you correctly, you are looking for another case where `m < n` as an additional case to return empty list?

chshersh: Yeah, I meant exactly that 🙂

chshersh:
@Jeffrey04 Great work on chapter two! Improvements look awesome 💯 
Nice and idiomatic implementation of `rewind` 👍🏻 
I'm going to close this PR, as it looks great 🔝

Jeffrey04:
> @Jeffrey04 Great work on chapter two! Improvements look awesome 100
> Nice and idiomatic implementation of `rewind` 👍🏻
> I'm going to close this PR, as it looks great top

thanks (:

PR #367, opened 2020-11-29
Chapter2 solution

srahmi:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -411,7 +411,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Int -> Int -> Int
 squareSum x y = (x + y) * (x + y)
 
 
@@ -432,7 +432,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next x = x + 1
 
 {- |
 After you've implemented the function (or even during the implementation), you
@@ -473,7 +473,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = n `mod` 10
 
 
 {- |
@@ -503,7 +504,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if abs x < abs y then x else y
 
 
 {- |
@@ -536,9 +537,14 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
-
+mid :: Int -> Int -> Int -> Int
+mid x y z 
+    | x < y && y < z && z > x = y
+    | x < y && y > z && z > x = z
+    | x > y && y < z && z > x = x
+    | x < y && y > z && z < x = x
+    | x < y && y < z && z < x = z
+    | x > y && y > z && z < x = y
 {- |
 =⚔️= Task 8
 
@@ -551,7 +557,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c 
+    | c `elem` "aeyuioAEYUIO" = True
+    | otherwise = False
 
 
 {- |
@@ -615,7 +624,13 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+
+sumLast2 :: Integral a => a -> a
+sumLast2 n = 
+   let 
+      last2 = n `mod` 100
+   in
+      (last2 `div` 10) + (last2 `mod` 10)
 
 
 {- |
@@ -636,8 +651,12 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Integral a => a -> a
+firstDigit n = let 
+                   rest = n `div` 10
+               in
+                 if rest < 10 then rest else firstDigit rest
+    
 
 {-
 You did it! Now it is time to the open pull request with your changes
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -310,7 +310,7 @@ of a list between two given positions inclusive (starting from zero).
 
 Remember that each function returns a new list.
 
->>> subList 3 5 [1 .. 10]
+>>> subList 3 5 [1 .. 10] 
 [4,5,6]
 >>> subList 3 0 [True, False, False, True, False]
 []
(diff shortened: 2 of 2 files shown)


src/Chapter1.hs
```diff
@@ -473,7 +473,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = n `mod` 10
```
vrom911, line 477: Your implementation is almost correct :ok: 

src/Chapter1.hs
```diff
@@ -536,9 +537,14 @@ value after "=" where the condition is true.
...
-mid x y z = error "mid: not implemented!"
-
+mid :: Int -> Int -> Int -> Int
+mid x y z 
+    | x < y && y < z && z > x = y
+    | x < y && y > z && z > x = z
+    | x > y && y < z && z > x = x
+    | x < y && y > z && z < x = x
+    | x < y && y < z && z < x = z
+    | x > y && y > z && z < x = y
```
vrom911, line 547: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

As you covered all cases by these checks, you actually know that if the input didn't match any previous cases, it means that the result is definitely `y`. You can skip checking, and write `otherwise = y` instead of this last line 👍🏼 
Also you can remove the first case(which returns `y`) as it also be covered by such `otherwise` case 🙂

src/Chapter1.hs
```diff
@@ -551,7 +557,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c 
+    | c `elem` "aeyuioAEYUIO" = True
+    | otherwise = False
```
vrom911, line 563: Nice one! 👍🏼 

```haskell
isVowel c = c `elem` "aeyuioAEYUIO"
```

src/Chapter1.hs
```diff
@@ -615,7 +624,13 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+
+sumLast2 :: Integral a => a -> a
+sumLast2 n = 
+   let 
+      last2 = n `mod` 100
+   in
+      (last2 `div` 10) + (last2 `mod` 10)
```
vrom911, line 633: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
last2 `mod` 10
last2 `div` 10
```

```haskell
(a, b) = last2 `divMod` 10
```

Also, you need to use `abs` in here to cover negative cases 👌🏼

src/Chapter1.hs
```diff
@@ -636,8 +651,12 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Integral a => a -> a
+firstDigit n = let 
+                   rest = n `div` 10
+               in
+                 if rest < 10 then rest else firstDigit rest
```
vrom911, line 658: I see, that for a negative number like -19, this implementation will return the negative number itself instead of the last digit because the first check n < 10 will always succeed for negative numbers.

src/Chapter2.hs
```diff
@@ -327,7 +327,10 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList p1 p2 xs 
+   | p2 < p1 = []
+   | otherwise = let rest = drop p1 xs 
+                       in take (p2 - p1 + 1) rest
```
vrom911, line 333: Great one!
But it could work not as expected on negative input. You can check that both numbers are not negative 🙂

src/Chapter2.hs
```diff
@@ -340,8 +343,9 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 -- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
-
+firstHalf :: [a] -> [a]
+firstHalf xs = let l = length xs `div` 2
```
vrom911, line 347: Wonderful usage of the `let-in` construction! 👏🏼

src/Chapter2.hs
```diff
@@ -491,7 +495,9 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird42 ::[Int] -> Bool
+isThird42 (_:_:x:_) = x == 42
```
vrom911, line 499: You don't need to compare `z` with 42 separately, you can pattern match on `42` directly in the pattern :)

src/Chapter2.hs
```diff
@@ -611,7 +618,11 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+takeEven [] = []
+takeEven [x] = [x]
+takeEven [x,_] = [x]
+takeEven (x:_:ys)  = x : takeEven ys
```
vrom911, line 625: Fantastic pattern-matching skills 🎉 Bravo!

src/Chapter2.hs
```diff
@@ -718,7 +729,8 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate (x:xs) = replicate x x ++ smartReplicate xs
```
vrom911, line 733: This is right! 💯 

```haskell
[3,1,2,4,0,1]
```

src/Chapter2.hs
```diff
@@ -771,13 +784,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Ord a => a -> [a] -> [a]
+listElementsLessThan x  = filter (< x) 
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: Num c => [c] -> [c] -> [c]
+pairMul  = zipWith (*)
```
vrom911, line 795: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -848,8 +867,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
-
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
vrom911, line 872: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Awesome work on the first two Chapters 🎉 
Very interesting and thoughtful solutions 💪🏼 

I have noticed that you didn't fill in answers to GHCi prompt, I wonder if you need help with that. Let me know if there are any questions! We are here to help 🤗

srahmi:
> Awesome work on the first two Chapters 🎉
> Very interesting and thoughtful solutions 💪🏼
> 
> I have noticed that you didn't fill in answers to GHCi prompt, I wonder if you need help with that. Let me know if there are any questions! We are here to help 🤗

Thank you so much for your help.
About GHCi prompt I didn't notice them because I was so excited to start coding 😄

chshersh:
@srahmi I'm going to close this PR, as we can't merge it to our repo 🙂

PR #369, opened 2020-12-01
solutions for chapter 3

Jeffrey04:
### Solutions for Chapter 3

Not too happy with the code I wrote for the last task but ran out of idea on how to fix it (each commit shows slightly different attempt to the problem)
(prolly didn't do well in other tasks too)

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -1215,13 +1215,13 @@ instance Fighter FighterMonster where
 
 instance FMonster FighterMonster
 
-engageBattle :: (Fighter a, Fighter b) => a -> b -> Either a b
+engageBattle :: (Fighter a, Fighter b) => a -> b -> (Maybe a, Maybe b)
 engageBattle attacker defender =
   go 0 (0, 0) (Just attacker) (Just defender)
   where
-    go :: (Fighter a, Fighter b) => Int -> (Int, Int) -> Maybe a -> Maybe b -> Either a b
-    go _ _ Nothing (Just dfder) = Right dfder
-    go _ _ (Just atker) Nothing = Left atker
+    go :: (Fighter a, Fighter b) => Int -> (Int, Int) -> Maybe a -> Maybe b -> (Maybe a, Maybe b)
+    go _ _ Nothing (Just dfder) = (Nothing, Just dfder)
+    go _ _ (Just atker) Nothing = (Just atker, Nothing)
     go turn (attackerIdx, defenderIdx) (Just atker) (Just dfder)
       | even turn =
         let (atkerNew, dfderNew) = actions atker !! attackerIdx $ dfder
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -1129,9 +1129,11 @@ newtype StatAttack = StatAttack Int deriving (Show)
 
 newtype StatDefense = StatDefense Int deriving (Show)
 
+type ActionOutcome a b = (Maybe a, Maybe b)
+
 class Fighter a where
-  actions :: (Fighter b) => a -> [b -> (Maybe a, Maybe b)]
-  attack :: (Fighter b) => a -> b -> (Maybe a, Maybe b)
+  actions :: (Fighter b) => a -> [b -> ActionOutcome a b]
+  attack :: (Fighter b) => a -> b -> ActionOutcome a b
   health :: a -> StatHealth
 
   -- takes damage, and return an updated fighter with new HP
@@ -1139,15 +1141,15 @@ class Fighter a where
 
 class (Fighter a) => FKnight a where
   -- assume no resource management
-  castSpell :: (Fighter b) => a -> b -> (Maybe a, Maybe b)
+  castSpell :: (Fighter b) => a -> b -> ActionOutcome a b
 
   defense :: a -> StatDefense
 
   -- assuming bottomless potion
-  drinkPotion :: (Fighter b) => a -> b -> (Maybe a, Maybe b)
+  drinkPotion :: (Fighter b) => a -> b -> ActionOutcome a b
 
 class (Fighter a) => FMonster a where
-  runAway :: (Fighter b) => a -> b -> (Maybe a, Maybe b)
+  runAway :: (Fighter b) => a -> b -> ActionOutcome a b
   runAway _ target = (Nothing, Just target)
 
 data FighterKnight = MkFighterKnight
@@ -1176,8 +1178,8 @@ instance Fighter FighterKnight where
     | newHealth > 0 = Just x {fighterKnightHealth = StatHealth newHealth}
     | otherwise = Nothing
     where
-      (StatHealth hp) = health x
-      (StatDefense def) = defense x
+      StatHealth hp = health x
+      StatDefense def = defense x
       newHealth = if y > def then hp + def - y else hp
 
 instance FKnight FighterKnight where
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -373,6 +378,23 @@ after the fight. The battle has the following possible outcomes:
...
 -}
+data LivingThing = MkLivingThing {
+  entityHealth :: Int,
+  entityAttack :: Int,
+  entityGold :: Int
+} deriving (Show)
+type Knight = LivingThing
+type Monster = LivingThing
+
+fight :: Monster -> Knight -> Int
```
vrom911, line 389: Nice usage of type aliases!

```haskell
> fight knight monster
> fight monster knight
```

```haskell
```

src/Chapter3.hs
```diff
@@ -480,6 +502,41 @@ After defining the city, implement the following functions:
...
+{- ORMOLU_ENABLE -}
+data Castle = MkCastle String
+
+data Wall = BrickWall
+
+data Facility = Church | Library
+
+data People = Citizen
+
+data House = MkHouse [People]
```
vrom911, line 514: This is the correct data type, that can represent what we have in requirements 👍🏼 

A few thoughts:
It is more comfortable to work with numbers instead of the length of the list of void elements. In fact, the `People` type doesn't contain any information at all. So maybe you could think of how to represent only possible situations with the limits that the house has on people. In your case, there are no limits, which could lead to illegal states.
With this representation, this is possible. However, we actually can restrict this on type-level by providing more representative types 🙂 This is actually the greatest power of Haskell and type system 😍 

src/Chapter3.hs
```diff
@@ -480,6 +502,41 @@ After defining the city, implement the following functions:
...
+
+data City = MkCityNoCastle Facility [House] | MkCityWithCastle Castle [Wall] Facility [House]
```
vrom911, line 516: The same note about the list of empty `Wall`s. I see that you again need a number 🙂

src/Chapter3.hs
```diff
@@ -560,22 +617,29 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defense $ armor * dexterity
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage damage) (Defense defense) (Health health) = Health $ health + defense - damage
```
vrom911, line 642: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -907,9 +977,39 @@ Implement instances of "Append" for the following types:
...
+  append x y = combine Empty x y
+    where
+      -- rebuild cons, unlike implmenting reverse in previous chapter
+      teardown :: List a -> List a -> List a
+      teardown acc Empty = acc
+      teardown acc (Cons z zs) = Cons z (teardown acc zs)
+      -- reconstruct the list, one after another
+      combine :: List a -> List a -> List a -> List a
+      combine acc Empty Empty = acc
+      combine acc xs ys = combine (teardown (teardown acc xs) ys) Empty Empty
```
vrom911, line 1004: I don't quite get the logic here. You reverse lists twice, but this gives the normal order.
Append doesn't change the order of elements in List, so I think you don't need `teardown` for implementation of this function.

src/Chapter3.hs
```diff
@@ -907,9 +977,39 @@ Implement instances of "Append" for the following types:
...
+
+-- note2self: constraint means type a implements Append already, and now we are implementing for (Maybe a)
+instance (Append a) => Append (Maybe a) where
+  append _ Nothing = Nothing
+  append Nothing _ = Nothing
```
vrom911, line 1009: While this is valid implementation and nothing in the task description says differently, in reality, it would be more beneficial is this function returned `Just` when at least one argument is `Just`.

src/Chapter3.hs
```diff
@@ -971,6 +1071,31 @@ implement the following functions:
...
+{- ORMOLU_ENABLE -}
+data Weekday = Monday | Tuesday | Wednesday | Thursday | Friday | Saturday | Sunday deriving (Show, Eq, Ord, Enum)
+
+isWeekend :: Weekday -> Bool
+isWeekend x
+  | x == Saturday || x == Sunday = True
+  | otherwise = False
+
+nextDay :: Weekday -> Weekday
+nextDay Sunday = Monday
```
vrom911, line 1083: That is right!
But you could use the instance of `Bounded` here:

```suggestion
    | day == maxBound = minBound
```

src/Chapter3.hs
```diff
@@ -971,6 +1071,31 @@ implement the following functions:
...
+daysToParty :: Weekday -> Int
+daysToParty day =
+  length
+    ( takeWhile
+        (/= Friday)
+        ( drop
+            (fromEnum day)
+            (cycle (enumFrom Monday))
+        )
+    )
```
vrom911, line 1095: Interesting.. But it is far from an optimal solution as you can guess 😄 
 
You could just use the `Enum` instance here and the fact that there is 7 days in a weak. You can find `mod ... 7` useful 🙂

src/Chapter3.hs
```diff
@@ -1006,6 +1131,107 @@ Implement data types and typeclasses, describing such a battle between two
...
 
+{- ORMOLU_ENABLE -}
+newtype StatHealth = StatHealth Int deriving (Show)
+
+newtype StatAttack = StatAttack Int deriving (Show)
+
+newtype StatDefense = StatDefense Int deriving (Show)
+
+class Fighter a where
+  actions :: (Fighter b) => a -> [b -> (Maybe a, Maybe b)]
```
vrom911, line 1142: You use `(Maybe a, Maybe b)` a lot. You can move it into a separate data type and also write a few helper functions for that 🙂

src/Chapter3.hs
```diff
@@ -1006,6 +1131,107 @@ Implement data types and typeclasses, describing such a battle between two
...
+class (Fighter a) => FKnight a where
+  -- assume no resource management
+  castSpell :: (Fighter b) => a -> b -> (Maybe a, Maybe b)
+
+  defense :: a -> StatDefense
+
+  -- assuming bottomless potion
+  drinkPotion :: (Fighter b) => a -> b -> (Maybe a, Maybe b)
+
+class (Fighter a) => FMonster a where
```
vrom911, line 1158: Interesting classes 👍🏼 Nice idea!

src/Chapter3.hs
```diff
@@ -1006,6 +1131,107 @@ Implement data types and typeclasses, describing such a battle between two
...
+
+  attack attacker target = (Just attacker, damage target (fighterKnightAttack attacker))
+
+  health = fighterKnightHealth
+  damage x (StatAttack y)
+    | newHealth > 0 = Just x {fighterKnightHealth = StatHealth newHealth}
+    | otherwise = Nothing
+    where
+      (StatHealth hp) = health x
+      (StatDefense def) = defense x
```
vrom911, line 1189: I think you don't need parentheses in such expressions 🙂

vrom911 (approved):
Wow, great Chapter! Lots of interesting types you constructed here 👷🏼 👏🏼

vrom911:
@Jeffrey04 it looks neat now! I am going to close this PR 🙂 Very well done 👌🏼

PR #370, opened 2020-12-01
solutions for chapter 4

Jeffrey04:
### Solutions for Chapter 4

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -506,12 +506,9 @@ Implement the 'Applicative' instance for our 'List' type.
 {- ORMOLU_ENABLE -}
 -- copy pasta from chap3 instead of importing
 append :: List a -> List a -> List a
-append x y = merge y x
-  where
-    merge :: List a -> List a -> List a
-    merge Empty zs = zs
-    merge acc Empty = acc
-    merge acc (Cons z zs) = Cons z (merge acc zs)
+append Empty xs = xs
+append xs Empty = xs
+append (Cons x xs) ys = Cons x (append xs ys)
 
 instance Applicative List where
   pure :: a -> List a
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -509,6 +509,7 @@ append :: List a -> List a -> List a
 append x y = merge y x
   where
     merge :: List a -> List a -> List a
+    merge Empty zs = zs
     merge acc Empty = acc
     merge acc (Cons z zs) = Cons z (merge acc zs)
 
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -1,3 +1,4 @@
+{- ORMOLU_DISABLE -}
 {- 👋 Welcome to Chapter Four / Four of our journey, Staunch Wanderer!
 
 Our adventure together has almost come to an end. It has been a
@@ -114,22 +115,30 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
+Char :: *
 
 >>> :k Bool
+Bool :: *
 
 >>> :k [Int]
+[Int] :: *
 
 >>> :k []
+[] :: * -> *
 
 >>> :k (->)
+(->) :: * -> * -> *
 
 >>> :k Either
+Either :: * -> * -> *
 
 >>> data Trinity a b c = MkTrinity a b c
 >>> :k Trinity
+Trinity :: * -> * -> * -> *
 
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
+IntBox :: (* -> *) -> *
 
 -}
 
@@ -259,6 +268,7 @@ name.
 
 > QUESTION: Can you understand why the following implementation of the
   Functor instance for Maybe doesn't compile?
+  ANSWER: there is no transformation from (a) to (f a) hence violates the signature
 
 @
 instance Functor Maybe where
@@ -293,7 +303,8 @@ values and apply them to the type level?
 -}
 instance Functor (Secret e) where
     fmap :: (a -> b) -> Secret e a -> Secret e b
-    fmap = error "fmap for Box: not implemented!"
+    fmap _ (Trap e) = Trap e
+    fmap f (Reward a) = Reward $ f a
 
 {- |
 =⚔️= Task 3
@@ -306,6 +317,11 @@ typeclasses for standard data types.
 data List a
     = Empty
     | Cons a (List a)
+
+instance Functor List where
+  fmap :: (a -> b) -> List a -> List b
+  fmap _ Empty = Empty
+  fmap f (Cons x xs) = Cons (f x) $ fmap f xs
 
 {- |
 =🛡= Applicative
@@ -472,10 +488,11 @@ Implement the Applicative instance for our 'Secret' data type from before.
 -}
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    (<*>) (Trap e) _ = Trap e
+    (<*>) (Reward f) m1 = fmap f m1
 
 {- |
 =⚔️= Task 5
@@ -488,7 +505,22 @@ Implement the 'Applicative' instance for our 'List' type.
   may also need to implement a few useful helper functions for our List
   type.
 -}
+{- ORMOLU_ENABLE -}
+instance Applicative List where
+  pure :: a -> List a
+  pure x = Cons x Empty
 
+  (<*>) Empty _ = Empty
+  (<*>) fs ms =
+    goM Empty ms
+    where
+      goF acc Empty _ = acc
+      goF acc (Cons f fTail) m = goF (Cons (f m) acc) fTail m
+
+      goM acc Empty = acc
+      goM acc (Cons m mTail) = goF (goM acc mTail) fs m
+
+{- ORMOLU_DISABLE -}
 
 {- |
 =🛡= Monad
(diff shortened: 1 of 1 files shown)


src/Chapter4.hs
```diff
@@ -629,7 +662,7 @@ Can you implement a monad version of AND, polymorphic over any monad?
 🕯 HINT: Use "(>>=)", "pure" and anonymous function
 -}
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM a b = (a >>= (\x -> pure (x &&))) <*> b
```
Jeffrey04, line 665: This fails when `a = Just False` and `b = Nothing`
shouldn't all computation involving `Nothing` returns `Nothing`?

Jeffrey04: also tried with
```
andM a b = do
  m<-a
  n<-b
  return (m && n)
```
and simpler
```
fmap (&&) a <*>
```
still failing the case

Jeffrey04: kinda fixed in the latest revision, not sure if the solution is correct tho

src/Chapter4.hs
```diff
@@ -259,6 +267,7 @@ name.
 
 > QUESTION: Can you understand why the following implementation of the
   Functor instance for Maybe doesn't compile?
+  ANSWER: there is no transformation from (a) to (f a) hence violates the signature
```
chshersh, line 270: That's correct 👍🏻 In other words, variable `x` has type `Maybe a`, so you can return it because you need to return a value of type `Maybe b`. The compiler is not smart enough to guess what you mean here, so you need to pattern match on the existing value and return a different value using the `Nothing` constructor.

src/Chapter4.hs
```diff
@@ -472,10 +486,11 @@ Implement the Applicative instance for our 'Secret' data type from before.
 -}
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
```
chshersh, line 489: Great eta-reduced implementation! 💎

src/Chapter4.hs
```diff
@@ -472,10 +486,11 @@ Implement the Applicative instance for our 'Secret' data type from before.
...
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    (<*>) (Trap e) _ = Trap e
+    (<*>) (Reward f) m1 = fmap f m1
```
chshersh, line 493: Nice work on reusing `Functor` instance for `Secret`! :world_map:

src/Chapter4.hs
```diff
@@ -488,7 +503,22 @@ Implement the 'Applicative' instance for our 'List' type.
...
+
+  (<*>) Empty _ = Empty
+  (<*>) fs ms =
+    goM Empty ms
+    where
+      goF acc Empty _ = acc
+      goF acc (Cons f fTail) m = goF (Cons (f m) acc) fTail m
 
+      goM acc Empty = acc
+      goM acc (Cons m mTail) = goF (goM acc mTail) fs m
```
chshersh, line 519: I would strongly advise adding type signatures to local functions (`goF` and `goM` in this case). For example, after adding a type signature you may notice some similarities with existing functions and probably assign clearer names 🙂 Current implementation is a bit difficult to read because of existing names.

If you have only a single recursive function in `where`, it's okay to call it `go`, as you already have the name for the top-level function, so no need to create an extra name for recursive helper.

src/Chapter4.hs
```diff
@@ -629,7 +659,7 @@ Can you implement a monad version of AND, polymorphic over any monad?
 🕯 HINT: Use "(>>=)", "pure" and anonymous function
 -}
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM a b = a >>= (\x -> if x then b >>= (\y -> pure (x && y)) else pure x)
```
chshersh, line 662: That's correct implementation! Though, it could be simplified a bit, knowing the fact that `True && x` is always `x`.

```suggestion
andM a b = a >>= (\x -> if x then b else pure False)
```

chshersh (approved):
:tada: Fantastic work!

src/Chapter4.hs
chshersh: If you don't use the `>>=` operator (like in the last implementation), then the `Monad` constraint is too strong, and it is enough to specify only `Applicative`. However, with `Applicative` you are not able to implement a lazier version of `andM`.

The ordinary `&&` if lazy on the second argument. Meaning that if it meets `False` as the first argument, there is no need to calculate another arg. In `andM` we would like to keep the same behaviour, so the

```
Just False `andM` _
```

Jeffrey04:
> tada Fantastic work!
> Thank you so much for taking this course! I hope you had fun man_dancing microphone
> 
> We would really appreciate your [feedback on the course](https://docs.google.com/forms/d/e/1FAIpQLScBVhLxq5CgGnAfIGUE-fCoOUqeGkDY2HXzbT7KV2jjLOsmjQ/viewform)!
> 
> You can also ask any questions and we can discuss any topics you would like to dig into as well slightly_smiling_face Don't be a stranger!

thanks for the feedback, it was overall a great fun experience (it is nice to have a mentor to talk to while going through the materials, especially on types(Chapter 3) and Monad (Chapter 4) where I struggle the most)

src/Chapter4.hs
Jeffrey04: (:

src/Chapter4.hs
Jeffrey04: (:

src/Chapter4.hs
Jeffrey04: AH!!!!! thanks (:

src/Chapter4.hs
Jeffrey04: (:

src/Chapter4.hs
Jeffrey04: I suppose one should do `flatten` another `map` lol, but can't find a way to make it

src/Chapter4.hs
Jeffrey04: got it, thanks for the explanation (:

src/Chapter4.hs
```diff
@@ -508,15 +508,17 @@ instance Applicative List where
...
-      goF acc (Cons f fTail) m = goF (Cons (f m) acc) fTail m
+      loopM :: List b -> List a -> List (a -> b) -> List b
+      loopM acc Empty _ = acc
+      loopM acc (Cons mHead mTail) f = loopF (loopM acc mTail f) f mHead
 
-      goM acc Empty = acc
-      goM acc (Cons m mTail) = goF (goM acc mTail) fs m
+      loopF :: List b -> List (a -> b) -> a -> List b
+      loopF acc Empty _ = acc
+      loopF acc (Cons f fTail) m = Cons (f m) $ loopF acc fTail m
```
chshersh: Now it's a bit cleaner what's going on 🙂 So, my suggestion to simplify the implementation is the following:

1. In the `<*>` operator, handle two cases with `Empty` separately: `_ <*> Empty` and `Empty <*> _`.
2. After that, you will have a third case where the first pattern is `Cons`: `Cons f fs <*> xs`.
3. In that third case, you can apply the `f` function to `xs` and append to the result of `fs <*> xs`. For this, you need to implement a function for appending two lists: `append :: List a -> List a -> List a`.

I believe, this new final implementation will be cleaner and simpler 🙂

Jeffrey04: hmm, let me try again

src/Chapter4.hs
```diff
@@ -504,21 +504,22 @@ Implement the 'Applicative' instance for our 'List' type.
   type.
 -}
 {- ORMOLU_ENABLE -}
+-- copy pasta from chap3 instead of importing
+append :: List a -> List a -> List a
+append x y = merge y x
+  where
+    merge :: List a -> List a -> List a
+    merge acc Empty = acc
```
vrom911, line 512: You can add one more case to make your solution even more efficient:

```haskell
    merge Empty xs = xs
```

In that case, you won't need to walk through each element of the second list when the first one is empty and return the list instantly 🚤

Also, you can inline merge function directly into the append, if you will switch the arguments in the pattern matching 🙂

src/Chapter4.hs
```diff
@@ -504,21 +504,22 @@ Implement the 'Applicative' instance for our 'List' type.
...
-  (<*>) fs ms = loopM Empty ms fs
-    where
-      loopM :: List b -> List a -> List (a -> b) -> List b
-      loopM acc Empty _ = acc
-      loopM acc (Cons mHead mTail) f = loopF (loopM acc mTail f) f mHead
-
-      loopF :: List b -> List (a -> b) -> a -> List b
-      loopF acc Empty _ = acc
-      loopF acc (Cons f fTail) m = Cons (f m) $ loopF acc fTail m
+  (<*>) (Cons f fs) ms = append (f <$> ms) (fs <*> ms)
```
vrom911, line 522: This is a really fantastic solution 👍🏼

Jeffrey04: Thanks

Jeffrey04:
Just an additional question regarding this, is the "Kinds" mentioned in this chapter related to higher kinded type? Do you have more introductory guides/text u think that would be helpful if I want to learn more about it?

src/Chapter4.hs
Jeffrey04: OOOOOOOOOOOooooOOOOOOO, thanks

chshersh:
@Jeffrey04 Yes, kinds are very related to higher-kinded types. Similar to higher-order functions that take other functions as arguments, higher-kinded types take other type constructors as arguments. The `IntBox` type from this chapter is an example of such higher-kinded type.

You can check the "Pure Gold by ImpurePics" section in the first issue of @bindthegap FP magazine, to get an overview of kinds:

* https://bindthegap.news/issues/01nov2020.html

The following blog post is also a good introduction to kinds in Haskell:

* https://diogocastro.com/blog/2018/10/17/haskells-kind-system-a-primer/

Jeffrey04:
thanks for the info (:

vrom911:
Thanks again for taking the course! 🙌🏼 
I am going to close this PR, but feel free to reach out to us if you have any questions, don't be a stranger! 🤗 

And Merry Christmas 🎄 ❄️

Jeffrey04:
thanks for all the help and guidance

PR #372, opened 2020-12-16
* finished Chapter1

zexee:
### Solutions for Chapter 1
cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
@@ -301,43 +301,43 @@ expressions in GHCi
   functions and operators first. Remember this from the previous task? ;)
 
 >>> 1 + 2
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+3
 
 >>> 10 - 15
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+-5
 
 >>> 10 - (-5)  -- negative constants require ()
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+15
 
 >>> (3 + 5) < 10
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> True && False
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+False
 
 >>> 10 < 20 || 20 < 5
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> 2 ^ 10  -- power
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+1024
 
 >>> not False
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> div 20 3  -- integral division
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+6
 
 >>> mod 20 3  -- integral division remainder
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+2
 
 >>> max 4 10
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+10
 
 >>> min 5 (max 1 2)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+2
 
 >>> max (min 1 10) (min 5 7)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+5
 
 Because Haskell is a __statically-typed__ language, you see an error each time
 you try to mix values of different types in situations where you are not
@@ -429,6 +429,7 @@ task is to specify the type of this function.
 49
 -}
 
+squareSum :: Int -> Int -> Int
 squareSum x y = (x + y) * (x + y)
 
 
@@ -449,7 +450,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next x = x + 1
 
 {- |
 After you've implemented the function (or even during the implementation), you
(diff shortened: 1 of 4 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +491,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
```
vrom911, line 495: Nice, Your implementation is almost correct :ok: 

src/Chapter1.hs
```diff
@@ -520,7 +522,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if abs(x) < abs(y) then x else y
```
vrom911, line 525: You can omit parentheses for the arguments 🙂 
Also, in Haskell functions and its arguments are always separated by spaces 👾 

```suggestion
closestToZero x y = if abs x < abs y then x else y
```

src/Chapter1.hs
```diff
@@ -554,7 +556,14 @@ value after "=" where the condition is true.
...
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | x >= y && y >= z = y
+    | z >= y && y >= z = y
+    | y >= x && x >= z = x
+    | z >= x && x >= y = x
+    | x >= z && z >= y = z
+    | y >= z && z >= x = z
```
vrom911, line 566: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

But as you correctly noticed, you have covered all possible combinations.
That meant that when pattern matching would come to these two cases, the answer will always be `z`. So in reality, you don't need to check these two conditions:

```suggestion
    | otherwise = z
```

src/Chapter1.hs
```diff
@@ -632,7 +648,14 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+    let
+      last1 :: Int -> Int
+      last1 x = mod x 10
+      last2 :: Int -> Int
+      last2 x = mod x 100
```
vrom911, line 657: I guess here you need to take `mod` of `div` of the `last1`.

But in this case, you can see that you use both:

```haskell
mod (abs x) 10
div (abs x) 10
```

```haskell
(lastDiv, last1) = divMod (abs n) 10
```

src/Chapter1.hs
```diff
@@ -653,7 +676,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+    | n < 10 = n
```
vrom911, line 681: Cool recursion implementation :clap: 
I see, that for a negative number like -19, this implementation will return the negative number itself instead of the last digit because the first check n < 10 will always succeed for negative numbers.

src/Chapter2.hs
```diff
@@ -336,7 +349,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList i j a
+    | i <= j = take (j-i+1) (drop i a)
```
vrom911, line 353: Nice catch on the condition! 👍🏼 However, note that on the negative input it could show wrong results.

src/Chapter2.hs
```diff
@@ -501,7 +517,9 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird42 :: [Int] -> Bool
+isThird42 [_,_,42,_] = True
```
vrom911, line 521: This pattern matching is for 4-elements-only list with the `42` on the third position as it is written in `[]`.
If you want to patent match on __any-sized__ list with 42, you need to use `( : )` construction to pattern match on:

 ```suggestion
isThird42 (_:_:42:_) = True
```

File: src/Chapter2.hs [hunk 8]
```diff
@@ -606,7 +624,8 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate (x:xs) = [x, x] ++ (duplicate xs)
```
vrom911, line 627: Great one!

It is more efficient to construct lists using the cons operator directly without introducing intermediate values.
```suggestion
duplicate (x:xs) = x : x : duplicate xs
```

src/Chapter2.hs
```diff
@@ -621,7 +640,10 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven (x:y:xs) = x : takeEven xs
+takeEven [x] = [x]
+takeEven [] = []
```
vrom911, line 646: Great pattern matching skills 💯

src/Chapter2.hs
```diff
@@ -728,7 +750,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concat (map (\x -> replicate x x) l)
```
vrom911, line 753: Awesome! 👍🏼 

src/Chapter2.hs
```diff
@@ -741,7 +763,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: (Eq a) => a -> [[a]] -> [[a]]
+contains x = filter (\y -> elem x y)
```
vrom911, line 767: 💪🏼 
Note that you can eta-reduce in the lambda function too 🙂 

```suggestion
contains x = filter (elem x)
```

src/Chapter2.hs
```diff
@@ -781,13 +804,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: (Num a) => [a] -> [a] -> [a]
+pairMul = zipWith (*)
```
vrom911, line 815: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -842,7 +867,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate x l
+    | x < 0 = []
+    | otherwise = (drop x l) ++ (take x l)
```
vrom911, line 873: Very interesting solution 🎨 👏🏼 

Although it won't cover cases when the rotation argument is bigger than the list's length. 

You can use `take` with the `cycle` function as an alternative solution 🙂

src/Chapter2.hs
```diff
@@ -858,7 +886,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind (x:xs) = (rewind xs) ++ [x]
```
vrom911, line 890: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Great Chapter One and Two! 🎉 
Very interesting and thoughtful solutions 👍🏼

src/Chapter2.hs
zexee: Thank you very much!!

vrom911:
@zexee If you'd fancy our review on other chapter feel free to open separate PR and summon us so we know that you want us to take a look 🙂 I will close this one for now, just to clarify if you want to share your other solutions with us 🤗 

Merry Christmas 🎄

PR #374, opened 2021-01-09
Chapter 4 solution

imsekun:
### Solutions for Chapter 4

This was a very challenging chapter, but it was fun!

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -37,8 +37,9 @@ if you want some feedback on your solutions.
 Perfect. Let's crush this!
 -}
 
-{-# LANGUAGE ConstraintKinds #-}
-{-# LANGUAGE InstanceSigs    #-}
+{-# LANGUAGE ConstraintKinds     #-}
+{-# LANGUAGE InstanceSigs        #-}
+{-# LANGUAGE FlexibleContexts    #-}
 
 module Chapter4 where
 
@@ -130,7 +131,7 @@ Either :: * -> * -> *
 Trinity :: * -> * -> * -> *
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
-IntBox :: * -> *
+IntBox :: (* -> *) -> *
 -}
 
 {- |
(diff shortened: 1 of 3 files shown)


src/Chapter4.hs
```diff
@@ -611,6 +630,19 @@ Implement the 'Monad' instance for our lists.
...
+  Cons a as >>= f = go (f a) (as >>= f)
+    where go :: List a -> List a -> List a
+          go Empty bs = bs
+          go (Cons a as) bs = Cons a (go as bs)
+
+replicate' :: Int -> a -> List a
+replicate' n x
+  | n == 1 = Cons x Empty
+  | n > 1 = Cons x (replicate' (n - 1) x)
+  | otherwise = Empty
```
chshersh, line 645: This function would be a great example of using the recursive `go` pattern! In the beginning, you can check all invalid inputs. E.g. if the number is negative and return an empty list immediately. And then, in the recursive `go` you can assume that it's possible and implement a simple recursive function with two pattern cases.

src/Chapter4.hs
```diff
@@ -629,7 +661,14 @@ Can you implement a monad version of AND, polymorphic over any monad?
...
 andM :: (Monad m) => m Bool -> m Bool -> m Bool
-andM = error "andM: Not implemented!"
+andM m1 m2 = do
+  a <- m1
+
+  if a then do
+    b <- m2
+
+    pure b
+  else pure a
```
chshersh, line 671: Great implementation! And a nice way of using the `do`-notation 💎 

If you want to practice with monads more, you can try implementing this function using the `>>=` operator directly, and it will be a nice one-liner 🙂 

Also, in the end, you don't need to return `pure a`, you can return `pure False`, because in the `else` branch you already know that `a` is `False`

chshersh (approved):
:tada: Fantastic work!

chshersh:
@imsekun I also see that according to the diff, there are some typos in Chapter Three. We can't merge this PR to our repository, because it contains solutions. But if you could submit only fixes to typos separately, we would greatly appreciate that 🙏🏻

vrom911:
Great one! Thank you for taking this course 🙏🏼

PR #375, opened 2021-01-12
finished chapter 1

golanor:
### Solutions for Chapter 1

cc @vrom911 @chshersh

I can't seem to run the tests on my environment - after installing the modules it can't find any of the packages.

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
@@ -301,43 +301,43 @@ expressions in GHCi
   functions and operators first. Remember this from the previous task? ;)
 
 >>> 1 + 2
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+3
 
 >>> 10 - 15
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+-5
 
 >>> 10 - (-5)  -- negative constants require ()
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+15
 
 >>> (3 + 5) < 10
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> True && False
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+False
 
 >>> 10 < 20 || 20 < 5
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> 2 ^ 10  -- power
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+1024
 
 >>> not False
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> div 20 3  -- integral division
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+6
 
 >>> mod 20 3  -- integral division remainder
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+2
 
 >>> max 4 10
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+10
 
 >>> min 5 (max 1 2)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+2
 
 >>> max (min 1 10) (min 5 7)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+5
 
 Because Haskell is a __statically-typed__ language, you see an error each time
 you try to mix values of different types in situations where you are not
@@ -429,6 +429,7 @@ task is to specify the type of this function.
 49
 -}
 
+squareSum :: Integer -> Integer -> Integer
 squareSum x y = (x + y) * (x + y)
 
 
@@ -449,7 +450,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next x = x + 1
 
 {- |
 After you've implemented the function (or even during the implementation), you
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +491,9 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
```
chshersh, line 496: Your implementation is almost correct :ok: 

src/Chapter1.hs
```diff
@@ -519,8 +522,9 @@ branches because it is an expression and it must always return some value.
 👩‍🔬 Due to lazy evaluation in Haskell, only the expression from the branch
   satisfying the check will be returned and, therefore, evaluated.
 -}
+
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if (abs x) > (abs y) then y else x
```
chshersh, line 527: `()` are redundant here. I see how it can be confusing to see when to use them and when not to use, especially in the beginning. But it's usually better to remove redundant brackets for cleaner code :broom: 
```suggestion
closestToZero x y = if abs x > abs y then y else x 
```

src/Chapter1.hs
```diff
@@ -554,7 +558,11 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | (x > y && x < z) || (x < y && x > z) = x
+    | (y > x && y < z) || (y < x && y > z) = y
+    | (z > x && z < y) || (z < x && z > y) = z
```
chshersh, line 565: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

It looks like you can replace the last case with a single otherwise and return z in here.

Also, I think that you can use `>=` and `<=` operators instead of `<` and `>` to cover cases when some of the arguments are equal.

src/Chapter1.hs
```diff
@@ -568,7 +576,15 @@ True
...
-isVowel c = error "isVowel: not implemented!"
+
+isVowel :: Char -> Bool
+isVowel c
+    | c == 'a' = True
+    | c == 'e' = True
+    | c == 'i' = True
+    | c == 'o' = True
+    | c == 'u' = True
+    | not (c == 'a') = False
```
chshersh, line 587: In this cases, you can use `otherwise` to tell the compiler to return `False` on all other possible values
```suggestion
    | otherwise = False
```

src/Chapter1.hs
```diff
@@ -632,7 +648,10 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Integer -> Integer
+sumLast2 n = let x = mod n 10
+                 y = mod (div n 10) 10
+                 in x + y
```
chshersh, line 654: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
```

Also, extra handling of negative numbers is required here to pass all the tests 🍏

src/Chapter1.hs
```diff
@@ -653,7 +672,8 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Integer -> Integer
+firstDigit n = if (div n 10) == 0 then n else firstDigit (div n 10)
```
chshersh, line 676: Nice implementation! But I'm afraid, it hangs on negative numbers, because

```haskell
ghci> div (-5) 10
-1
```

chshersh (approved):

I can suggest paying more attention to corner cases.

And regarding your problem with building the modules, can you tell your OS, `ghc` and `cabal` versions, and how did you build the project? If you can paste your error, it would be also helpful in troubleshooting 🙂

golanor:
> Nice work on solving our challenges!
> And well done with the Advanced Task medal_sports
> 
> I left my review and some comments. Hope you find them helpful slightly_smiling_face
> I can suggest paying more attention to corner cases.
> 
> And regarding your problem with building the modules, can you tell your OS, `ghc` and `cabal` versions, and how did you build the project? If you can paste your error, it would be also helpful in troubleshooting slightly_smiling_face
Sure, and thanks for the helpful review!

ghc version:
`The Glorious Glasgow Haskell Compilation System, version 8.10.3`

cabal version:
`cabal-install version 3.2.0.0
compiled using version 3.2.1.0 of the Cabal library`

I'm using ArchLinux, and I installed Cabal and ghc using the arch repositories (using pacman).

When I run make test-chapter1 i get:
`
Use -v (or `:set -v` in ghci) to see a list of the files searched for.
|
15 | import qualified Distribution.Simple.PackageIndex as Pkg
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

/home/or/Code/haskell/learn4haskell/dist-newstyle/tmp/src-2264879/ghc-paths-0.1.0.12/dist/setup/setup.hs:17:1:
error:
Could not find module ‘System.Exit’
There are files missing in the ‘base-4.14.1.0’ package,
try running 'ghc-pkg check'.
Use -v (or `:set -v` in ghci) to see a list of the files searched for.
|
17 | import System.Exit
| ^^^^^^^^^^^^^^^^^^

/home/or/Code/haskell/learn4haskell/dist-newstyle/tmp/src-2264879/ghc-paths-0.1.0.12/dist/setup/setup.hs:18:1:
error:
Could not find module ‘System.IO’
There are files missing in the ‘base-4.14.1.0’ package,
try running 'ghc-pkg check'.
Use -v (or `:set -v` in ghci) to see a list of the files searched for.
|
18 | import System.IO
| ^^^^^^^^^^^^^^^^

/home/or/Code/haskell/learn4haskell/dist-newstyle/tmp/src-2264879/ghc-paths-0.1.0.12/dist/setup/setup.hs:19:1:
error:
Could not find module ‘Data.IORef’
There are files missing in the ‘base-4.14.1.0’ package,
try running 'ghc-pkg check'.
Use -v (or `:set -v` in ghci) to see a list of the files searched for.
|
19 | import Data.IORef
| ^^^^^^^^^^^^^^^^^

/home/or/Code/haskell/learn4haskell/dist-newstyle/tmp/src-2264879/ghc-paths-0.1.0.12/dist/setup/setup.hs:20:1:
error:
Could not find module ‘Data.Char’
There are files missing in the ‘base-4.14.1.0’ package,
try running 'ghc-pkg check'.
Use -v (or `:set -v` in ghci) to see a list of the files searched for.
|
20 | import Data.Char
| ^^^^^^^^^^^^^^^^

/home/or/Code/haskell/learn4haskell/dist-newstyle/tmp/src-2264879/ghc-paths-0.1.0.12/dist/setup/setup.hs:21:1:
error:
Could not find module ‘Data.Maybe’
There are files missing in the ‘base-4.14.1.0’ package,
try running 'ghc-pkg check'.
Use -v (or `:set -v` in ghci) to see a list of the files searched for.
|
21 | import Data.Maybe
| ^^^^^^^^^^^^^^^^^

/home/or/Code/haskell/learn4haskell/dist-newstyle/tmp/src-2264879/ghc-paths-0.1.0.12/dist/setup/setup.hs:22:1:
error:
Could not find module ‘System.Directory’
There are files missing in the ‘directory-1.3.6.0’ package,
try running 'ghc-pkg check'.
Use -v (or `:set -v` in ghci) to see a list of the files searched for.
|
22 | import System.Directory
| ^^^^^^^^^^^^^^^^^^^^^^^
)
Failed to build exe:hsc2hs from hsc2hs-0.68.7 (which is required by
test:doctest-chapter1 from learn4haskell-0.0.0.0). See the build log above for
details.

make: *** [Makefile:11: test-chapter1] Error 1
`

vrom911:
I am not an Arch user, so can't help much with this platform, sorry. 

But here is a discussion which may have some useful links for you:
https://github.com/kowainik/stan/issues/365

PR #377, opened 2021-01-14
Chapter2

DamjanOstrelic:
### Solutions for Chapter 2

cc @vrom911 @chshersh

Can't get one test to pass on a formatting issue I suppose!

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -186,7 +186,7 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
-[10, 2, 3, 1, 5]
+[10,2,3,1,5]
 >>> [] ++ [1, 4]  -- [] is an empty list
 [1, 4]
 >>> 3 : [1, 2]
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create a list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
@@ -186,31 +186,31 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
-
+[10, 2, 3, 1, 5]
 >>> [] ++ [1, 4]  -- [] is an empty list
-
+[1, 4]
 >>> 3 : [1, 2]
-
+[3, 1, 2]
 >>> 4 : 2 : [5, 10]  -- prepend multiple elements
-
+[4, 2, 5, 10]
 >>> [1 .. 10]  -- list ranges
-
+[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
 >>> [10 .. 1]
-
+[]
 >>> [10, 9 .. 1]  -- backwards list with explicit step
-
+[10,9,8,7,6,5,4,3,2,1]
 >>> length [4, 10, 5]  -- list length
-
+[3]
 >>> replicate 5 True
-
+[True, True, True, True, True]
 >>> take 5 "Hello, World!"
-
+"Hello"
 >>> drop 5 "Hello, World!"
-
+", World"
 >>> zip "abc" [1, 2, 3]  -- convert two lists to a single list of pairs
-
+[('a', 1), ('b', 2), ('c', 3)]
 >>> words "Hello   Haskell     World!"  -- split the string into the list of words
-
+["Hello", "Haskell", "World!"]
 
 
 👩‍🔬 Haskell has a lot of syntax sugar. In the case with lists, any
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -349,7 +353,11 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 -- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf l
+    | len <= 1 = l
+    | otherwise = take (len `div` 2) l
+  where len = length l
```
vrom911, line 360: Nice usage of the `where` construction! 💯

src/Chapter2.hs
```diff
@@ -621,7 +634,12 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven [x] = [x]
+takeEven [x, _] = [x]
```
vrom911, line 640: That's right! 💪🏼 

If you put these two cases after the last check here, you could join them into just one case:

```haskell
takeEven (x:_) = [x]
```

src/Chapter2.hs
```diff
@@ -728,7 +746,14 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate (x:xs) = replicate x x ++ smartReplicate xs
```
vrom911, line 750: This is right! 💯 

```haskell
[3,1,2,4,0,1]
```

src/Chapter2.hs
```diff
@@ -728,7 +746,14 @@ value of the element itself
...
+
+{- How can we do this with map + replicate? The closes I got was:
+smartReplicate xs = concat (map (replicate x) xs)
+where x = head xs
+, but that just replicates each element by the first elements value.
+Is there a way to get the current element value when using map? -}
```
vrom911, line 756: I hope the previous comment answers this question, but let me know if you'd like to discuss it more detailed 🙂

src/Chapter2.hs
```diff
@@ -741,7 +766,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains _ [] = []
+contains x xs = filter (elem x) xs
```
vrom911, line 770: you don't have to check the empty list separately as `filter` works on any lists 🙂 

Also, then you can see that you can apply eta-reduction to this function as well  🙂

src/Chapter2.hs
```diff
@@ -781,13 +807,16 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: Num a => [a] -> [a] -> [a]
+pairMul = zipWith (*)
```
vrom911, line 818: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -842,7 +871,14 @@ list.
...
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate _ [] = []
+rotate 0 xs = xs
+rotate a xs
+  | a < 0 = []
+  | otherwise =
+    take l (drop a (cycle xs))
```
vrom911, line 880: Nice catch! You can actually even extend it further!  It does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Using `mod` you can see a way to optimize this :slightly_smiling_face: 

(Hint: you maybe will need to join these two cases )

src/Chapter2.hs
```diff
@@ -858,7 +894,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [Int] -> [Int]
+rewind [] = []
+rewind xs = last xs : rewind (init xs)
```
vrom911, line 899: Your solution is correct! However, it is slow. In lists, it is quite slow to add or take anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 But `last` function is very slow and need to walk the whole list at each iteration.

src/Chapter2.hs
```diff
@@ -858,7 +894,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [Int] -> [Int]
```
vrom911, line 897: Here you can have a less strict type, as you don't really use the fact that `Int`s are inside in the implementation. So it could be:

```suggestion
rewind :: [a] -> [a]
```

vrom911 (approved):
Awesome work on the chapter 🚆 🥳

src/Chapter2.hs
DamjanOstrelic: Oh, I see, just got used to having the most complicated logic at the end. I think the `takeEven [x, _] = [x]` case is actually redundant. How about just removing it:

```
takeEven :: [a] -> [a]
takeEven [] = []
takeEven [x] = [x]
takeEven (x : _ : xs) = x : takeEven xs
```

Because `xs` in the last case can be an empty list so any list with just 2 elements is also covered with that case.

src/Chapter2.hs
DamjanOstrelic: Would really like to get to the bottom of it - I understand your comment above, however, what I can't figure out is what value to pass to the partial `replicate` function that will be used to map over the list:

`smartReplicate xs = concat (map (replicate x) xs)`

I think this is along the right lines, but how do I specify what the correct `x` is?

src/Chapter2.hs
DamjanOstrelic: I see:

`contains x = filter (elem x)` is all that's needed

DamjanOstrelic:
Thanks for the comments, will have a think about the `rewind` and `rotate` functions 👍

src/Chapter2.hs
vrom911: Yes, this is absolutely right and quite elegant 💯

src/Chapter2.hs
vrom911: You need to pass to the `replicate` function the current element you are on while mapping the list, so it may look like this:

```haskell
smartReplicate xs = concat (map (\x -> replicate x x) xs)
```

This mean, that we only need to pass the function `replicate to `map`. Remember that the `map` has the following type:

```haskell
map :: (a -> b) -> [a] -> [b]
```

So we need to give it just a function that somehow transforms the current single element in the list. It has the type `a -> b` and in our case , we could think of such function as the following `f` function:

```haskell
f :: Int -> [Int]
f I = replicate i i
```

Remember that `replicate :: Int -> Int -> [Int]` so the types are correct.
And as I mentioned above, you can use `concatMap` instead of `concat` and `map`:

```haskell
smartReplicate = concatMap f
  where
    f :: Int -> [Int]
    f I = replicate i i
```

Does this makes sense?

PR #378, opened 2021-01-19
Chapter3

DamjanOstrelic:
### Solutions for Chapter 3

cc @vrom911 @chshersh

Have to come back to task 9.

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -1065,13 +1065,14 @@ instance Append [a] where
     append :: [a] -> [a] -> [a]
     append x y = x ++ y
 
-instance (Append a) => Append (Maybe a) where
+instance Append a => Append (Maybe a) where
     append :: Maybe a -> Maybe a -> Maybe a
     append (Just x) (Just y) = Just (append x y)
     append (Just x) _ = Just x
     append _ (Just y) = Just y
     append _ _ = Nothing
 
+{- I get the idea behind this function, but what's the valid syntax to call it?? -}
 
 {-
 =🛡= Standard Typeclasses and Deriving
@@ -1192,27 +1193,7 @@ properties using typeclasses, but they are different data types in the end.
 Implement data types and typeclasses, describing such a battle between two
 contestants, and write a function that decides the outcome of a fight!
 -}
-class Entity x where
-    attack :: x -> Int
 
-data T9Knight = T9Knight
-    {
-        t9knightHealth :: Int,
-        t9knightAttack :: Int,
-        t9knightDefence :: Int,
-        t9knightActions :: [KnightAction]
-    } deriving (Show)
-data T9Monster = T9Monster
-    {
-        t9MonsterHealth :: Int,
-        t9MonsterAttack :: Int
-    } deriving (Show)
-
-data KnightAction
-    = KnightAttack
-    | KnightPotion
-    | KnightSpell
-    deriving (Show)
 
 
 {-
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -489,6 +489,11 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data Meals
+    = Breakfast
+    | Lunch
+    | Dinner
+
 {- |
 =⚔️= Task 4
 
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -480,6 +514,76 @@ After defining the city, implement the following functions:
...
+{- Please help me understand best practices because of immutability:
+with the above definitions loaded the following function:
+    `myCity = buildCastle myCity "myCastle"` will run successfully,
+however, when `myCity` is called it afterwards it will hang.
+I am guessing this is because of immutability.
+So if I instead run:
+    `newCity = buildCastle myCity "myCastle"`
+both `myCity` and `newCity` are accessible.
+How can I use `buildCastle` without having to change the city name?
+ -}
```
chshersh, line 585: You are correctly noticed that with the `newCity` definition everything will work 👍🏻 One reason for this is immutability, and the second reason is name shadowing. 

So, when you write the following code on the top-level:

```haskell
myCity = City myCastle myStructure myHouses 
myCity = buildCastle myCity "myCastle"
```

what actually happens, is that you first define a top-level value called `myCity`, and then you redefine this top-level value. But on the second line, the `myCity` argument passed to `buildCastle` is the new `myCity`, not the old one. When you introduced `myCity` on the second time, you redefined and used the new redefined definition at the same time. 

This is called "names shadowing", and when you compile Haskell programs with warnings, GHC tells you about such situations, so you don't redefine your existing variables.

src/Chapter3.hs
```diff
@@ -561,21 +665,51 @@ introducing extra newtypes.
...
+calculatePlayerDamage :: Attack -> Strength -> Damage
+calculatePlayerDamage attack strength = Damage (attackValue attack + strengthValue strength)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense armor dexterity = Defense (armorValue armor * dexterityValue dexterity)
+
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit damage defense health = Health (healthValue health + defenseValue defense - damageValue damage)
```
chshersh, line 712: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -910,6 +1055,24 @@ Implement instances of "Append" for the following types:
...
 
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append (Gold x) (Gold y) = Gold (x + y)
+
+newtype Gold = Gold Int
+
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append x y = x ++ y
```
chshersh, line 1066: You can eta-reduce the implementation of `append` for brevity. So you can see how easy it is to create instances :slightly_smiling_face: 
```suggestion
    append = (++)
```

src/Chapter3.hs
```diff
@@ -910,6 +1055,24 @@ Implement instances of "Append" for the following types:
...
+
+instance Append a => Append (Maybe a) where
+    append :: Maybe a -> Maybe a -> Maybe a
+    append (Just x) (Just y) = Just (append x y)
+    append (Just x) _ = Just x
+    append _ (Just y) = Just y
+    append _ _ = Nothing
```
chshersh, line 1073: This is an absolutely correct solution!

src/Chapter3.hs
```diff
@@ -971,6 +1134,31 @@ implement the following functions:
...
+    | Wednesday
+    | Thursday
+    | Friday
+    | Saturday
+    | Sunday
+    deriving (Show, Eq, Enum, Bounded)
+
+isWeekend :: Day -> Bool 
+isWeekend day
+    | elem day [Friday, Saturday, Sunday] = True
```
chshersh, line 1149: I think `Friday` is not a part of the weekend 🤔 
But otherwise, the function definition looks good 👍🏻 

The only improvement I can think of is that `elem` already returns `Bool`, so you don't need guards, and can write just:

```haskell
isWeekend day = elem day [Saturday, Sunday]
```

src/Chapter3.hs
```diff
@@ -971,6 +1134,31 @@ implement the following functions:
...
+
+isWeekend :: Day -> Bool 
+isWeekend day
+    | elem day [Friday, Saturday, Sunday] = True 
+    | otherwise = False
+
+nextDay :: Day -> Day
+nextDay day
+    | day == maxBound = minBound
+    | otherwise = succ day
```
chshersh, line 1155: That is right! 💯 

Your solution is also a bit more maintainable. If you decide, for example, to have the "American" system of week days in the data type constructors, you won't need to change this function at all! 🙂

src/Chapter3.hs
```diff
@@ -971,6 +1134,31 @@ implement the following functions:
...
+
+daysToParty :: Day -> Int 
+daysToParty day
+    | fromEnum day > fromEnum Friday = fromEnum Friday - fromEnum day + fromEnum (maxBound :: Day) + 1
+    | otherwise = fromEnum Friday - fromEnum day
```
chshersh, line 1160: That's a great implementation! 👏🏻 

If you remember how `mod` works for negative numbers, you can implement this function even shorter 🙂

chshersh (approved):
Great work on Chapter Three!

Your code looks nice 👍🏻  I left some feedback and general suggestions

DamjanOstrelic:
@chshersh Thanks a lot for the feedback, sorry for the late response, I somehow missed the notifications! :)

chshersh:
@DamjanOstrelic No worries at all!

PR #381, opened 2021-02-07
Chapter2

GalinndanGalanodel:
Solutions for chapter2-revised. 
If it is still ok to ask for feedback, I've revised my solutions after previous feedback, but I wanted to see if the changes were ok.

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -556,7 +561,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+  | c == 'a' || c == 'e' || c == 'o' || c == 'i' || c == 'u' = True
+  | otherwise = False
```
chshersh, line 567: Since both `==` and `||` operators return `Bool`, you can avoid using guards and just return the result of the or-expression:

```suggestion
isVowel c = c == 'a' || c == 'e' || c == 'o' || c == 'i' || c == 'u'
```

src/Chapter2.hs
```diff
@@ -336,7 +336,13 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList a b xs
+  | b >= a = []
```
chshersh, line 340: I think here you wanted to use a different comparison operator. When you call `subList from to` and `to` is lower than `from`, then the resulting list is empty
```suggestion
  | b < a = []
```

src/Chapter2.hs
```diff
@@ -349,8 +355,8 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 -- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
-
+firstHalf :: [a] -> [a]
+firstHalf l = let x = div (length l) 2 in take x l
```
chshersh, line 359: Great usage of variables! 👏🏻

src/Chapter2.hs
```diff
@@ -620,7 +629,10 @@ Write a function that takes elements of a list only on even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven [x] = [x]
+takeEven (x:_:xs) = x : takeEven xs
```
chshersh, line 635: Awesome pattern-matching skills 💎

src/Chapter2.hs
```diff
@@ -727,7 +739,8 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate (x:xs) = concatMap (replicate x) (smartReplicate xs)
```
chshersh, line 743: I think this implementation is not correct. It will replicate every number in the list (except the first one) as many times as the value of the head of the list. But what you want is to replicate each number as many times as the value of that number itself.

Try to implement this function by `concatMap myReplicate` and define `myReplicate` as a helper function that does the main work. After that, you probably can replace your custom `myReplicate` with a lambda function for shorter code 🙂

src/Chapter2.hs
```diff
@@ -740,8 +753,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: Int -> [[Int]] -> [[Int]]
+contains a l = filter (elem a) l
```
chshersh, line 757: You can eta-reduce `l` here as well 🙂 
```suggestion
contains a = filter (elem a)
```

src/Chapter2.hs
```diff
@@ -841,7 +856,11 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate a l
+  | null l = []
+  | a <= 0 = []
+  | otherwise = take (length l) (drop a (cycle l))
```
chshersh, line 863: That's a correct solution! Though, it does a little bit more work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

chshersh (approved):
Hi @GalinndanGalanodel,

Great revised solutions! 🔝 
And nice job on the advanced task 🧩 

I left some more feedback, but your code already looks great 👍🏻

vrom911:

PR #382, opened 2021-02-13
Chapter 1

dangdennis:

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -631,9 +645,17 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
+sumLast2 :: Int -> Int
+sumLast2 n = penultimate n + last n
+  where
+    penultimate :: Int -> Int
+    penultimate x = mod (div x 10) 10
```
chshersh, line 652: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 
I can add to use `abs` to handle negative numbers as well.

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
```

src/Chapter1.hs
```diff
@@ -652,8 +674,10 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+  | n < 10 = n
+  | n >= 10 = firstDigit (div n 10)
```
chshersh, line 680: Great recursive implementation 💯 

You probably can think of a few ways to make this function work for negative numbers as well 🙂 Otherwise it returns the negative number itself immediately instead of the last digit.

chshersh (approved):
Great work!
And nicely done on the advanced task 💎

dangdennis:
Thank you so much for the help!

PR #384, opened 2021-02-17
Completed chapter 1 tasks

saatvikshah:
### Solutions for Chapter 1
Thanks a lot for setting up Learn4Haskell - its an awesome initiative! Definitely hoping to get feedback.
cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -491,7 +491,7 @@ Implement a function that returns the last digit of a given number.
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit :: Int -> Int
-lastDigit = (`mod` 10) . (abs)
+lastDigit = (`mod` 10) . abs
 
 
 {- |
@@ -521,7 +521,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = if (abs x) < (abs y) then x else y
+closestToZero x y = if abs x < abs y then x else y
 
 
 {- |
@@ -642,13 +642,13 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 sumLast2 :: Int -> Int
-sumLast2 = positiveSumLast2 . abs
+sumLast2 = go . abs
   where
-    positiveSumLast2 :: Int -> Int
-    positiveSumLast2 n = last + secondLast
+    go :: Int -> Int
+    go n = lastNum + secondLastNum
       where
-        last = n `mod` 10
-        secondLast = ((`mod` 10).(`div` 10)) n
+        (divNum, lastNum) = n `divMod` 10
+        secondLastNum = divNum `mod` 10
 
 {- |
 =💣= Task 10*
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +490,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit = (`mod` 10) . (abs)
```
vrom911, line 494: Great implementation! We have not explained eta-reduction and function composition, but you got it right 💪🏼 

One minor thing: parenthesis are redundant here, you can clean it up a bit:

```suggestion
lastDigit = (`mod` 10) . abs
```

src/Chapter1.hs
```diff
@@ -520,7 +521,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if (abs x) < (abs y) then x else y
```
vrom911, line 524: Here as well 🙂 

```suggestion
closestToZero x y = if abs x < abs y then x else y
```

src/Chapter1.hs
```diff
@@ -631,9 +641,14 @@ Implement a function that returns the sum of the last two digits of a number.
...
-
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 = positiveSumLast2 . abs
+  where
+    positiveSumLast2 :: Int -> Int
+    positiveSumLast2 n = last + secondLast
+      where
+        last = n `mod` 10
```
vrom911, line 650: `last` is the name of a standard function. Haskell allows name shadowing, that's why it works. Though, in practice, it's usually better to avoid name clashes with other functions cause it could be confusing 🙂

src/Chapter1.hs
```diff
@@ -631,9 +641,14 @@ Implement a function that returns the sum of the last two digits of a number.
...
+        secondLast = ((`mod` 10).(`div` 10)) n
```

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
(x, y) = divMod n 10
```

Also, in Haskell operators are separated with spaces 👾  So you can add them for `.` as well 🙂

src/Chapter1.hs
```diff
@@ -652,8 +667,13 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit = (positiveFirstDigit . abs)
```
vrom911, line 671:
```suggestion
firstDigit = positiveFirstDigit . abs
```

vrom911 (approved):
Great work! Very nice solutions 👏🏼 And amazing work on the advanced task 🧁

src/Chapter1.hs
saatvikshah: Ack

src/Chapter1.hs
saatvikshah: I setup the Vscode Haskell plugin which seems to act up at times, but has begun to reliably show me redundant parens :)

src/Chapter1.hs
saatvikshah: Ack - adjusted to
```
...
(divNum, lastNum) = n `divMod` 10
secondLastNum = divNum `mod` 10
```

src/Chapter1.hs
saatvikshah: Ack, I also shadowed the `elem` fn here it seems.

vrom911:
The changes look great! I am going to close this PR 🙂 💪🏼 👏🏼

PR #385, opened 2021-02-20
Chapter 1, 2 solutions

voldemoriarty:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -520,7 +521,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if (abs x) < (abs y) then x else y
```
chshersh, line 524: `()` are redundant here. I see how it can be confusing to see when to use them and when not to use, especially in the beginning. But it's usually better to remove redundant brackets for cleaner code :broom: 
```suggestion
closestToZero x y = if abs x < abs y then x else y
```

src/Chapter1.hs
```diff
@@ -554,7 +555,13 @@ value after "=" where the condition is true.
...
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+	| (x >= y) && (y >= z) = y
+	| (z >= y) && (y >= x) = y
+	| (y >= x) && (x >= z) = x
+	| (z >= x) && (x >= y) = x
+	| otherwise = z
```
chshersh, line 564: Nice usage of guards! 💂🏻

src/Chapter1.hs
```diff
@@ -568,7 +575,14 @@ True
...
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+	| c == 'a' || c == 'A' = True
+	| c == 'e' || c == 'E' = True
+	| c == 'i' || c == 'I' = True
+	| c == 'o' || c == 'O' = True
+	| c == 'u' || c == 'U' = True
+	| otherwise = False
```
chshersh, line 585: That's a great implementation! 👍🏻 In the future, you can improve it by using standard function [toLower](http://hackage.haskell.org/package/base-4.14.1.0/docs/GHC-Unicode.html#v:toLower), but we explain imports of other functions only in the next chapter, so no need to do anything now except being aware of alternatives 🙂

src/Chapter1.hs
```diff
@@ -632,8 +646,14 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = (d1 n) + (d2 n)
+	where 
+		d1 :: Int -> Int
+		d1 x = mod (abs x) 10
 
+		d2 :: Int -> Int
+		d2 x = d1 (div (abs x) 10)
```
chshersh, line 656: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
```

src/Chapter1.hs
```diff
@@ -653,7 +673,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n = if (an) < 10 then (an) else firstDigit (div an 10)
+	where 
+		an = abs n
```
chshersh, line 679: Great usage of `where` for creating the local variable and using it! 👏🏻

chshersh (approved):
Awesome work on Chapter One!
And a wonderful solution to the advanced task 🏆

voldemoriarty:
cc @vrom911 @chshersh 

I've commited Chapter 2 solutions here if that's okay

src/Chapter2.hs
```diff
@@ -336,7 +336,7 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList b e l = if (b < 0 || e < 0) then [] else take (e-b+1) (drop b l)
```
chshersh, line 339: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding a separate case to your check :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -621,7 +627,10 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven (zero:_:xs) = zero:takeEven xs
+takeEven (zero:_) = [zero]
```
chshersh, line 633: Great pattern-matching skills! 🥇

src/Chapter2.hs
```diff
@@ -728,7 +737,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concatMap (\x -> replicate x x) l
```
chshersh, line 740: Great! 👍🏼 
You can eta-reduce this function as well 🙂

src/Chapter2.hs
```diff
@@ -741,7 +750,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains val list = filter (\l -> elem val l) list
```
chshersh, line 754: Now, after you've mastered eta-reduction, you can apply such a technique to this function as well  🙂 
But your implementation is already great 💯 Here you have two opportunities for eta-reduction:

* Lambda
* The `list` argument

src/Chapter2.hs
```diff
@@ -842,7 +854,8 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate n l = if (n >= 0) then take (length l) (drop n (cycle l)) else []
```
chshersh, line 858: That's a correct solution! Though, it does a little bit more work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -858,7 +871,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind l = (rewind (tail l)) ++ [(head l)]
```
chshersh, line 876: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
@voldemoriarty Chapter 2 looks good 👍🏻 
I reviewed the solutions to this chapter. The code looks great 👏🏻

vrom911:

PR #386, opened 2021-02-20
Chapter 1 solved.

CiottiGiorgio:

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -561,7 +561,7 @@ mid :: Int -> Int -> Int -> Int
 mid x y z
   | z <= x && x <= y || y <= x && x <= z    = x
   | x <= y && y <= z || z <= y && y <= x    = y
-  | x <= z && z <= y || y <= z && z <= x    = z
+  | otherwise                               = z
 
 {- |
 =⚔️= Task 8
@@ -679,9 +679,9 @@ firstDigit :: Int -> Int
 firstDigit n =
   let m = abs n
   in if m < 10
-    then m
-    else let r = div m 10
-      in firstDigit r
+      then m
+      else let r = div m 10
+           in firstDigit r
 
 
 {-
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
@@ -301,43 +301,43 @@ expressions in GHCi
   functions and operators first. Remember this from the previous task? ;)
 
 >>> 1 + 2
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+3
 
 >>> 10 - 15
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+-5
 
 >>> 10 - (-5)  -- negative constants require ()
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+15
 
 >>> (3 + 5) < 10
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> True && False
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+False
 
 >>> 10 < 20 || 20 < 5
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> 2 ^ 10  -- power
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+1024
 
 >>> not False
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> div 20 3  -- integral division
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+6
 
 >>> mod 20 3  -- integral division remainder
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+2
 
 >>> max 4 10
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+10
 
 >>> min 5 (max 1 2)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+2
 
 >>> max (min 1 10) (min 5 7)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+5
 
 Because Haskell is a __statically-typed__ language, you see an error each time
 you try to mix values of different types in situations where you are not
@@ -428,7 +428,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Int -> Int -> Int
 squareSum x y = (x + y) * (x + y)
 
 
@@ -449,7 +449,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next x = x + 1
 
 {- |
 After you've implemented the function (or even during the implementation), you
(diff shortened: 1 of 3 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +490,10 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = 
+  let m = abs n
+  in mod m 10
```
vrom911, line 496: Nice usage of the `let-in` construction! 👍🏼

src/Chapter1.hs
```diff
@@ -554,7 +557,11 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | z <= x && x <= y || y <= x && x <= z    = x
+  | x <= y && y <= z || z <= y && y <= x    = y
+  | x <= z && z <= y || y <= z && z <= x    = z
```

Because of that, you need to add another guard – `| otherwise = ...`, to tell the compiler, that your pattern matching is exhaustive 🙂 But actually, you can replace the last case with it, as you don't need to check this condition. If previous cases are not satisfied, the last one is definitely truth:

```suggestion
  | otherwise = z
```

src/Chapter1.hs
```diff
@@ -632,7 +646,15 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+  let m = abs n
+  in lastDigit m + firstToLastDigit m
+  where
+    firstToLastDigit :: Int -> Int
+    firstToLastDigit r = 
+      let k = div r 10
+      in mod k 10
```
vrom911, line 657: Very nice solution 👏🏼 And reusage of `lastDigit` 🔝

vrom911 (approved):
Great work on the chapter, congrats 🎉 
And amazing work on the advanced task as well 👏🏼

src/Chapter2.hs
```diff
@@ -349,7 +372,8 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 -- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf l = subList 0 (div (length l) 2 - 1) l
```
vrom911, line 376: Clever reusage of the `sublist` function 👍🏼

src/Chapter2.hs
```diff
@@ -621,7 +648,10 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven [x] = [x]
+takeEven (x:_:xs) = x : takeEven xs
```
vrom911, line 654: Awesome pattern-matching skills 💪🏼

src/Chapter2.hs
```diff
@@ -741,8 +771,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: Int -> [[Int]] -> [[Int]]
+contains c listOfLists = filter (\x -> elem c x) listOfLists
```
vrom911, line 775: Now, after you've mastered eta-reduction, you can apply such a technique to this function and lambda-function as well  🙂 

src/Chapter2.hs
```diff
@@ -781,13 +811,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 822: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -858,7 +891,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
vrom911, line 896: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Another wonderful chapter 👏🏼

src/Chapter3.hs
```diff
@@ -460,6 +484,28 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
...
+  pastry :: String
+  , cCookingTime :: Int
+}
+
+data Steak = MkSteak {
+  nEggs :: Int
+  , pcCookingTime :: Int
+}
+
+data Meal = SecondBreakfast Croissant | Lunch Steak
```
chshersh, line 506: That's a nice idea you have here! 🥐

src/Chapter3.hs
```diff
@@ -480,6 +526,45 @@ After defining the city, implement the following functions:
    and at least 10 living __people__ inside in all houses of the city in total.
 -}
 
+data Wall = MkWall
+
+data Castle = ACastle String Wall | NoCastle
```
chshersh, line 531: With this data type representation, you will always have walls inside the castle. But the task says that a castle might not have a wall at all.

src/Chapter3.hs
```diff
@@ -480,6 +526,45 @@ After defining the city, implement the following functions:
...
+
+data RecreationBuilding = Church | Library
+
+data House = OneHouse | TwoHouse | ThreeHouse | FourHouse
```
chshersh, line 535: This is a nice and type-safe representation of number of people in the house 👍🏻

src/Chapter3.hs
```diff
@@ -560,22 +645,32 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (MkArmor x) (MkDexterity y) = MkDefense (x * y)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (MkDamage x) (MkDefense y) (MkHealth z) = MkHealth (x + y - z)
```
chshersh, line 673: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -910,6 +1019,18 @@ Implement instances of "Append" for the following types:
...
+  goldAmount :: Int
+}
+
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append (MkGold x) (MkGold y) = MkGold (x+y)
+
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append xs1 xs2 = xs1 ++ xs2
```
chshersh, line 1032: You can eta-reduce the implementation of `append` for brevity. So you can see how easy it is to create instances :slightly_smiling_face: 
```suggestion
    append = (++)
```

src/Chapter3.hs
```diff
@@ -971,6 +1092,42 @@ implement the following functions:
...
 -}
 
+data Weekday = Monday | Tuesday | Wednesday | Thursday | Friday | Saturday | Sunday
+  deriving (Show, Eq, Bounded, Enum)
+
+isWeekend :: Weekday -> Bool
+isWeekend day
+  | day == pred maxBound  = True
+  | day == maxBound       = True
+  | otherwise             = False
```
chshersh, line 1102: This is a nice and generic implementation!
Though, you can improve it even further to make it work automatically even if you decide to change to the American style of weekdays

src/Chapter3.hs
```diff
@@ -971,6 +1092,42 @@ implement the following functions:
...
+-- I would like to be able to express the cycling property of the function nextDay in terms of its minBound and maxBound
+-- I've had no luck so far...
+nextDay :: Weekday -> Weekday
+nextDay maxBound = minBound
+nextDay x = succ x
+-}
+
+-- I'm not quite happy with this either because I would have loved to use succ. Oh well w/e.
+nextDay :: Weekday -> Weekday
+nextDay x = cycle [minBound .. maxBound] !! (fromEnum x + 1)
```
chshersh, line 1114: Your commented solution doesn't work because during pattern matching you don't compare the input variable with `maxBound`. You just give the name `maxBound` to your variable, and the variable name is a pattern that always succeeds. This is possible due to _name shadoing_. In Haskell, you can override global names with local names. Fortunately, GHC produces a warning in such situations, so you can fix them.

Instead of pattern matching on `maxBound`, compare your input with `maxBound`. And you will be able to use `succ` in this case:

```haskell
nextDay day
    | day == maxBound = minBound
    | otherwise = ...
```

src/Chapter3.hs
```diff
@@ -971,6 +1092,42 @@ implement the following functions:
...
+daysToParty x       = 1 + daysToParty (nextDay x)
+-}
+
+-- Am I using a bazooka to shoot a fly here?
+daysToParty :: Weekday -> Int
+daysToParty = go 0
+  where
+    go :: Int -> Weekday -> Int
+    go n Friday = n
+    go n x = go (n + 1) (nextDay x)
```
chshersh, line 1129: You can write even better implementation using `fromEnum` and `mod`, especially if you remember how `mod` works for negative numbers from Chapter 1 🙂

chshersh (approved):
Great work on Chapter Three! 👏🏻

vrom911:

PR #387, opened 2021-02-20
Chapter1

minjipack:

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -644,15 +644,21 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 :: Int -> Int  
-sumLast2 n = div lastTwoDigits 10 + mod lastTwoDigits 10
-  where
-    lastTwoDigits = mod (abs n) 100
+sumLast2 :: Int -> Int
+sumLast2 n =
+ let (sl, l) = divMod (abs n) 10
+ in (mod sl 10) + l
 
 {-
 I got stuck for awhile as I did not carefully read integral division (div)
 num = abcd
 get c and get d and c + d
+
+Option1:
+sumLast2 :: Int -> Int  
+sumLast2 n = div lastTwoDigits 10 + mod lastTwoDigits 10
+  where
+    lastTwoDigits = mod (abs n) 100
 
 Option2:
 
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -449,7 +454,10 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next x = x + 1
+{-
+greater than max -> minimum
```
vrom911, line 459: Good catch!
We will discuss this in future chapters 😄

src/Chapter1.hs
```diff
@@ -632,7 +644,24 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int  
+sumLast2 n = div lastTwoDigits 10 + mod lastTwoDigits 10
```

```haskell
mod x 10
div x 10
```

```haskell
(x, y) = divMod ... 10
```

src/Chapter1.hs
```diff
@@ -653,8 +682,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit n = if absN < 10 then mod absN 10 else firstDigit (div absN 10)
```
vrom911, line 686: If `abs n < 10` you don't need to do `mod` on that 🙂 
So you can simplify it a bit:
```suggestion
firstDigit n = if absN < 10 then absN else firstDigit (div absN 10)
```

vrom911 (approved):
Great work! Very nice solutions! 🎉 

And awesome work on the advanced task as well 👏🏼

src/Chapter1.hs
minjipack: Oh yes! You are absolutely right. Thank you for your comment :)

src/Chapter1.hs
minjipack: Looking forward to future chapters  🍡

src/Chapter1.hs
minjipack: Thanks! `divMod` is a cool thing to know! I don't think I can come up with a shorter answer with `divMod` ...  :( The following is what I came up with.

```
sumLast2 :: Int -> Int
sumLast2 n =
 let (secondLast, last) = divMod n 10
 in (mod secondLast 10) + last
```

vrom911 (approved):
These changes are awesome! Ready to go! 🚢

vrom911:
I am going to close this PR as we can't merge it into the repo, but we are hoping to see you in the next Chapter 🎉

PR #388, opened 2021-02-21
Chapter 2 solutions

saatvikshah:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -354,6 +354,7 @@ ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
 subList start end xs
+  | end < start = []
   | start >= 0 && end >= 0 = (take (end - start + 1) . drop start) xs
   | otherwise = []
 
@@ -368,9 +369,7 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 firstHalf :: [a] -> [a]
-firstHalf xs
-  | even len = take (div len 2) xs
-  | otherwise = take (div (len-1) 2) xs
+firstHalf xs = take (div len 2) xs
   where
     len = length xs
 
@@ -876,8 +875,13 @@ list.
 -}
 rotate :: Int -> [a] -> [a]
 rotate num lst 
-  | num >= 0  = take (length lst) (drop num (cycle lst))
-  | otherwise = []
+  | num < 0 || len == 0 = []
+  | reducedNum == 0     = lst
+  | reducedNum >= 0     = take len (drop reducedNum (cycle lst))
+  | otherwise           = []
+  where
+    reducedNum = num `mod` len
+    len = length lst
 
 {- |
 =💣= Task 12*
@@ -894,11 +898,11 @@ and reverses it.
   cheating!
 -}
 rewind :: [a] -> [a]
-rewind lst = case lst of
-  [] -> []
-  [x] -> [x]
-  xs -> last xs :  rewind (init xs)
-
+rewind = go []
+  where
+    go :: [a] -> [a] -> [a]
+    go xs (y:ys) = go (y:xs) ys
+    go xs [] = xs
 
 {-
 You did it! Now it is time to open pull request with your changes
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,42 +136,48 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
+(:) :: a -> [a] -> [a]
 
 
 Reverse a list:
 >>> :t reverse
+reverse :: [a] -> [a]
 
 
 Take first N elements of a list:
 >>> :t take
+take :: Int -> [a] -> [a]
 
 
 Create a list from N same elements:
 >>> :t replicate
+replicate :: Int -> a -> [a]
 
 
 Split a string by line breaks:
 >>> :t lines
+lines :: String -> [String]
 
 
 Join a list of strings with line breaks:
 >>> :t unlines
+unlines :: [String] -> String
 
 
 -}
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +353,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList start end xs
+  | start >= 0 && end >= 0 = (take (end - start + 1) . drop start) xs
```
chshersh, line 357: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding one more check for this special case :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -348,9 +367,12 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
-
+firstHalf :: [a] -> [a]
+firstHalf xs
+  | even len = take (div len 2) xs
+  | otherwise = take (div (len-1) 2) xs
```
chshersh, line 373: You don't need to check `len` separately, because `div` by two for an odd number and the previous one is the same

src/Chapter2.hs
```diff
@@ -501,8 +523,10 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
-
+isThird42 :: [Int] -> Bool
+isThird42 xs = case xs of
+  (_ : _ : 42 : _) -> True 
+  _ -> False
```
chshersh, line 529: Great pattern matching skills! 🥇

src/Chapter2.hs
```diff
@@ -606,7 +630,9 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate lst = case lst of
+  (x:xs) -> x : x : duplicate xs
+  [] -> []
```
chshersh, line 635: I see that you prefer `case-of` expressions. In some cases, when the function is short, it's okay to pattern match on the top-level. Especially handy to avoid coming up with extra names for variables 🙂 
```suggestion
duplicate [] = []
duplicate (x:xs) = x : x : duplicate xs
```

src/Chapter2.hs
```diff
@@ -781,13 +812,14 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
--- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Ord a => a -> [a] -> [a]
+listElementsLessThan x = filter (<x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
chshersh, line 822: Eta-reduction award! 🏆 

Awesome implementations to all previous functions 💯

src/Chapter2.hs
```diff
@@ -842,7 +874,10 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate num lst 
+  | num >= 0  = take (length lst) (drop num (cycle lst))
```
chshersh, line 879: That's a correct solution! Though, it does a little bit more work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -858,7 +893,11 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind lst = case lst of
+  [] -> []
+  [x] -> [x]
+  xs -> last xs :  rewind (init xs)
```
chshersh, line 900: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. For instance, taking the last element of the list and removing it. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

You can read a bit more about the `go` pattern here: https://kowainik.github.io/posts/haskell-mini-patterns#recursive-go

chshersh (approved):
Wonderful work on all tasks! 👏🏻 🔝

src/Chapter2.hs
saatvikshah: Ack, I see because im wastefully dropping in that case - revised to:
```
subList start end xs
  | end < start = []
  | start >= 0 && end >= 0 = (take (end - start + 1) . drop start) xs
  | otherwise = []
```

src/Chapter2.hs
saatvikshah: 🤦‍♂️ - yep silly of me 👍

src/Chapter2.hs
saatvikshah: 😁

src/Chapter2.hs
saatvikshah: By `pattern match on the top-level` do you mean using the suggested change - I didn't fully understand how it avoids coming up with extra names for variables? Isn't the only difference going to be repeating the function name vs.  using `case varname of`?

src/Chapter2.hs
saatvikshah: Yep, can just rotate `num `mod` (length lst)` times.

src/Chapter2.hs
saatvikshah: At first I was a bit confused by the comment as I was indeed applying the cons operator - but I later understood the problem being (1) Getting `last xs` each time is going to be prohibitive because Haskell lists are Singly Linked lists (2) Doing the `init xs` each time will produce list copies.

Here is the revised `go` pattern solution:
```
rewind :: [a] -> [a]
rewind = go []
  where
    go :: [a] -> [a] -> [a]
    go xs (y:ys) = go (y:xs) ys
    go xs [] = xs
```

src/Chapter2.hs
chshersh: @saatvikshah Yes, the suggested change can be called as _pattern match on the top-level_. It avoids coming up for the variable name `lst`, becase with `case of` you need to use a variable for the list.

Generally speaking, top-level pattern matching vs case-of pattern matching is a matter of taste. I prefer to use `case-of` when the function name is long or body implementation for each case is long. However, it's not that convenient to pattern match on multiple arguments with `case-of`, so the top-level approach still can be preferred.

Anyway, use your own judgement in each case 🙂

chshersh:
@saatvikshah All the changes look great to me! 🔝 I'm going to close this PR as we can't merge solutions to our repo, but it looks great nevertheless 👍🏻

PR #391, opened 2021-03-01
Chapter2

minjipack:

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create a list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
@@ -186,31 +186,44 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
+[10, 2, 3, 1, 5]
 
 >>> [] ++ [1, 4]  -- [] is an empty list
+[1, 4]
 
 >>> 3 : [1, 2]
+[3, 1, 2]
 
 >>> 4 : 2 : [5, 10]  -- prepend multiple elements
+[4, 2, 5, 10]
 
 >>> [1 .. 10]  -- list ranges
+[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
 
 >>> [10 .. 1]
+[]
+{- Interesting... -}
 
 >>> [10, 9 .. 1]  -- backwards list with explicit step
+[10, 9, 8, 7, 6, 5, 4, 3, 2, 1]
 
 >>> length [4, 10, 5]  -- list length
+3
 
 >>> replicate 5 True
+[True, True, True, True, True]
 
 >>> take 5 "Hello, World!"
+"Hello"
 
 >>> drop 5 "Hello, World!"
+", Wrold!"
 
 >>> zip "abc" [1, 2, 3]  -- convert two lists to a single list of pairs
+[(a, 1), (b, 2), (c, 3)]
 
 >>> words "Hello   Haskell     World!"  -- split the string into the list of words
-
+["Hello", "Haskell", "World!"]
 
 
 👩‍🔬 Haskell has a lot of syntax sugar. In the case with lists, any
@@ -254,6 +267,7 @@ original list, so you don't need to worry about accidentally spoiling
 values of variables you defined before.
 -}
 
+{- TODO: Just out of curiosity, why cons-lists is called "cons"? -}
 {- |
 =🛡= List implementation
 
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +350,29 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList startPos endPos l
+  | startPos >= 0 && startPos <= endPos && endPos < len = drop startPos (take (endPos + 1) l)
+  | startPos >= 0 && startPos < endPos && endPos >= len = drop startPos l
+  | otherwise = []
```
vrom911, line 356: Interesting solution! 💡 

Due to the way `drop` and `take` functions work, you can simplify conditions a bit in here. Also, you could avoid using `length` right in the start. This is important because taking `length` is time-consuming operation and we always try to avoid this when possible.

As for solution, you can first start with the check on reasonable input. This will save time to check it first and return an empty list if the input is not correct.

Something like this will work:

```haskell
subList startPos endPos l
    | startPos < 0 || endPos < 0 || endPos < startPos = []
    | otherwise = take (endPos - startPos + 1) (drop startPos l)
```

src/Chapter2.hs
```diff
@@ -348,9 +384,10 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
-
+firstHalf :: [a] -> [a]
+firstHalf l = take halfLen l
+  where
+    halfLen = div (length l) 2
```
vrom911, line 390: Awesome usage of the `where` construction 🧁

src/Chapter2.hs
```diff
@@ -391,6 +428,22 @@ isZero 0 = True
 isZero n = False
 @
 
+{-
+TODO: did not know underscore is called hole. Do we also use this term in other languages?
```
vrom911, line 432: I'm afraid I can't certainly say that. `_` is called hole as you can put it as the placeholder to fill later. It is also called wildcard sometimes.

src/Chapter2.hs
```diff
@@ -391,6 +428,22 @@ isZero 0 = True
...
+MY UNDERSTANDING IN MY OWN LANGUAGE
+switch-case VS pattern-matching
+switch-case : developers need to explicitly define "break" or "return" for each case;
+whereas, Haskell internally handles "break" or "return" for us for a matching branch with a simpler syntax.
+In order to efficiently implement this pattern matching, cover specific cases(i.e., A, B, not (A or B)) first
+similar to how we come up with an efficient way to implement if-then-else.
+
+case-if VS pattern-matching
+'case' uses arrows (->) instead of "=" for branch results.
+when function names are long, or pattern-matching on functions is awkward
```
vrom911, line 444: This is a really nice sum-up! Good job 👍🏼

src/Chapter2.hs
```diff
@@ -621,7 +681,15 @@ Write a function that takes elements of a list only in even positions.
...
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven l = go (length l) l
+  where
+    go :: Int -> [a] -> [a]
+    go 0 _ = []
+    go 1 n = n
+    go 2 (x:_:xys) = x : xys
+    go i (x:_:xys) = x : (go (i-2) xys)
```
vrom911, line 691: Very creative! 🎨 

Note that you can implement this function with simple pattern matching as well. It requires fewer function calls and a bit more efficient by memory and time as you use `length`, which is not optimal.

src/Chapter2.hs
```diff
@@ -728,7 +796,14 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concat (map (\x -> replicate x x) l)
```
vrom911, line 799: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -741,9 +816,12 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+
+contains :: Int -> [[Int]] -> [[Int]]
+contains x l = filter (\e -> (elem x e)) l
```
vrom911, line 821: Now, after you've mastered eta-reduction, you can apply such a technique to this function and lambda function as well  🙂 

src/Chapter2.hs
```diff
@@ -741,9 +816,12 @@ the list with only those lists that contain a passed element.
...
 
 
+-- TODO: what does Eta mean? https://sookocheff.com/post/fp/eta-conversion/
```
vrom911, line 824: Eta is a greek letter to represent the process 🙂

src/Chapter2.hs
```diff
@@ -781,14 +859,24 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
+listElementsLessThan :: Int -> [Int] -> [Int]
 listElementsLessThan x l = filter (< x) l
 
--- Can you eta-reduce this one???
+-- Can you eta-reduce this one??? 
+{- NO 
+Why? I am guessing that if there are more than one parameters, the order of parameters matters.
+For example, zipWith (-) a b  VS zipWith (-) b a
```
vrom911, line 871: Yes, but they would be applied in the same order as they come, so no need to worry about that. You can think of it as 2-step eta reduction. First, you remove the last argument and get:

```haskell
pairMul xs = zipWith (*) xs
```

But you immediately see that the above function could be eta-reduced itself:

```haskell
pairMul = zipWith (*)
```

Hope this makes sense. Let me know if you want to discuss eta-reduction more! 🙂

src/Chapter2.hs
```diff
@@ -781,14 +859,24 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
+{- NO 
+Why? I am guessing that if there are more than one parameters, the order of parameters matters.
+For example, zipWith (-) a b  VS zipWith (-) b a 
+-}
+pairMul :: [Int] -> [Int] -> [Int]
 pairMul xs ys = zipWith (*) xs ys
 
+{-
+TODO: Curious about what it means by 
+- provide more composable interfaces.
```
vrom911, line 878: Is your question applied to something specific, or you wonder what is composability generally?

src/Chapter2.hs
```diff
@@ -858,8 +952,13 @@ and reverses it.
...
 -}
-rewind = error "rewind: Not Implemented!"
-
+rewind :: [a] -> [a]
+rewind l = go (length l) l
+  where
+    go :: Int -> [a] -> [a]
+    go 0 _ = []
+    go 1 l1 = l1
+    go i (x:xs) = (go (i-1) xs) ++ [x]
```
vrom911, line 961: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Awesome work! 💎 
Let me know if you have any questions 🤗

chshersh:

PR #392, opened 2021-03-02
Chapter 3 Solutions

saatvikshah:
### Solutions for Chapter 3

Very challenging compared to the previous weeks! Looking forward to reviews.

Follow up re: the "boss" question: I believe that trying to share characteristics like AttackDamage/Health is not exactly supported(as Haskell doesn't seem to have a way to do row polymorphism). So while I did end up using typeclasses to constraint/customize the behavior, it was a lot harder to reason about and I felt more verbose. How would you compare the implementation in this PR to [this one without using typeclasses](https://github.com/saatvikshah/learn4haskell/commit/e018cb9f7ade2cc4805feb5cec0749aab60c0c52)?

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -343,6 +343,12 @@ Define the Book product data type. You can take inspiration from our description
 of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
+data Book = MkBook
+  { bookName          :: String,
+    bookAuthor        :: String,
+    bookPublishedYear :: Int,
+    bookGenre         :: String
+  } deriving (Show)
 
 {- |
 =⚔️= Task 2
@@ -371,8 +377,31 @@ after the fight. The battle has the following possible outcomes:
  ⊛ Monster defeats the knight. In that case return -1
  ⊛ Neither the knight nor the monster wins. On such an occasion, the knight
    doesn't earn any money and keeps what they had before.
-
 -}
+data Knight = MkKnight
+  { knightHealth        :: Int,
+    knightAttackDamage  :: Int,
+    knightGoldHeld      :: Int
+  } deriving (Show)
+
+data Monster = MkMonster
+  { monsterHealth        :: Int,
+    monsterAttackDamage  :: Int,
+    monsterGoldHeld      :: Int
+  } deriving (Show)  
+
+fight :: Monster -> Knight -> Int
+fight monster knight
+  | kDmg >= mHealth = kGold + mGold
+  | mDmg >= kHealth = -1
+  | otherwise = kGold
+  where
+    kDmg = knightAttackDamage knight
+    mDmg = monsterAttackDamage monster
+    kHealth = knightHealth knight
+    mHealth = monsterHealth monster
+    kGold = knightGoldHeld knight
+    mGold = monsterGoldHeld monster
 
 {- |
 =🛡= Sum types
@@ -459,6 +488,10 @@ and provide more flexibility when working with data types.
 Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
+data Breakfast = Milk
+               | Eggs
+               | Bananas
+               | AvocadoToast
 
 {- |
 =⚔️= Task 4
@@ -479,6 +512,35 @@ After defining the city, implement the following functions:
    complicated task, walls can be built only if the city has a castle
    and at least 10 living __people__ inside in all houses of the city in total.
 -}
+data ChurchOrLibrary = Church | Library
+newtype House = House Int
+type Castle = Maybe String
+data Walls = Walls
+
+data City =  MkCity
+  { cityCastle          :: Castle,
+    cityHouses          :: [House],
+    cityCoL             :: ChurchOrLibrary,
+    cityWalls           :: Maybe Walls
+  }
+
+buildCastle :: City -> String -> City
+buildCastle city castleName =  city { cityCastle = Just castleName }
+
+buildHouse :: City -> House -> City
+buildHouse city@(MkCity _ houses _ _) newHouse@(House numPeople)
+  | numPeople > 0 && numPeople <= 4 = city { cityHouses = newHouse:houses }
+  | otherwise  = error "Number of people must be b/w 1-4."
+
+getHousePeople :: House -> Int
+getHousePeople (House x) = x;
+
+buildWalls :: City -> Walls -> City
+buildWalls (MkCity Nothing _ _ _) _ = error "No castle! Can't build walls."
+buildWalls city@(MkCity _ houses _ _) walls
+  | numPeople >= 10 = city{cityWalls = Just walls}
+  | otherwise        = error "<10 people in the city! Can't build walls."
+  where numPeople = sum (map getHousePeople houses)
 
 {-
 =🛡= Newtypes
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -479,6 +512,35 @@ After defining the city, implement the following functions:
...
+buildHouse :: City -> House -> City
+buildHouse city@(MkCity _ houses _ _) newHouse@(House numPeople)
+  | numPeople > 0 && numPeople <= 4 = city { cityHouses = newHouse:houses }
+  | otherwise  = error "Number of people must be b/w 1-4."
+
+getHousePeople :: House -> Int
+getHousePeople (House x) = x;
+
+buildWalls :: City -> Walls -> City
+buildWalls (MkCity Nothing _ _ _) _ = error "No castle! Can't build walls."
```
chshersh, line 539: Usually, it is not a good idea to use `error`, because you can't easily catch such errors. In such situations you have multiple choices:

* Restrict input to disallow invalid states
* Don't change input on invalid arguments (i.e. return `City` without changes)
* Return a new custom data type and return it as a result, e.g.:
```haskell
data BuildWallsResult
    = NoCastle
    | NotEnoughPeople Int
    | NewCity City
```

src/Chapter3.hs
```diff
@@ -560,22 +622,30 @@ introducing extra newtypes.
...
 
-calculatePlayerDefense :: Int -> Int -> Int
-calculatePlayerDefense armor dexterity = armor * dexterity
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense (Armor armor) (Dexterity dexterity) = Defense (armor * dexterity)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit (Damage damage) (Defense defense) (Health health) = Health (health + defense - damage)
```
chshersh, line 648: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -903,13 +986,38 @@ We will call such a typeclass "Append". You can find its definition below.
 Implement instances of "Append" for the following types:
 
   ✧ The "Gold" newtype where append is the addition
-  ✧ "List" where append is list concatenation
+  ✧ "List" where append is list concatenation(assuming not [a])
```
chshersh, line 989: In this task, it would be enough to use `[a]`. But practising with the custom `Cons` list is a good exercise 🙂 You will more with such lists in Chapter 4.

src/Chapter3.hs
```diff
@@ -903,13 +986,38 @@ We will call such a typeclass "Append". You can find its definition below.
...
   ✧ *(Challenge): "Maybe" where append is appending of values inside "Just" constructors
 
 -}
 class Append a where
     append :: a -> a -> a
 
+newtype Gold = Gold Int
+getGold :: Gold -> Int
+getGold (Gold x) = x
```
chshersh, line 998: You can define this getter as a `newtype` record field
```suggestion
newtype Gold = Gold
    { unGold :: Int
    }
```

src/Chapter3.hs
```diff
@@ -903,13 +986,38 @@ We will call such a typeclass "Append". You can find its definition below.
...
+            y_i = getGold y
+
+instance Append a => Append (List a) where
+  append :: List a -> List a -> List a
+  append Empty ys = ys
+  append (Cons x xs) ys =  Cons x (append xs ys)
+
+-- Why can't I "append :: Maybe a -> Maybe a -> a", where (Nothing, Nothing) gives error("nullptr")?
+instance Append a => Append (Maybe a) where
+  append :: Maybe a -> Maybe a -> Maybe a
```
chshersh, line 1017: The type signature of the `Append` typeclass forces to have the resulting type the same as the input type.

```haskell
append :: a -> a -> a
```

So, when you need to return `Maybe`, it looks reasonable `Nothing` on `(Nothing, Nothing)`. Generally, errors are discouraged, and it's usually better to avoid them when possible.

If you want to learn more about the `Append` typeclass, you can look at the `Semigroup` typeclass from the standard library, which has the same meaning:

* http://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Semigroup.html#t:Semigroup

src/Chapter3.hs
```diff
@@ -970,6 +1078,23 @@ implement the following functions:
 
 🕯 HINT: to implement this task, derive some standard typeclasses
 -}
+data DayOfWeek = Mon | Tue | Wed | Thu | Fri | Sat | Sun deriving (Show, Eq, Ord, Enum)
+
+isWeekend :: DayOfWeek -> Bool 
+isWeekend = (>=Sat)
```
chshersh, line 1084: That is right!

```suggestion
    | day == maxBound = minBound
```

src/Chapter3.hs
```diff
@@ -970,6 +1078,23 @@ implement the following functions:
...
+
+nextDay :: DayOfWeek -> DayOfWeek
+nextDay day
+  | day >= Fri = Mon
+  | otherwise = succ day
```
chshersh, line 1089: This function should return just the next day, not the "next working day". Otherwise, your implementation would be totally correct 👍🏻

chshersh (approved):
Hey @saatvikshah, great work on Chapter Three 👏🏻 
This chapter is indeed more difficult than the previous ones (not intentionally, though) because the type system in Haskell is very different from other languages, and requires thinking in a different way.

Your solutions look great! And you nailed the advanced task 🔨 

The boss task has multiple ways to be solved. Your alternative solution without using the typeclasses indeed looks simpler. But the idea behind using typeclasses is to have more extensible and general solution.

For example, when using typeclasses, you can implement the following function:

```haskell
data FightResult first second
    = FirstWins first
    | SecondWins second
    | Tie

fight :: (Fighter first, Fighter second) => first -> second -> FightResult first second
```

Your current solution doesn't implement this polymorphic function. Having such function will allow to fight `Knight` and `Monster` in any arbitrary combination.

You correctly implemented separate data types for `Monster` and `Knight` actions. You can go one step further by storing a sequence of available actions inside the `Monster` and `Knight` types. This will eliminate the `Either ...` argument in the `applyAction` method of the `Fighter` typeclass. And, moreover, it will easily allow you introduce new types of fighters and reuse the same polymorphic `fight` function.

src/Chapter3.hs
```diff
@@ -371,8 +377,31 @@ after the fight. The battle has the following possible outcomes:
...
+  | kDmg >= mHealth = kGold + mGold
+  | mDmg >= kHealth = -1
+  | otherwise = kGold
+  where
+    kDmg = knightAttackDamage knight
+    mDmg = monsterAttackDamage monster
+    kHealth = knightHealth knight
+    mHealth = monsterHealth monster
+    kGold = knightGoldHeld knight
+    mGold = monsterGoldHeld monster
```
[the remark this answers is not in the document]
saatvikshah: Wow, love this feature. I think `{-# OPTIONS_GHC -Werror=missing-fields #-}` is a much better solution than adding `!` to every record field type - don't you think? (IIUC one can go one way or the other?)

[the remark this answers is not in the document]
chshersh: @saatvikshah it is considered a best-practice to use strict fields in data types anyway, so not a problem generally 🙂 You can even go one step further and enable `{-# LANGUAGE StrictData #-}` in modules to avoid using `!` before each type. There are very few cases where you really want to have lazy fields, so such cases should be thought through and can be marked explicitly.

Anyway, on CI you usually want to enable `-Werror` for all warnings to avoid having all warnings in your code at all, so no need to specify `-Werror` for this particular warning

chshersh:

PR #393, opened 2021-03-05
Learning Haskell: Chapter 1 

zachyking:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -72,6 +72,7 @@ the `.hs` extension.
 -}
 module Chapter1 where
 
+import Data.Char
 {- |
 In Haskell, we have __expressions__. Expressions can be represented by some
 primitive values (numbers: 1, 100; characters: 'a', 'z'; booleans: True, False;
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -72,6 +72,7 @@ the `.hs` extension.
 -}
 module Chapter1 where
 
+import Data.Char
```
chshersh, line 75: Nice work on using imports! We explain them only in the next chapter, but it's good that you're already one step ahead 🦶🏻

src/Chapter1.hs
```diff
@@ -490,7 +491,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = n `mod` 10
```
chshersh, line 495: Your implementation is almost correct :ok: 

src/Chapter1.hs
```diff
@@ -520,7 +522,12 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y
+    | x == 0 || y == 0 = 0
+    | x > 0 && y > 0 = min x y
+    | x > 0 && y < 0 = if x < (y * (-1)) then x else y
+    | y > 0 = if (x * (-1)) < y then x else y
+    | otherwise = if x > y then x else y
```
chshersh, line 530: This implementation looks almost correct, but it's a bit more complicated than it could be. I recommend using the [abs](http://hackage.haskell.org/package/base-4.14.1.0/docs/GHC-Num.html#v:abs) function to simplify the logic here

src/Chapter1.hs
```diff
@@ -553,8 +560,11 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z = do
+    let maxN = foldr1 max [x, y, z]
+    let minN = foldr1 min [x, y, z]
+    [n | n <- [x,y,z], n < maxN, n > minN]!!0
```
chshersh, line 567: Interesting solution 🤔 Nice that you have an understanding of folding and typeclasses. We will cover them later

src/Chapter1.hs
```diff
@@ -568,7 +578,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c = elem (toLower c) ['a', 'e', 'i', 'o', 'u', 'y']
```
chshersh, line 582: Absolutely correct! 👍🏼 

src/Chapter1.hs
```diff
@@ -631,8 +642,22 @@ Implement a function that returns the sum of the last two digits of a number.
...
+-- getLast2 y =
+--     let lastN = y `mod` 10
+--         secondLastN = y `mod` 100
+--     in(lastN, secondLastN)
+
+
+sumLast2:: Int -> Int
+sumLast2 n  
+    | n == 0 = 0
+    | n < 10 = n
```
chshersh, line 655: Nice implementation! But this guard will return the negative number itself on the negative input

chshersh (approved):
Nice work on Chapter One tasks 👍🏻 

I left my review with some feedback 🙂

zachyking (commented):
Thanks for comments, I really appreciate it

chshersh:
Hey @zachyking! Great improvements to the Chapter One solutions 💯 
I'm going to close this PR as we can't merge solutions to our repo, but feel free to open PRs with solutions for future chapters 🏁

PR #396, opened 2021-03-13
Chapter2 solutions

zachyking:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: String
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create a list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
@@ -186,31 +186,43 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
+[10,2,3,1,5]
 
 >>> [] ++ [1, 4]  -- [] is an empty list
+[1,4]
 
 >>> 3 : [1, 2]
+[3,1,2]
 
 >>> 4 : 2 : [5, 10]  -- prepend multiple elements
+[4,2,5,10]
 
 >>> [1 .. 10]  -- list ranges
+[1,2,3,4,5,6,7,8,9,10]
 
 >>> [10 .. 1]
+[]
 
 >>> [10, 9 .. 1]  -- backwards list with explicit step
+[10,9,8,7,6,5,4,3,2,1]
 
 >>> length [4, 10, 5]  -- list length
+3
 
 >>> replicate 5 True
+[True,True,True,True,True]
 
 >>> take 5 "Hello, World!"
+"Hello"
 
 >>> drop 5 "Hello, World!"
+", World!"
 
 >>> zip "abc" [1, 2, 3]  -- convert two lists to a single list of pairs
+[('a',1),('b',2),('c',3)]
 
 >>> words "Hello   Haskell     World!"  -- split the string into the list of words
-
+["Hello","Haskell","World!"]
 
 
 👩‍🔬 Haskell has a lot of syntax sugar. In the case with lists, any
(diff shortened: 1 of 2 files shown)


src/Chapter2.hs
```diff
@@ -336,8 +348,11 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
-
+subList x y z = if (x < 0 || y < 1 || null z) then [] else if cutTailN y z > 0 then drop x ((iterate init z) !! cutTailN y z) else if x >= 0 then drop x z else []
```
vrom911, line 351: Wow, this is some advanced logic 😮 

I think it could be simplified. You can use `drop` and `take` functions to achieve the same. Also to check the border cases checks you can use guards 🙂 
Let me know if I need to elaborate on that.

src/Chapter2.hs
```diff
@@ -348,8 +363,8 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf:: [a] -> [a]
+firstHalf l = (iterate init l)!! (length l - ((length l) `div` 2))
```
vrom911, line 367: Here as well you can use the `take` function instead of `iterate` 🙂 That is also will be more performant.

src/Chapter2.hs
```diff
@@ -621,8 +640,10 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
-
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven (x:xs:xss) = x : takeEven xss
+takeEven (x:xs) = [x]
```
vrom911, line 646: In here `xs` is only could be `[]` as all other cases are covered above 🙂 So you can right it so:
```suggestion
takeEven [x] = [x] 
```

src/Chapter2.hs
```diff
@@ -621,8 +640,10 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
-
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven (x:xs:xss) = x : takeEven xss
```
vrom911, line 645: Great pattern-matching skills 👏🏼 

Note, that you can use wildcard here, as you don't really use `xs`. Besides, `xs` is usually used for naming lists, but in this case, it is just an element 🙂 

```suggestion
takeEven (x:_:xs) = x : takeEven xs
```

src/Chapter2.hs
```diff
@@ -728,7 +749,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concat(map (\x -> replicate (abs x) x) l)
```
vrom911, line 752: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -741,7 +762,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains n l = map (\x -> if n `elem` x then x else []) l
```
vrom911, line 766: In this case, you would end up with a bunch of empty lists in the result (which were not present in the initial input).
What you can do instead is to use `filter` function, to actually just leave the "right" lists 🙂

src/Chapter2.hs
```diff
@@ -781,13 +803,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x) 
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 814: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -858,7 +885,10 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = (rewind xs) ++ [x]
```
vrom911, line 890: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Awesome work on the Chapter 💪🏼 👏🏼 🎉

src/Chapter2.hs
zachyking: I was pretty sure there is something much simpler, thanks you! 🙂 I believe I managed to implement all comments and I really appreciate it.

src/Chapter2.hs
```diff
@@ -364,7 +371,12 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 firstHalf:: [a] -> [a]
-firstHalf l = (iterate init l)!! (length l - ((length l) `div` 2))
+firstHalf [] = []
+firstHalf l =
+    let len = length l
+    in take (len - len `div` 2) l
```
vrom911, line 377: You can just have `len div 2`  and it will be enough 🙂 
```suggestion
    in take (len `div` 2) l
```

chshersh:

PR #397, opened 2021-03-16
Chapter 1 solution

syukronrm:
### Solutions for Chapter 1

Hi, very grad I found this repository.

cc @vrom911 @chshersh

Edit: I forgot to run the test locally.
Edit 2: Done!

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -489,8 +490,8 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = abs n `mod` 10
```
vrom911, line 494: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1:

src/Chapter1.hs
```diff
@@ -568,7 +574,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+  | c `elem` "aiueo" = True
+  | otherwise = False
```
vrom911, line 580: Nice one! 👍🏼 

```haskell
isVowel c = c `elem` "aiueo"
```

src/Chapter1.hs
```diff
@@ -631,8 +640,12 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+  let absN = abs n
+      last1 = absN `mod` 10
+      last2 = (absN `mod` 100) `div` 10
```
vrom911, line 647: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

One hint to make your solution even shorter: you can see that you use both `mod` and `div` of ten. 

```haskell
(x, y) = divMod absN 10
```

src/Chapter1.hs
```diff
@@ -652,9 +665,13 @@ Implement a function that returns the first digit of a given number.
...
-
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit n = firstDigitAbs (abs n)
+  where
+    firstDigitAbs :: Int -> Int
+    firstDigitAbs absN
+      | absN `div` 10 == 0 = absN
+      | otherwise = firstDigit (absN `div` 10)
```
vrom911, line 674: This is absolutely great idea to have the additional recursive function 🧠 
So you can just check on `abs` only in the beginning 👍🏼 
I guess, you meant this here:

```suggestion
      | otherwise = firstDigitAbs (absN `div` 10)
```

vrom911 (approved):
Absolutely great job on the Chapter! 🎉 👏🏼

src/Chapter1.hs
syukronrm: Great! I didn't realize that.

src/Chapter1.hs
syukronrm: I haven't found that kind of function in the other language stdlib. Somehow reading the Prelude documentation is pretty intimidating for me, need to practice more.

syukronrm:
Appreciate your thorough responses @vrom911!

src/Chapter1.hs
vrom911: Don't worry about that! The Prelude docs are not in the best shape at the moment and currently, the work is going on in that direction, so I hope it would be much more approachable in the future. But for now, don't hesitate to ask about anything here 🙂

chshersh:

PR #398, opened 2021-03-18
Solutions for Chapter 1

polemius:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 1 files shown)


polemius:
Looks like here is a problem on CI. Because on my local machine all test for chapter 1 finished successfully.
Here is an error from the CI:
```
doctest-chapter1: /usr/local/.ghcup/ghc/8.10.4/bin/ghc-8.10.4: getPermissions:getFileStatus: does not exist (No such file or directory)
```

src/Chapter1.hs
```diff
@@ -520,7 +521,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if (abs x) - (abs y) < 0 then x else y
```
chshersh, line 524: `()` are redundant here. I see how it can be confusing to see when to use them and when not to use, especially in the beginning. But it's usually better to remove redundant brackets for cleaner code :broom: 

Also, I think comparison of `abs`olute values would work better here instead of subtraction. There are could be edge cases where this doesn't work. And generally, looks a little bit shorter 🙂 
```suggestion
closestToZero x y = if abs x < abs y then x else y
```

src/Chapter1.hs
```diff
@@ -554,7 +555,14 @@ value after "=" where the condition is true.
...
 
-mid x y z = error "mid: not implemented!"
+firstBetween :: Int -> Int -> Int -> Bool
+firstBetween x y z = (x <= y && x >= z) || (x <= z && x >= y)
+
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | firstBetween x y z = x
+  | firstBetween y x z = y
+  | otherwise = z
```
chshersh, line 565: Nice solution with a helper function! 👏🏻

src/Chapter1.hs
```diff
@@ -568,7 +576,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c = elem c "aeiouy"
```
chshersh, line 580: Absolutely correct! 👍🏼 

src/Chapter1.hs
```diff
@@ -632,7 +641,13 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = last1 (abs n) + last2 (abs n)
+  where
+    last1 :: Int -> Int
+    last1 x = mod x 10
+    last2 :: Int -> Int
+    last2 x = div (mod x 100 - last1 x) 10
```
chshersh, line 650: You don't need to subtract `last1 x` from `mod x 100`, because `div foo 10` removes the last digit anyway

chshersh, line 650: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (abs n) 10
div (abs n) 10
```

```haskell
```

src/Chapter1.hs
```diff
@@ -653,8 +668,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit n = 
+  if x < 10 then x else firstDigit (div (x - mod x 10) 10)
+  where x = abs n
```
chshersh, line 674: Nice usage of local variables inside `where`! 🔝

chshersh (approved):
Great work on Chapter One tasks! 👏🏻 
And you nailed the advanced task 👍🏻 

I see that CI has problems because of the following issue in the Haskell setup action:

* https://github.com/haskell/actions/issues/30

I guess, we should wait for this issue to be resolved upstream before CI starts working again. But I hope it won't stop you from continuing the course, as you can run the tests locally 💻

src/Chapter1.hs
polemius: Cool, thanks! For sure your suggested solution is better, but out of curiosity trying to find _edge cases where this doesn't work_ for my solution and could not came up with even one 🤔 Could you provide some examples?

src/Chapter1.hs
polemius: Thanks ☺️

src/Chapter1.hs
polemius: So the explicit pattern matching would be something like this?

```haskell
isVowel :: Char -> Bool
isVowel c
  | 'a' = True
  | 'e' = True
  ... all vowels
  | otherwise = False
```

src/Chapter1.hs
polemius: Thank you for suggestion! 
With `divMod` the solution will be:

```haskell
sumLast2 :: Int -> Int
sumLast2 n = fst x + snd x
  where
    x = divMod (mod (abs n) 100) 10
```

The `mod (abs n) 100` drops everything before the last 2 digits.
And `divMod ... 10` return the pair with of two last digits.

polemius:
As I complete all tasks for Chapter 1 I am closing this PR

src/Chapter1.hs
chshersh: I was thinking about the lowest and greatest `Int` values but my quick local testing wasn't able to find a case where this function breaks, so it seems actually working 👍🏻

src/Chapter1.hs
chshersh: Yes, exactly!

src/Chapter1.hs
chshersh: That's a great solution! 👏🏻 

You can even use pattern-matching on the `x` variable but that's optional of course 🙂 

```haskell
sumLast2 :: Int -> Int
sumLast2 n = a + b
  where
    (a, b) = divMod (mod (abs n) 100) 10

PR #399, opened 2021-03-20
Solutions for Chapter 2

polemius:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,44 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
+(++) :: [a] -> [a] -> [a]
 
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create a list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
@@ -186,31 +187,43 @@ Evaluate the following expressions in GHCi and insert the answers. Try
 to guess first, what you will see.
 
 >>> [10, 2] ++ [3, 1, 5]
+[10,2,3,1,5]
 
 >>> [] ++ [1, 4]  -- [] is an empty list
+[1,4]
 
 >>> 3 : [1, 2]
+[3,1,2]
 
 >>> 4 : 2 : [5, 10]  -- prepend multiple elements
+[4,2,5,10]
 
 >>> [1 .. 10]  -- list ranges
+[1,2,3,4,5,6,7,8,9,10]
 
 >>> [10 .. 1]
+[]
 
 >>> [10, 9 .. 1]  -- backwards list with explicit step
+[10,9,8,7,6,5,4,3,2,1]
 
 >>> length [4, 10, 5]  -- list length
+3
 
 >>> replicate 5 True
+[True,True,True,True,True]
 
 >>> take 5 "Hello, World!"
+"Hello"
 
 >>> drop 5 "Hello, World!"
+", World!"
 
 >>> zip "abc" [1, 2, 3]  -- convert two lists to a single list of pairs
+[('a',1),('b',2),('c',3)]
 
 >>> words "Hello   Haskell     World!"  -- split the string into the list of words
-
+["Hello","Haskell","World!"]
 
 
 👩‍🔬 Haskell has a lot of syntax sugar. In the case with lists, any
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -336,7 +349,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList start end list
+  | start < 0 || end < 0 = []
```
chshersh, line 353: Your implementation is correct when the right bound is lower than the left. But I think you can implement it more efficiently by adding a separate case to your check :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -349,7 +364,9 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 -- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf l = take halfLenght l
+  where
+    halfLenght = div (length l) 2
```
chshersh, line 369: Great solution! Could you also add a type signature to this function? 🙂

src/Chapter2.hs
```diff
@@ -501,8 +518,23 @@ True
...
+    isThird42 (_:_:42) = True
+    @
+  I am getting an error:
+
+  src/Chapter2.hs:522:16: error:
+    • No instance for (Num [Int]) arising from the literal ‘0’
+    • In the pattern: 0
+      In the pattern: _ : 0
+      In the pattern: _ : _ : 0
+-}
```
chshersh, line 537: When you use pattern `(x : xs)`, the `x` variable has the type of the list element, and `xs` has the type of list. So when you writ e `(x : 42)` the compiler tries to match types of the number 42 and the list and fails (because number type is not the same as the list type)

src/Chapter2.hs
```diff
@@ -621,7 +654,10 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven [x] = [x]
+takeEven (e:_:xs) = e : takeEven xs
```
chshersh, line 660: Great pattern matching skills! 👏🏻

src/Chapter2.hs
```diff
@@ -728,7 +764,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concat (map (\a -> replicate a a) l)
```
chshersh, line 767: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -741,7 +777,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains number lists = filter (\list -> elem number list) lists
```
chshersh, line 781: Now, after you've mastered eta-reduction, you can apply such a technique to this function as well  🙂 
Here you have two opportunities for eta-reduction: the last argument to `filter` and the lambda

src/Chapter2.hs
```diff
@@ -842,7 +880,11 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate n l
+  | length l == 0 = []
```
chshersh, line 885: `length` is slow for lists. So, if you want to efficiently check list on emptiness, it's better to use the standard function `null` or pattern matching

src/Chapter2.hs
```diff
@@ -842,7 +880,11 @@ list.
...
+  | n < 0 = []
+  | otherwise = drop n (take (length l + n) (cycle l))
```
chshersh, line 887: That's a correct solution! Though, it does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can think of a way to optimize this :slightly_smiling_face:

src/Chapter2.hs
```diff
@@ -858,8 +900,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
-
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = (rewind xs) ++ [x]
```
chshersh, line 905: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

chshersh (approved):
Wonderful work on Chapter 3 tasks! 💎

src/Chapter2.hs
polemius: Thank you for explanation!

src/Chapter2.hs
polemius: Thanks ☺️

src/Chapter2.hs
polemius: Yes, sure. I forgot about it 🤦‍♂️

src/Chapter2.hs
polemius: Thank you for suggestion. Now function is shorter. 👍

src/Chapter2.hs
polemius: Thank you for suggestion! I have rewritten this function with recursive go patter 👍

src/Chapter2.hs
polemius: Cool! Thank you for suggestion. I found that rotating array with 6 element by 34 elements is the same that rotating by 4 elements. I have rewritten this function

chshersh:
@polemius The new changes look good! I'm going to close this PR, as we can't merge solutions to our repo 🏁

PR #400, opened 2021-03-22
Chapter 2 solutions

syukronrm:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,42 +136,52 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
+[True, False] :: [Bool]
 
 
 String is a list of characters:
 >>> :t "some string"
+"some string" :: [Char]
 
 
 Empty list:
 >>> :t []
+[] :: [a]
 
 
 Append two lists:
 >>> :t (++)
+(++) :: [a] -> [a] -> [a]
 
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
+(:) :: a -> [a] -> [a]
 
 
 Reverse a list:
 >>> :t reverse
+reverse :: [a] -> [a]
 
 
 Take first N elements of a list:
 >>> :t take
+take :: Int -> [a] -> [a]
 
 
 Create a list from N same elements:
 >>> :t replicate
+replicate :: Int -> a -> [a]
 
 
 Split a string by line breaks:
 >>> :t lines
+lines :: String -> [String]
 
 
 Join a list of strings with line breaks:
 >>> :t unlines
+unlines :: [String] -> String
 
 
 -}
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -348,8 +375,10 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf l = take len l
+  where
+    len = length l `div` 2
```
vrom911, line 381: Nice usage of the `where` construction! 👏🏼  Makes code very easily readable 👍🏼

src/Chapter2.hs
```diff
@@ -605,9 +636,18 @@ Implement a function that duplicates each element of the list
...
+-- Solution 1
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate l = go [] (reverse l)
+  where
+    go :: [a] -> [a] -> [a]
+    go result [] = result
+    go result (x:xs) = go (x : x : result) xs
 
+-- Solution 2, took some time to think about this solution
```
vrom911, line 647: Nice one! 

This solution has a number of advantages compared to the previous one: it is more efficient and it is also streaming (the previous one won't work on infinite lists).
Well done 👍🏼

src/Chapter2.hs
```diff
@@ -621,7 +661,10 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven (x:_:xs) = x : takeEven xs
```
vrom911, line 666: Great pattern matching skills 💯

src/Chapter2.hs
```diff
@@ -727,8 +770,10 @@ value of the element itself
 
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
+-- I don't see where to use `map`
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate (x:xs) = (replicate x x) ++ (smartReplicate xs)
```
vrom911, line 776: This is right! 💯 

```haskell
[3,1,2,4,0,1]
```

src/Chapter2.hs
```diff
@@ -781,13 +827,14 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
--- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 837: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -842,7 +889,11 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+
+rotate :: Int -> [a] -> [a]
+rotate n l
+  | n < 0 = []
+  | otherwise = take (length l) (drop n (cycle l))
```
vrom911, line 896: That's the correct solution! 🌞 

Though, there is a way to improve it a bit 💠  
It does a little bit more than work when the given number is big and the list is small. You will cycle through the list a lot. Maybe you can optimize this using `mod`, for example :slightly_smiling_face:

vrom911 (approved):
Amazing!
And fantastic work on the advanced task as well 👏🏼

chshersh:

PR #401, opened 2021-03-23
Chapter3

zachyking:
### Solutions for Chapter 3

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -49,10 +49,10 @@ In this module, we enable the "InstanceSigs" feature that allows writing type
 signatures in places where you can't by default. We believe it's helpful to
 provide more top-level type signatures, especially when learning Haskell.
 -}
-{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE InstanceSigs, MultiParamTypeClasses #-}
 
 module Chapter3 where
-
+import Data.Maybe
 {-
 =🛡= Types in Haskell
 
@@ -344,6 +344,21 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+data Book = MkBook
+    { bookPublishYear   :: Int
+    , bookTitle         :: String
+    , bookAuthor        :: String
+    } deriving (Show)
+
+
+-- bible :: Book
+-- bible = MkBook
+--     { bookPublishYear = 0
+--     , bookTitle = "Bible"
+--     , bookAuthor = "Unknown"
+--     } 
+
+
 {- |
 =⚔️= Task 2
 
@@ -374,6 +389,46 @@ after the fight. The battle has the following possible outcomes:
 
 -}
 
+data Fighter = MkFighter
+    { fighterHealthPts    :: Int,
+      fighterAttackPts    :: Int,
+      fighterGold         :: Int
+    } deriving (Show)
+
+knight :: Fighter
+knight = MkFighter
+    { fighterHealthPts = 10,
+      fighterAttackPts = 2,
+      fighterGold = 5
+    }
+
+monster :: Fighter
+monster = MkFighter
+    { fighterHealthPts = 15,
+      fighterAttackPts = 1,
+      fighterGold = 11
+    }
+
+
+fight :: Fighter -> Fighter -> Int
+fight kn mn 
+    | fighterHealthPts kn < 1 = if fighterHealthPts mn < 1 then fighterGold kn else -1
+    | fighterHealthPts mn < 1 = (fighterGold kn) + (fighterGold mn)
+    | fighterAttackPts kn < 1 = if fighterAttackPts mn < 1 then fighterGold kn else -1
+    | fighterAttackPts mn < 1 = (fighterGold kn) + (fighterGold mn)
+    | otherwise = if knToDeathRounds kn mn >= mnToDeathRounds kn mn then (fighterGold kn) + (fighterGold mn) else -1
+    where 
+        knToDeathRounds :: Fighter -> Fighter -> Int
+        knToDeathRounds kn mn = if (fighterHealthPts kn) `rem` (fighterAttackPts mn) > 0 then (fighterHealthPts kn) `div` (fighterAttackPts mn) + 1 else (fighterHealthPts kn) `div` (fighterAttackPts mn)
+        mnToDeathRounds :: Fighter -> Fighter -> Int
+        mnToDeathRounds kn mn = if (fighterHealthPts mn) `rem` (fighterAttackPts kn) > 0 then (fighterHealthPts mn) `div` (fighterAttackPts kn) + 1 else (fighterHealthPts mn) `div` (fighterAttackPts kn) 
+
+    -- let mnHealthLeft = fighterHealthPts mn - fighterAttackPts kn
+    -- in if mnHealthLeft > 0 then 
+    --     let knHealthLeft = fighterHealthPts kn - fighterAttackPts mn
+    --     in if knHealthLeft > 0 then fight (kn { fighterHealthPts = knHealthLeft }) (mn { fighterHealthPts = mnHealthLeft }) else -1
+    --     else (fighterGold kn) + (fighterGold mn)
+           
 {- |
 =🛡= Sum types
 
@@ -460,6 +515,17 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data Meal
+    = Breakfast String
+    | Brunch String
+    | Snack String
+    | Lunch String
+    | TeaTime String
+    | Supper String
+    | Appetizer String
+    | MidnightSnack String
+    | Feast String
+
 {- |
 =⚔️= Task 4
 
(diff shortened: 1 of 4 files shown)


src/Chapter3.hs
```diff
@@ -374,6 +389,46 @@ after the fight. The battle has the following possible outcomes:
...
+
+monster :: Fighter
+monster = MkFighter
+    { fighterHealthPts = 15,
+      fighterAttackPts = 1,
+      fighterGold = 11
+    }
+
+
+fight :: Fighter -> Fighter -> Int
```
vrom911, line 413: I see that you use the same type for `Monster` and `Knight`. This is a nice catch, because these two entities have the same structure and characteristics.
The only issue with reusage of the same data type for Monster and Knight in here is that you can easily mix them when calling the function:

```haskell
> fight knight monster
> fight monster knight
```

```haskell
```

src/Chapter3.hs
```diff
@@ -374,6 +389,46 @@ after the fight. The battle has the following possible outcomes:
...
+fight kn mn 
+    | fighterHealthPts kn < 1 = if fighterHealthPts mn < 1 then fighterGold kn else -1
+    | fighterHealthPts mn < 1 = (fighterGold kn) + (fighterGold mn)
+    | fighterAttackPts kn < 1 = if fighterAttackPts mn < 1 then fighterGold kn else -1
+    | fighterAttackPts mn < 1 = (fighterGold kn) + (fighterGold mn)
+    | otherwise = if knToDeathRounds kn mn >= mnToDeathRounds kn mn then (fighterGold kn) + (fighterGold mn) else -1
+    where 
+        knToDeathRounds :: Fighter -> Fighter -> Int
+        knToDeathRounds kn mn = if (fighterHealthPts kn) `rem` (fighterAttackPts mn) > 0 then (fighterHealthPts kn) `div` (fighterAttackPts mn) + 1 else (fighterHealthPts kn) `div` (fighterAttackPts mn)
```
vrom911, line 422: I think this function is a bit overcomplicated 🙂 
What you actually need to check is that the health of one fighter is bigger than the attack of another fighter. So you can make two checks to see if that the case and act accordingly 👌🏼

src/Chapter3.hs
```diff
@@ -460,6 +515,17 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
...
+data Meal
+    = Breakfast String
+    | Brunch String
+    | Snack String
+    | Lunch String
+    | TeaTime String
+    | Supper String
+    | Appetizer String
+    | MidnightSnack String
+    | Feast String
```
vrom911, line 527: Yummy 😋

src/Chapter3.hs
```diff
@@ -480,6 +546,81 @@ After defining the city, implement the following functions:
    and at least 10 living __people__ inside in all houses of the city in total.
 -}
 
+data Person 
+    = Person
```
vrom911, line 550: This data type actually doesn't bring any weight to your system. Do you think that you can redesign the solution a bit without empty data types?

src/Chapter3.hs
```diff
@@ -480,6 +546,81 @@ After defining the city, implement the following functions:
    and at least 10 living __people__ inside in all houses of the city in total.
 -}
 
+data Person 
+    = Person 
+    deriving (Show)
+
+data House = House 
+    { houseResidents :: [Person]
```
vrom911, line 554: Interesting representation. However, this doesn't guarantee that the instance of this data type will be correct. Your data type allows the house to be without a person, or with 1000 persons. Also, the `Person` data type itself is redundant as it is just an empty type. You can notice that it is more convenient to have a notion of "Number of Persons" in this type. And to restrict the number of persons, you can create 4 constructors for each option of houses.

src/Chapter3.hs
```diff
@@ -480,6 +546,81 @@ After defining the city, implement the following functions:
...
+    } deriving (Show)
+
+data CityBuilding 
+    = Church 
+    | Library 
+    deriving (Show)
+
+data Castle = Castle
+    { castleName    :: String,
+      cityWalls     :: Int
```
vrom911, line 564: Nice data type!
I think that walls could be represented by a more suitable data type, as currently it allows to have negative number of walls 🙂

src/Chapter3.hs
```diff
@@ -480,6 +546,81 @@ After defining the city, implement the following functions:
...
+
+
+data MagicalCity = MagicalCity
+    { cityBuilding    :: CityBuilding,
+      houses          :: [House],
+      cityCastle      :: Maybe Castle
+    } deriving (Show)
+
+isChurch :: CityBuilding -> Bool
+isChurch (Church) = True
```
vrom911, line 575: Parenthesis are redundant 🙂 
```suggestion
isChurch Church = True
```

src/Chapter3.hs
```diff
@@ -480,6 +546,81 @@ After defining the city, implement the following functions:
...
+isChurch _ = False 
+
+hasChurch :: MagicalCity -> Bool
+hasChurch mC = isChurch (cityBuilding mC)
+
+hasCastle :: MagicalCity -> Bool
+hasCastle mC = not (isNothing (cityCastle mC))
```
vrom911, line 582: You can use `isJust` here 👌🏼

src/Chapter3.hs
```diff
@@ -480,6 +546,81 @@ After defining the city, implement the following functions:
...
+
+wallCount :: Maybe Castle -> Int
+wallCount c
+    | not (isNothing c) = cityWalls (fromJust c)
+    | otherwise = 0
```
vrom911, line 587: It is better to pattern match explicitely on the input.
It is not a good practice to use partial functions in your code (`fromJust`, for example), even if you are sure that they are safe there. If there is a way to avoid that it is better to take it 👍🏼

src/Chapter3.hs
```diff
@@ -480,6 +546,81 @@ After defining the city, implement the following functions:
...
+    cityCastle = Just (Castle {castleName = "Magical Castle", cityWalls = 5}) 
+    }
+-- testCity = MagicalCity { cityBuilding = Church, houses = [ House { houseResidents = [Person, Person, Person, Person, Person, Person] }, House { houseResidents = [Person, Person, Person, Person, Person, Person] } ], cityCastle = Just (Castle {castleName = "Magical Castle", cityWalls = 5}) }
+-- testCity = MagicalCity { cityBuilding = Church, houses = [House [Person]], cityCastle = Nothing }
+
+buildCastle :: MagicalCity -> String -> MagicalCity
+buildCastle mC castleN = mC { cityCastle = newCastle mC castleN}
+    where
+      newCastle :: MagicalCity -> String -> Maybe Castle
+      newCastle mC castleN = if hasCastle mC then Just ((fromJust (cityCastle mC)) { castleName = castleN}) else Just (Castle {castleName = castleN, cityWalls = 0})
```
vrom911, line 608: `mC` and `castleN` are already in scope so there is no need to pass them to this function.

And also, here as well it is better to pattern match on the castle 🙂

src/Chapter3.hs
```diff
@@ -480,6 +546,81 @@ After defining the city, implement the following functions:
...
+buildCastle :: MagicalCity -> String -> MagicalCity
+buildCastle mC castleN = mC { cityCastle = newCastle mC castleN}
+    where
+      newCastle :: MagicalCity -> String -> Maybe Castle
+      newCastle mC castleN = if hasCastle mC then Just ((fromJust (cityCastle mC)) { castleName = castleN}) else Just (Castle {castleName = castleN, cityWalls = 0}) 
+
+buildHouse :: MagicalCity -> MagicalCity
+buildHouse mC =
+    let peopleNum = if hasCastle mC then if hasChurch mC then 4 else 3 else if hasChurch mC then 2 else 1
+    in mC {houses = House (replicate peopleNum Person) : houses mC}
```
vrom911, line 613: Interesting rules 🤔 
Note, that an alternative design could be having a number as an input 🙂

src/Chapter3.hs
```diff
@@ -560,22 +701,50 @@ introducing extra newtypes.
...
+calculatePlayerDefense :: Armor -> Dexterity -> Defense
+calculatePlayerDefense armor dexterity = Defense ((armorPts armor) * (dexterityPts dexterity))
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: Damage -> Defense -> Health -> Health
+calculatePlayerHit damage defense health = Health ((healthPts health) + (defensePts defense) - (damagePts damage))
 
 -- The second player hits first player and the new first player is returned
 hitPlayer :: Player -> Player -> Player
```
vrom911, line 750: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -907,9 +1108,27 @@ Implement instances of "Append" for the following types:
...
+data Gold = Gold
+  { gold  :: Int
+  } deriving (Show)
+
 class Append a where
     append :: a -> a -> a
 
+instance Append Gold where
+    append :: Gold -> Gold -> Gold
+    append x y = Gold((gold x) + (gold y))
```
vrom911, line 1120: parenthesis are redundant for the function applications 🙂 
Also, in Haskell, there are spaces between functions and arguments, even if there are parenthesis there 👾 
```suggestion
    append x y = Gold (gold x + gold y)
```

src/Chapter3.hs
```diff
@@ -907,9 +1108,27 @@ Implement instances of "Append" for the following types:
...
+
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append x y = concat [x, y]
```
vrom911, line 1124: Creative solution 💡 🎨 

But it is not the most efficient 🙂 
You can use pattern matching to make it fast :)

src/Chapter3.hs
```diff
@@ -907,9 +1108,27 @@ Implement instances of "Append" for the following types:
...
+
+instance (Append a) => Append (Maybe a) where
+    append :: Maybe a -> Maybe a -> Maybe a
+    append (Just x) (Just y) = Just (append x y)
+    append Nothing (Just y) = Just y
+    append (Just x) Nothing = Just x
+    append Nothing Nothing = Nothing
```
vrom911, line 1131: This is an absolutely correct solution!

src/Chapter3.hs
```diff
@@ -970,7 +1189,30 @@ implement the following functions:
...
+    | Thursday 
+    | Friday
+    | Saturday
+    | Sunday
+    deriving(Show, Enum, Bounded)
+
+isWeekend :: DayOfWeek -> Bool
+isWeekend (Saturday) = True
+isWeekend (Sunday) = True
+isWeekend (_) = False
```
vrom911, line 1205: No need for brackets:

```suggestion
isWeekend Saturday = True
isWeekend Sunday = True
isWeekend _ = False
```

src/Chapter3.hs
```diff
@@ -970,7 +1189,30 @@ implement the following functions:
...
+
+nextDay :: DayOfWeek -> DayOfWeek
+nextDay (Sunday) = Monday
```
vrom911, line 1208: That is right!
You could also use the instance of `Bounded` here:

```
    | day == maxBound = minBound
```

src/Chapter3.hs
```diff
@@ -970,7 +1189,30 @@ implement the following functions:
...
+nextDay d = succ d
+
+daysToParty :: DayOfWeek -> Int
+daysToParty (Friday) = 0
+daysToParty d 
+    | (fromEnum d) < (fromEnum Friday) = (fromEnum Friday) - (fromEnum d)
+    | otherwise = (abs ((fromEnum Friday) - (fromEnum d))) + (fromEnum Friday)
```
vrom911, line 1215: You can skip all of the parenthesis here.
Also, you can see that you reuse the same function call many times, so it would make sense to put it into the variables.
Moreover, instead of the manual check, you could use `div .. 7` to get the day 🙂

src/Chapter3.hs
```diff
@@ -1006,6 +1248,143 @@ Implement data types and typeclasses, describing such a battle between two
...
+    } deriving(Show)
+
+testMonster :: Monster
+testMonster = Monster { monsterHealth = HealthPoints { healthPoints = 24 }, monstersAttackPower = 1, monstersLastAttack = Nothing}
+
+testMonster2 :: Monster
+testMonster2 = Monster { monsterHealth = HealthPoints { healthPoints = 15 }, monstersAttackPower = 13, monstersLastAttack = Nothing}
+
+data Warrior
+    = WarMonster { warMonster :: Monster }
```
vrom911, line 1293: Could you please clarify, why this type is necessary? I see that it is a wrapper around the `Monster` 🤔

src/Chapter3.hs
```diff
@@ -1006,6 +1248,143 @@ Implement data types and typeclasses, describing such a battle between two
...
+    | WarKnight  { warKnight :: Knight }
+    deriving(Show)
+
+class DoAttack a b where
+    doAttack :: a -> b -> (a, b)
+    
+instance DoAttack Knight Knight where
```
vrom911, line 1300: I don't think that you should allow fights between two knights though ⚔️

src/Chapter3.hs
```diff
@@ -1006,6 +1248,143 @@ Implement data types and typeclasses, describing such a battle between two
...
+instance DoAttack Knight Monster where
+    doAttack :: Knight -> Monster -> (Knight, Monster)
+    doAttack aK dM 
+        | isNothing (knightsLastAttack aK) || (knightsLastAttack aK) == Just IncreaseDefence 
+            = (aK { knightsLastAttack = Just KnightAttacking }, dM { monsterHealth = HealthPoints { healthPoints = healthPoints (monsterHealth dM) - (knightsAttackPower aK)}})
+        | (knightsLastAttack aK) == Just KnightAttacking 
+            = (aK { knightHealth = HealthPoints { healthPoints = (healthPoints (knightHealth aK)) + 1}, knightsLastAttack = Just Heal}, dM)
+        | otherwise = (aK { knightDefence = DefencePoints { defencePoints = (defencePoints (knightDefence aK)) + 1}, knightsLastAttack = Just IncreaseDefence}, dM)
+
+instance DoAttack Monster Monster where
```
vrom911, line 1318: And this one 🧟

src/Chapter3.hs
```diff
@@ -1006,6 +1248,143 @@ Implement data types and typeclasses, describing such a battle between two
...
+    isDead k = if (getHealthNum k) > 0 then False else True
+
+instance IsDead Monster where
+    isDead :: Monster -> Bool
+    isDead m = if (getHealthNum m) > 0 then False else True
+
+data BattleOutcome = BattleOutcome
+    { winner       :: Maybe Warrior,
+      totalRounds  :: Int
+    } deriving(Show)
```
vrom911, line 1357: Nice resulting data type with additional info 💯

vrom911 (approved):
Great stuff! A lot of wonderful types you constructed here 👷🏼 🎉

chshersh:

PR #404, opened 2021-03-29
Chapter4

zachyking:
### Solutions for Chapter 4

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -114,23 +114,25 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
-
+Char :: *
 >>> :k Bool
 
 >>> :k [Int]
-
+[Int] :: *
 >>> :k []
-
+[] :: * -> *
 >>> :k (->)
-
+(->) :: * -> * -> *
 >>> :k Either
+Either :: * -> * -> *
 
 >>> data Trinity a b c = MkTrinity a b c
 >>> :k Trinity
+Trinity :: * -> * -> * -> *
 
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
-
+IntBox :: (* -> *) -> *
 -}
 
 {- |
@@ -259,7 +261,7 @@ name.
 
 > QUESTION: Can you understand why the following implementation of the
   Functor instance for Maybe doesn't compile?
-
+Needs to check for nothing 
 @
 instance Functor Maybe where
     fmap :: (a -> b) -> Maybe a -> Maybe b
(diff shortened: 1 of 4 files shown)


src/Chapter4.hs
```diff
@@ -259,7 +261,7 @@ name.
 
 > QUESTION: Can you understand why the following implementation of the
   Functor instance for Maybe doesn't compile?
-
+Needs to check for nothing
```
chshersh, line 264: That's almost correct! In the implementation, the variable `x` has type `Maybe a` but the function needs to return `Maybe b`, so you can't just return `x`, the types don't match.

src/Chapter4.hs
```diff
@@ -472,11 +482,12 @@ Implement the Applicative instance for our 'Secret' data type from before.
...
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
-
+    Trap e <*> _ = Trap e
+    Reward f <*> x = fmap f x
```
chshersh, line 489: Nice work on reusing `Functor` instance for `Secret`! :world_map:

src/Chapter4.hs
```diff
@@ -489,6 +500,20 @@ Implement the 'Applicative' instance for our 'List' type.
...
+    pure a = Cons a (Empty)
+
+    (<*>) :: List (a -> b) -> List a -> List b
+    Empty <*> _ = Empty
+    _ <*> Empty = Empty
+    Cons f a <*> c = joinLists (fmap f c) (a <*> c)
+
+joinLists :: List a -> List a -> List a
+joinLists Empty c = c
+joinLists (Cons a b) c = Cons a (joinLists b c)
```
chshersh, line 514: This is the correct implementation! You can implement it even more efficiently by pattern-matching on the second list and return the first one immediately if the second is `Empty`. So you will avoid traversing the first list if the second is empty.

Usually, such function is called `append`. In Haskell, the standard typeclass for appending things is called `Semigroup`. So, if you want to practice more, you can implement instance of `Semigroup` for `List` and use it in the `Applicative` instance 🙂 

* http://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Semigroup.html#t:Semigroup

chshersh (approved):
:tada: Fantastic work!

<hr>

If you would like to continue your journey to Haskell, check out our [Haskell Knowledge Map](https://kowainik.github.io/images/Haskell_Knowledge_Map.png) for the list of topics to explore 🗺️

chshersh:

PR #405, opened 2021-03-31
Chapter1 and 2 Done

ramirezalejo:
### Solutions for Chapter 1 and 2
Hi @chshersh, 
Thanks for your reply on my previous PR, I thought you guys were not checking anymore the repo and I closed it, but here are the changes plus the changes for chapter 2.
I have struggled a lot with the eta-reduce... it is very hard for me to understand the evaluation of it with multiple variables.
Thank you for taking the time to help others to learn, your input in this new PR will be greatly appreciated.
Regards,
Alejandro

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -640,9 +640,9 @@ specifying complex expressions.
 -}
 
 sumLast2 :: Int -> Int
-sumLast2 n =  let last = mod n 10
+sumLast2 n =  let last1 = mod n 10
                   secondLast = mod (div n 10) 10
-              in (last + secondLast)
+              in (last1 + secondLast)
 
 
 {- |
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
@@ -301,43 +301,43 @@ expressions in GHCi
   functions and operators first. Remember this from the previous task? ;)
 
 >>> 1 + 2
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+3
 
 >>> 10 - 15
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+-5
 
 >>> 10 - (-5)  -- negative constants require ()
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+15
 
 >>> (3 + 5) < 10
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> True && False
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+False
 
 >>> 10 < 20 || 20 < 5
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> 2 ^ 10  -- power
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+1024
 
 >>> not False
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> div 20 3  -- integral division
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+6
 
 >>> mod 20 3  -- integral division remainder
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+2
 
 >>> max 4 10
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+10
 
 >>> min 5 (max 1 2)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+2
 
 >>> max (min 1 10) (min 5 7)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+5
 
 Because Haskell is a __statically-typed__ language, you see an error each time
 you try to mix values of different types in situations where you are not
@@ -428,7 +428,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Int -> Int -> Int
 squareSum x y = (x + y) * (x + y)
 
 
@@ -449,7 +449,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next x = x + 1
 
 {- |
 After you've implemented the function (or even during the implementation), you
@@ -490,7 +490,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
 
 
 {- |
@@ -520,7 +521,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if x == (min (abs x) (abs y)) then x else if y == (min (abs x) (abs y)) then y else if (abs x) == (min (abs x) (abs y)) then x else y
 
 
 {- |
@@ -553,9 +554,11 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
-
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | (y > x && x > z) || (z > x && x > y) = x
+    | (x > y && y > z) || (z > y && y > x) = y
+    | otherwise = z
 {- |
 =⚔️= Task 8
 
@@ -568,7 +571,11 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool 
+isVowel c  
+  | c == 'a' || c == 'e' || c == 'i' || c == 'o' || c == 'u' = True
+  | c == 'A' || c == 'E' || c == 'I' || c == 'O' || c == 'U' = True
+  | otherwise = False
 
 
 {- |
@@ -632,7 +639,10 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =  let last = mod n 10
+                  secondLast = mod (div n 10) 10
+              in (last + secondLast)
 
 
 {- |
@@ -653,7 +663,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit (n) = if (dividend n) > 0 then (firstDigit (dividend n)) else n
+  where dividend :: Int -> Int
+        dividend x = div x 10
 
 
 {-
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +490,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
```
chshersh, line 494: Your implementation is almost correct :ok: 

src/Chapter1.hs
```diff
@@ -520,7 +521,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if x == (min (abs x) (abs y)) then x else if y == (min (abs x) (abs y)) then y else if (abs x) == (min (abs x) (abs y)) then x else y
```
chshersh, line 524: That looks like a valid implementation 👍🏻 

But I think you can implement it shorter by using the `<` or `>` comparison operators and the `abs` function

src/Chapter1.hs
```diff
@@ -568,7 +571,11 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool 
+isVowel c  
+  | c == 'a' || c == 'e' || c == 'i' || c == 'o' || c == 'u' = True
+  | c == 'A' || c == 'E' || c == 'I' || c == 'O' || c == 'U' = True
+  | otherwise = False
```
chshersh, line 578: Great implementation! To write it slightly shorter, you could use the [`toLower`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Char.html#v:toLower) function from the `Data.Char` module in `base`. But we cover imports only in the second chapter, so no action is required here for now 🙂

src/Chapter1.hs
```diff
@@ -632,7 +639,10 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =  let last1 = mod n 10
+                  secondLast = mod (div n 10) 10
+              in (last1 + secondLast)
```
chshersh, line 645: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod n 10
div n 10
```

```haskell
```

src/Chapter2.hs
```diff
@@ -335,8 +335,11 @@ from it!
 
 ghci> :l src/Chapter2.hs
 -}
+
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList x y list
+  | x >= 0 = (take (y - x + 1) . drop x) list
```

src/Chapter2.hs
```diff
@@ -501,7 +505,9 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird42 :: [Int] -> Bool
+isThird42   (_ : _ : x : _ ) = if (x == 42) then True else False
```
chshersh, line 509: You don't need to compare `z` with 42 separately, you can pattern match on `42` directly in the pattern.
```suggestion
isThird42   (_ : _ : 42 : _ ) = True
```

src/Chapter2.hs
```diff
@@ -621,7 +628,13 @@ Write a function that takes elements of a list only in even positions.
...
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven y = go 0 y
+  where
+    go :: Int -> [a] -> [a]
+    go _ [] = [] 
+    go acc (x:xs) = if (mod acc 2) == 0 then x : go (acc + 1) xs else go (acc + 1) xs
```
chshersh, line 637: Nice idea on the helper functions :+1:

src/Chapter2.hs
```diff
@@ -728,7 +741,10 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate l = concat( map (\ (x) -> (replicate x x) ) l )
```
chshersh, line 745: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -741,7 +757,13 @@ the list with only those lists that contain a passed element.
...
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains ::  Int -> [[Int]] -> [[Int]]
+contains _ [] = []
+contains i list = go i list
+  where
+    go :: Int -> [[Int]] -> [[Int]]
+    go _ [] = [] 
+    go item (x:xs) = if elem item x then x : go item xs else go item xs
```
chshersh, line 766: That's a correct implementation 👍🏻  And you correctly use `elem` to check if the element is in the list.

Here you can use the standard function `filter` to automatically filter the list by the given predicate. So you won't need to implement manual recursion.

src/Chapter2.hs
```diff
@@ -781,13 +803,16 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: Num c => [c] -> [c] -> [c]
+--pairMul xs ys = zipWith (*) xs ys
+pairMul = zipWith (*)
```
chshersh, line 815: Eta-reduction award! 🏆 

I totally understand that understanding and working with eta-reduction can be challenging at first.

I can give some more explanation on the topic.

When eta-reducing, you need to remove arguments from both sides of the `=` sign. So the correct solution is:

```haskell
divideTenBy = div 10
```

See how it compares with the original code:

```diff
-divideTenBy x = div 10 x
+divideTenBy = div 10
```

The motivation behind eta-reduction if that if you have two functions `foo` and `bar` of the same type, and they return the same results for any arguments (so `foo x` always equal to `bar x` for any `x`). In such situations, we say that functions `foo` and `bar` are equal, and if we want to define `foo` in terms of `bar`, we can write `foo = bar` instead of a longer `foo x = bar x` (which means that `foo` returns the same result as `bar` for any `x`).

I hope it helps 🙂

src/Chapter2.hs
```diff
@@ -858,7 +888,11 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind [x] = [x]
+rewind (x:xs) = (last xs) : rewind([x] ++ take(length(xs) - 1) xs)
```
chshersh, line 894: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list or take the list element of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

src/Chapter2.hs
```diff
@@ -842,7 +867,12 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [Int] -> [Int]
+rotate times list
+  | mod times (length(list)) == 0 = list
+  | times > 0 = (take (length(list)) . drop (mod times (length(list)))) (cycle list)
+  | otherwise = []
```
chshersh, line 874: The implementation looks correct, but I see a few improvements:

* Check the list on emptiness first to avoid division by zero
* Check the `times` variable on invalid inputs (zero or negative)
* Move `mod times (length list)` into a separate variable (e.g. inside the `where`  clause) to avoid computing it twice
* Reduce the number of `()` to make the code more readable

chshersh (approved):
Congratulations on finishing Chapter 1 and 2! 👏🏻 
And nice solutions to the advanced tasks 💪🏻 

I left some comments with suggestions and explanations 🧑‍🏫

src/Chapter1.hs
ramirezalejo: Got it, I tested it and saw how unexpected the result was, I added abs to avoid issues with negative numbers

src/Chapter1.hs
ramirezalejo: You are right, I had a redundant condition there, just removed

src/Chapter1.hs
ramirezalejo: Noted!

src/Chapter1.hs
ramirezalejo: I see, I rewrote it to use the divMod, I also find it easier to read with it

src/Chapter2.hs
ramirezalejo: Nice one, I hope I get to know soon the standard functions, I just have one doubt, what about the checks? I had to remove the "smartReplicate [] = []" because when eta reduce the arguments did not match (I see in this case was not necessary, but I guess maybe will be required in some other case)

src/Chapter2.hs
ramirezalejo: I did it with filter to see how it looks like, I see your point

src/Chapter2.hs
ramirezalejo: It does, Thanks a lot!

src/Chapter2.hs
ramirezalejo: Great, I can see this be more efficient and I see how I can combine the case with where, removed not necessary parenthesis

src/Chapter2.hs
ramirezalejo: I understand, I was rotating in place which is very well  explained by the example of the train, when thinking about the problem I could not see how to solve it in another way, so I "brute force it", but now when looking at the other tasks I think I got it, I try to create a new array from reading the initial one

src/Chapter2.hs
ramirezalejo: I think I got it, making the checks first makes the code more readable

src/Chapter2.hs
ramirezalejo: I see what you mean, I changed it to use the boolean and then just negate it on every iteration, makes total sense

src/Chapter2.hs
chshersh: Yeah, for eta-reducing you indeed need to remove this extra check. Fortunately, it's not a problem, since both `map` and `concatMap` work perfectly on empty lists 👍🏻

src/Chapter2.hs
```diff
@@ -630,11 +631,11 @@ Write a function that takes elements of a list only in even positions.
...
 takeEven :: [a] -> [a]
 takeEven [] = []
-takeEven y = go 0 y
+takeEven y = go True y
   where
-    go :: Int -> [a] -> [a]
+    go :: Bool -> [a] -> [a]
     go _ [] = [] 
-    go acc (x:xs) = if (mod acc 2) == 0 then x : go (acc + 1) xs else go (acc + 1) xs
+    go even (x:xs) = if even then x : go (not even) xs else go (not even) xs
```
chshersh, line 638: `even` is a standard function in Haskell (that's why you can see the warning below). It works because Haskell allows shadowing of variables locally. But generally, it's better to avoid name collisions for readability. So, the local variable could be named `isEven`

src/Chapter2.hs
```diff
@@ -741,7 +755,9 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains ::  Int -> [[Int]] -> [[Int]]
+contains _ [] = []
```
chshersh, line 759: You don't need this extra special case, because `filter` returns empty list on empty list

src/Chapter2.hs
```diff
@@ -741,7 +755,9 @@ the list with only those lists that contain a passed element.
...
+contains i list = filter (\x -> (filter (\y -> y == i) x) /= []) list
```
chshersh, line 760: You correctly used `elem` before. And now you need to combine `filter` and `elem` to implement this function using a combination of standard functions:

```suggestion
contains i list = filter (\xs -> elem i xs) list
```

After you remove the special case on the line above, you can eta-reduce this function as well. Here you have two opportunities for eta-reduction:

* The `list` argument to `filter`
* The `xs` argument to `elem`

src/Chapter2.hs
```diff
@@ -858,7 +885,15 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind [x] = [x]
+rewind (x:xs) = go [x] xs
```
chshersh, line 891: You don't need special cases here 🙂 I believe, if you just define `rewind` in terms of eta-reduce helper `go` function, it will work just fine:
```suggestion
rewind = go []
```

src/Chapter2.hs
ramirezalejo: Got it, I changed it to avoid using the standard

src/Chapter2.hs
```diff
@@ -728,7 +742,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate = concatMap (\ (x) -> (replicate x x) )
```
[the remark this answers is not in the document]
ramirezalejo: Got it!

src/Chapter2.hs
ramirezalejo: Noted!

src/Chapter2.hs
ramirezalejo: I see, I forgot completely the hint this time, I fixed it and even eta reduced the list arg, I think I am starting to get it!

src/Chapter2.hs
ramirezalejo: You right, it works, I think this is amazing, it's just a bit hard to think this way

chshersh:
@ramirezalejo This looks great! 🏁 I'm going to close this PR as we can't merge solutions to our repo. But feel free to open other PRs with solutions to other chapters 🙂

PR #407, opened 2021-03-31
Finish Chapter1

oren-l:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -428,7 +428,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Num a => a -> a -> a
```

src/Chapter1.hs
```diff
@@ -490,7 +490,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Integral a => a -> a
+lastDigit n = (abs n) `mod` 10
```
vrom911, line 494: Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1: 

```suggestion
lastDigit n = abs n `mod` 10
```

src/Chapter1.hs
```diff
@@ -520,7 +521,9 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if abs(x) < abs(y)
```
vrom911, line 524: No need to add brackets to one variable 🙂 

Also, in Haskell, all functions are separated from their arguments by space 👾 

```suggestion
closestToZero x y = if abs x < abs y
```

src/Chapter1.hs
```diff
@@ -553,8 +556,11 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: Ord a => a -> a -> a -> a
+mid x y z
+  | x >= min y z && x <= max y z  = x
+  | y >= min x z && y <= max x z  = y
+  | otherwise                     = z
```
vrom911, line 563: Very elegant solution 💃🏼

src/Chapter1.hs
```diff
@@ -568,7 +574,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c = any (==c) "aeiou"
```
vrom911, line 578: Great 💡 
```suggestion
isVowel c = any (== c) "aeiou"
```

src/Chapter1.hs
```diff
@@ -631,8 +638,10 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Integral a => a -> a
+sumLast2 n = 
+  let last2 = (abs n) `mod` 100
```
vrom911, line 643:
```suggestion
  let last2 = abs n `mod` 100
```

src/Chapter1.hs
```diff
@@ -631,8 +638,10 @@ Implement a function that returns the sum of the last two digits of a number.
...
+  in  last2 `div` 10 + last2 `mod` 10
```

```haskell
mod ... 10
div ... 10
```

```haskell
(a, b) = divMod ... 10
```

src/Chapter1.hs
```diff
@@ -652,8 +661,12 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Integral a => a -> a
+firstDigit n = helper $ abs n
```
vrom911, line 665: Fantastic idea to use a helper function 💎

vrom911 (approved):
Congrats on finishing Chapter 1! 🎉 
Awesome start 👏🏼

chshersh:

PR #408, opened 2021-04-02
solution for chapter1

iurab:
### Solutions for Chapter1

cc @vrom911 @chshersh

Congratulations for the idea and the implementation.
It's exactly what I needed to try after some introduction videos about Haskell.

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -428,7 +428,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Num a => a -> a -> a
```

src/Chapter1.hs
```diff
@@ -449,7 +449,7 @@ Implement the function that takes an integer value and returns the next 'Int'.
   function body with the proper implementation.
 -}
 next :: Int -> Int
-next x = error "next: not implemented!"
+next x = x+1
```
vrom911, line 452: Nice!

In Haskell all functions and operators are separated from its arguments by spaces 👾 
```suggestion
next x = x + 1
```

src/Chapter1.hs
```diff
@@ -631,8 +641,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = x1 + x2
+  where
+    x1 = abs n `rem` 100 `div` 10
```
vrom911, line 647: When using functions written in the infix form, it is important to be cautious with the precedence as it is not explicitly determined for each function (unlike operators). So, it would be better to put braces in this case, to be sure that it will work as you expected.

Also, usually `rem` is used in pair with `quot`, and `mod` with `div`. You can also find such interesting functions like `divMod` or `quotRem` that could also be useful in here 🙂

src/Chapter2.hs
```diff
@@ -336,7 +348,13 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList x y zs
+  | x < 0 = []
+  | y < 0 = []
+  | null zs = []
+  | x > y = []
+  | x == 0    = head zs : subList x (y-1) (tail zs)
```
vrom911, line 356: In this case, you can just use `take` to take `y` elements 🙂

src/Chapter2.hs
```diff
@@ -336,7 +348,13 @@ from it!
...
+  | otherwise = subList (x-1) (y-1) (tail zs)
```
vrom911, line 357: Using `head` and `tail` is not efficient. You can take a look at `take` and `drop` functions that could help you here 🙂

src/Chapter2.hs
```diff
@@ -621,7 +642,14 @@ Write a function that takes elements of a list only in even positions.
...
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven = go 0
+  where
+     go :: Int -> [a] -> [a]
+     go _ [] = []
+     go acc (y:ys)
+      | even acc  = y : go (acc+1) ys
+      | otherwise = go (acc+1) ys
```
vrom911, line 652: Nice idea on the helper functions :+1:

A few additional thoughts on your idea: as you have only two states of the operation (odd or even), instead of accumulators of the current position number, you can just use `Bool`. And on each next recursion call, you just give it `not acc`. This will remove the necessity to check  `even` each time, which would save a couple of functions call on each interaction.

src/Chapter2.hs
```diff
@@ -728,7 +756,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate xs = concat (map (\x -> replicate x x) xs)
```
vrom911, line 759: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -741,7 +769,11 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Eq a => a -> [[a]] -> [[a]]
+contains _ [] = []
+contains x (y:ys)
+  | x `elem` y  = y : contains x ys
+  | otherwise   = contains x ys
```
vrom911, line 776: You can use `filter` function that already does exactly the logic you describe here 🙂 👍🏼

src/Chapter2.hs
```diff
@@ -781,13 +813,13 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
--- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
--- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 822: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -842,7 +874,11 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate n xs
+  | null xs   = []
+  | n < 0     = []
+  | otherwise = drop n (take (n+length xs) (cycle xs))
```
vrom911, line 881: Nice one! 👏🏼 

You can also use a trick for efficiency: you can notice that when `n` is big, this function will do much more cycles. So you can use `mod` on the length of the list 👍🏼

src/Chapter2.hs
```diff
@@ -858,7 +894,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind (x:xs) = rewind xs ++ [x]
```
vrom911, line 899: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Fantastic job! 🎉 👏🏼

vrom911:
@iurab looks great! This PR is ready to go 🚢 

I'm going to close it, as we can't merge solutions into our course. But feel free to open other PRs with the following Chapters 🙂

PR #425, opened 2021-09-12
Chapter 1 and 2

JosARodriguez:
### Solutions for Chapters 1 and 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a-> a-> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +491,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
```

src/Chapter1.hs
```diff
@@ -520,7 +522,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if (min (abs x) (abs y) == abs x) then x else y
```
vrom911, line 525: Great!
You could also use a comparison operator like `abs x < abs y` 🙂

src/Chapter1.hs
```diff
@@ -568,7 +573,14 @@ True
...
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c 
+    | c == 'a' = True
+    | c == 'e' = True
+    | c == 'i' = True
+    | c == 'o' = True
+    | c == 'u' = True
+    | otherwise = False
```
vrom911, line 583: Nice one! 👍🏼 

```haskell
isVowel c = c == 'a' || c == 'e' || ....
```

src/Chapter1.hs
```diff
@@ -631,8 +643,12 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
+sumLast2 :: Int -> Int
+sumLast2 n = a + b
+  where 
+    a = mod n 10
+    b =div (mod (n-a) 100)  10
```
vrom911, line 650: I don't think it is necessary to subtract `a` in this one 🤔 
Also, in this function you need also to consider negative input (hint: check out `abs` function 😉 )

src/Chapter1.hs
```diff
@@ -652,9 +668,14 @@ Implement a function that returns the first digit of a given number.
...
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit n = x n
+  where 
+    x :: Int -> Int
+    x n 
+      | n < 10 = n
```
vrom911, line 676: I see, that for a negative number like -19, this implementation will return the negative number itself instead of the last digit because the first check n < 10 will always succeed for negative numbers.

src/Chapter1.hs
```diff
@@ -652,9 +668,14 @@ Implement a function that returns the first digit of a given number.
...
+      | div (n - (mod n 10)) 10 < 10 = div (n - (mod n 10)) 10 
+      | otherwise = x (div (n - (mod n 10)) 10)
```
vrom911, line 678: No need to subtract in these cases as well👌🏼

src/Chapter2.hs
```diff
@@ -728,7 +737,8 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate [] = []
+smartReplicate (x:xs) = (replicate x x)++(smartReplicate xs)
```
vrom911, line 741: Great! 👍🏼 

Alternatively, you can use the standard function `concatMap` which is a combination of `concat` and `map` as the name suggests 🙂

src/Chapter2.hs
```diff
@@ -741,7 +751,16 @@ the list with only those lists that contain a passed element.
...
+contains :: Int -> [[Int]] -> [[Int]]
+contains _ [] = []
+contains n (x:xs) 
+  | (isThreeIn n x) /= [] = (isThreeIn n x):(contains n xs)
+  | otherwise = contains n xs
+  where
+    isThreeIn :: Int -> [Int] -> [Int]
+    isThreeIn m y
+      | (length (filter (==m) y))>0 = y
+      | otherwise = []
```
vrom911, line 763: There is a standard function [`elem`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Prelude.html#v:elem) that does this check for you 🙂

src/Chapter2.hs
```diff
@@ -781,13 +800,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x  = filter (< x) 
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 811: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -858,7 +882,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind x = (drop ((length x)-1) x) ++ rewind (take ((length x)-1) x)
```
vrom911, line 887: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Fantastic work 👏🏼

chshersh:

PR #427, opened 2021-09-16
First chapter finished!!

PabloSalinas27:
Going for the others!!
Very fun challenge, on the final boss got stuck bc I dndt see i was using > instead of ==

### Solutions for Chapter {1}

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -210,9 +210,12 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-True :: Bool>>> :t 'a'
+True :: Bool
+>>> :t 'a'
+'a' :: Char
+
 >>> :t 42
-<42 :: Num p => p>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
@@ -314,7 +317,7 @@ False
 >>> 10 < 20 || 20 < 5
 True
 >>> 2 ^ 10  -- power
-False
+1024
 >>> not False
 True
 >>> div 20 3  -- integral division
@@ -479,8 +482,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit :: Int -> Int 
-lastDigit n = mod n 10
+lastDigit :: Int -> Int
+lastDigit n = mod (abs n) 10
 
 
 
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +479,9 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int 
+lastDigit n = mod n 10
```
vrom911, line 483: Your implementation is almost correct :ok: 

src/Chapter1.hs
```diff
@@ -568,7 +564,14 @@ True
...
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool 
+isVowel c 
+  | c == 'a' = True 
+  | c == 'e' = True 
+  | c == 'i' = True 
+  | c == 'o' = True 
+  | c == 'u' = True 
+  | otherwise = False
```
vrom911, line 574: Nice one! 👍🏼 

```haskell
isVowel c = c == 'a' || c == 'e' || ...
```

src/Chapter1.hs
```diff
@@ -631,8 +634,13 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2::Int -> Int
+sumLast2 n = last n + prelast n
+  where
+    last::Int -> Int
+    last n = mod n 10
```
vrom911, line 641: `last` is the name of a standard function. Haskell allows name shadowing, that's why it works. Though, in practice, it's usually better to avoid name clashes with other functions cause it could be confusing 🙂

Also, I guess you could check what would happen on negative input here ;)

src/Chapter1.hs
```diff
@@ -631,8 +634,13 @@ Implement a function that returns the sum of the last two digits of a number.
...
+    prelast::Int -> Int
+    prelast n = mod (div n 10) 10
```
vrom911, line 643: You correctly noticed that it is the `div` and `mod`, cool 😎 

One hint to make your solution even shorter: you can see that you use both in `last` and `prelast`:

```haskell
mod smth 10
div smth 10
```

```haskell
(x, y) = divMod (abs n) 10
```

src/Chapter1.hs
```diff
@@ -652,9 +660,12 @@ Implement a function that returns the first digit of a given number.
 You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit::Int -> Int 
+firstDigit n 
+  | div n 10 == 0 = n
+  | otherwise = firstDigit (div n 10)
```
vrom911, line 666: Great one, but need to work on negative cases as well 🙂

vrom911 (approved):
Awesome work on the Chapter 👏🏼 🎉

chshersh:

PR #428, opened 2021-09-19
salkvus has finished Chapter1

salkvus:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -574,10 +574,7 @@ True
 False
 -}
 isVowel :: Char -> Bool
-isVowel c
-        | c == 'a' || c == 'e' || c == 'i' || c == 'o' || c == 'u' = True
-        | otherwise = False
-
+isVowel c = c == 'a' || c == 'e' || c == 'i' || c == 'o' || c == 'u'
 
 {- |
 == Local variables and functions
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -568,7 +573,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+        | c == 'a' || c == 'e' || c == 'i' || c == 'o' || c == 'u' = True
+        | otherwise = False
```
vrom911, line 579: Nice one! 👍🏼 

```haskell
isVowel c = c == 'a' || c == 'e' || c == 'i' ...
```

src/Chapter1.hs
```diff
@@ -628,11 +636,17 @@ Implement a function that returns the sum of the last two digits of a number.
...
+
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+      let num = mod (abs n) 100
+          in div num 10 + mod num 10
```
vrom911, line 649: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod num 10
div num 10
```

```haskell
(x, y) = divMod num 10
```

vrom911 (approved):
Awesome work 👍🏼

src/Chapter1.hs
salkvus: Thank you for your helpful comments

src/Chapter1.hs
salkvus: It turns out to be so simple :) Thanks

chshersh:

PR #432, opened 2021-10-01
[#430] Chapter 3 knight Monster clarification

### LINKED ISSUE #430: Chapter 3 knight Monster clarification
@vrom911 @chshersh Some clarifications needed please ?

```
Prepare to defend the honour of our kingdom! A monster attacks our brave knight.
Help him to fight this creature!
Define data types for Knights and Monsters, and write the "fight" function.
Both a knight and a monster have the following properties:
 ✦ Health (the number of health points)
 ✦ Attack (the number of attack units)
 ✦ Gold (the number of coins)
When a monster fights a knight, the knight hits first, and the monster hits back
only if it survives (health is bigger than zero). A hit decreases the amount of
health by the number represented in the "attack" field.
Implement the "fight" function, that takes a monster and a knight, performs the
fight following the above rules and returns the amount of gold the knight has
after the fight. The battle has the following possible outcomes:
 ⊛ Knight wins and takes the loot from the monster and adds it to their own
   earned treasure
 ⊛ Monster defeats the knight. In that case return -1
 ⊛ Neither the knight nor the monster wins. On such an occasion, the knight
   doesn't earn any money and keeps what they had before.
```
   
It seems the question is not really clear, should this be interpreted as a fight between knight and monster with multiple rounds
or a single round ?

Based on this [commit from a PR](https://github.com/kowainik/learn4haskell/pull/378/commits/085d26fe2dcf86f098ede96030329f02f93f3278) and many other PR's which are similar and accepted as an answer. I don't understand, how can the below function be true based on the following data:

```
balrog :: Monster
balrog = Monster {
          monsterHealth = 97,
          monsterAttack = 5,
          monsterGold   = 70
         }

arthur :: Knight
arthur = Knight {
          knightHealth = 93,
          knightAttack = 7,
          knightGold   = 82
         }


fight (Knight _ knightHealth knightAttack knightGold) (Monster monsterHealth monsterAttack monsterGold)
    | knightAttack >= monsterHealth = knightGold + monsterGold
    | monsterAttack >= knightHealth = -1
    | otherwise = knightGold
```

Based on the first condition `knightAttack >= monsterHealth` will be false, hence Arthur loses, but in fact if the fight goes on until either is dead (health 0) then Arthur would win.

chshersh:
Resolves #430

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -372,6 +372,8 @@ after the fight. The battle has the following possible outcomes:
  ⊛ Neither the knight nor the monster wins. On such an occasion, the knight
    doesn't earn any money and keeps what they had before.
 
+♫ NOTE: In this task, you need to implement only a single round of the fight.
+
 -}
 
 {- |


vrom911 (approved):
👌🏼

PR #435, opened 2021-10-02
Solve chapter 1

javatarz:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -492,7 +492,7 @@ Implement a function that returns the last digit of a given number.
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit :: Int -> Int
-lastDigit n = mod n 10
+lastDigit n = mod (abs n) 10
 
 
 {- |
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -553,8 +554,17 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+tripleApply :: (Int -> Int -> Int) -> Int -> Int -> Int -> Int
```
javatarz, line 557: I just guessed the syntax here based on the type declaration and rules specified above. The mid implementation still feels too complicated (and has too much repetition). I'm guessing we'll learn to optimise it upcoming chapters.

src/Chapter1.hs
```diff
@@ -568,7 +578,9 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+
+isVowel :: Char -> Bool
+isVowel c = c == 'a' || c == 'e' || c == 'i' || c == 'o' || c == 'u'
```
javatarz, line 583: The description asked to use guards but it didn't make sense to me here since the conditional already returns a `Bool`. Is there a reason why guards were recommended here?

chshersh:

PR #439, opened 2021-10-03
final boss yet to be done

pritindra:
### Solutions for Chapter {1}

Please help me with the final boss. Give me some kind of power-up !!

cc @vrom911 @chshersh


iburzynski:
@pritindra let's try to break the problem down a little to make the boss less intimidating.

It's a recursive problem, which means we have to think about two things:
1. What is the base (terminal) case, and what does it need to do?
2. What is the recursive case, and what does it need to do?

Recursive functions are easy to write in Haskell and can be done in a couple of ways, but I chose to use guards for this problem. We can write one guard statement for the base case and then a second one with 'otherwise' for the recursive case.

Since we're trying to get the first digit of a number, our base case guard statement is going to execute once there are no more digits to the left. That's where our program terminates and returns our answer. But how will we get there?

We want our recursive case to step over all the other digits from right to left, calling firstDigit each time until we hit the base case. A recursive case should always be shrinking its input each time it's called in order to guarantee we reach the base case and don't get stuck in an infinite loop: for example, if we have a 3 digit number, we want to first throw away the ones value, then the tens, and finally return the hundreds value.

So now the problem is a little clearer, and we can reframe it to get even closer to a solution.

1. How do we tell when we are at the leftmost digit (i.e. we can't step any further)? This needs to be a boolean expression that unlocks the first guard statement.
2. How do we step over a digit once we've determined it isn't the leftmost (i.e. if the boolean expression evaluates to false)?

The answer to both of these are related. They can be accomplished using integral division, which we were introduced to in this chapter. You've likely used it during the previous exercises.

pritindra:
Thanks a lot @iburzynski . I now understand it a bit better.

chshersh:

PR #453, opened 2021-10-09
CI: use major version and add workflow_dispatch

imba-tjd:
I think there is no need to lock minor versions.

workflow_dispatch will add a button on Actions tab to let users run the CI manually.


chshersh (approved):
Thanks! I wasn't aware of `workflow_dispatch`. This looks neat 💯

PR #472, opened 2021-10-19
Pritindra

pritindra:
### Solutions for Chapter {2}

give some feedback

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num t => t
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +491,10 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit x = mod x 10
```

src/Chapter1.hs
```diff
@@ -520,7 +524,9 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if (abs x) > (abs y) then y else x
```
vrom911, line 527: `()` are redundant here. I see how it can be confusing to see when to use them and when not to use, especially in the beginning. But it's usually better to remove redundant brackets for cleaner code :broom: 
```suggestion
closestToZero x y = if abs x > abs y then y else x

File: src/Chapter1.hs [hunk 3]
```diff
@@ -632,7 +649,13 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n =
+    let first = mod n 10
```
vrom911, line 654: Nice one!
But you may need to use `abs` to handle negative numbers correctly 🙂

src/Chapter2.hs
```diff
@@ -621,7 +641,12 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] = []
+takeEven[x] = [x]
+takeEven [x, y] = [x]
+takeEven (x:y:xs) = x : takeEven xs
```
vrom911, line 648: Great!
As `y` is not used in the right part, you can use wildcard instead of the name `y`:

```suggestion
takeEven (x:_:xs) = x : takeEven xs
```

src/Chapter2.hs
```diff
@@ -728,7 +753,8 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate x = concat (map (\t -> replicate t t) (x))
```
vrom911, line 756: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -784,11 +812,13 @@ divideTenBy :: Int -> Int
...
-
+--listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: (Ord a, Eq a) => a -> [a] -> [a]
+listElementsLessThan x = filter (< x)
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
-
+--pairMul xs ys = zipWith (*) xs ys
+pairMul :: (Num a) => [a] -> [a] -> [a]
+pairMul xs = zipWith (*) xs
```
vrom911, line 821: You can go further and eta-reduce `xs` as well:

```suggestion
pairMul = zipWith (*)
```

src/Chapter2.hs
```diff
@@ -842,7 +872,12 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+--rotate = error "rotate: Not implemented!"
+rotate :: (Num a) => Int -> [a] -> [a]
+rotate n xs 
+    | n < 0 = []
+    | otherwise = drop n $ take (length xs + n) $ cycle xs
```
vrom911, line 879: Absolutely correct!

You can optimise this solution if you use `mod` by the length of the list on `n`, so you won't do extra work on big input numbers 🙂

src/Chapter2.hs
```diff
@@ -858,8 +893,10 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
-
+--rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind xs = foldl (\x y -> y:x) [] xs
```
vrom911, line 899: Ohh, this is an awesome solution!

You can skip first pattern matching on `[]` as `fold` is covering this case already!

vrom911 (approved):
Fantastic job! 🎉 👏🏼

chshersh:

PR #473, opened 2021-10-20
Chapter 3

polemius:
### Solutions for Chapter 3

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -344,6 +344,21 @@ of a book, but you are not limited only by the book properties we described.
 Create your own book type of your dreams!
 -}
 
+data Book = Book
+  { author      :: String
+  , year        :: Int
+  , isbn        :: String
+  , description :: String
+  } deriving (Show)
+
+zeroToOne :: Book
+zeroToOne = Book {
+  author = "Peter Thiel"
+  , year = 2014
+  , isbn = "978075"
+  , description = "Notes on Startups, or Hot to build the future"
+}
+
 {- |
 =⚔️= Task 2
 
@@ -373,6 +388,22 @@ after the fight. The battle has the following possible outcomes:
    doesn't earn any money and keeps what they had before.
 
 -}
+
+data Creature = Creature
+  {
+  health  :: Int
+  , attack  :: Int
+  , gold    :: Int
+  } deriving (Show)
+
+type Knight = Creature
+type Monster = Creature
+
+fight :: Knight -> Monster -> Int
+fight knight monster
+  | attack knight >= health monster = gold knight + gold monster
+  | attack monster >= health knight = -1
+  | otherwise = gold knight
 
 {- |
 =🛡= Sum types
@@ -460,6 +491,11 @@ Create a simple enumeration for the meal types (e.g. breakfast). The one who
 comes up with the most number of names wins the challenge. Use your creativity!
 -}
 
+data Meal = Breakfast
+  | Brunch
+  | Lunch
+  | Supper
+
 {- |
 =⚔️= Task 4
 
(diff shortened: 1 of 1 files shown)


src/Chapter3.hs
```diff
@@ -374,6 +389,22 @@ after the fight. The battle has the following possible outcomes:
...
+  {
+  health  :: Int
+  , attack  :: Int
+  , gold    :: Int
+  } deriving (Show)
+
+type Knight = Creature
+type Monster = Creature
+
+fight :: Knight -> Monster -> Int
```
vrom911, line 402: Nice usage of type aliases!

```haskell
> fight knight monster
> fight monster knight
```

```haskell
newtype Monster = Monster Creature
newtype Knight = Knight Creature
```

src/Chapter3.hs
```diff
@@ -480,6 +516,37 @@ After defining the city, implement the following functions:
    and at least 10 living __people__ inside in all houses of the city in total.
 -}
 
+
+type CastleName = String
+type Walls = Int
+newtype House = House Int
```
vrom911, line 522: To avoid building houses with more than 4 people you need to restrict usage of the Constructor and allow to create `House`s only with, for example,  `mkHouse` function.

src/Chapter3.hs
```diff
@@ -480,6 +516,37 @@ After defining the city, implement the following functions:
...
+
+data City = CityWithCastle CastleName Walls MainBuilding Houses
+  | OpenCity MainBuilding Houses
+  deriving (Show)
+
+getHousePeople :: House -> Int
+getHousePeople (House x) = x
+
+buildCastle :: City -> CastleName -> City
+buildCastle (CityWithCastle _ walls mainBuilding houses) castleName = CityWithCastle castleName walls mainBuilding houses
```
vrom911, line 537: I think that here pattern matching is not exhaustive. You need also to look into the case with the `OpenCity` 🙂

src/Chapter3.hs
```diff
@@ -560,22 +627,31 @@ introducing extra newtypes.
...
+calculatePlayerDefense :: PlayerArmor -> PlayerDexterity -> PlayerDefense
+calculatePlayerDefense (Armor' armor) (Dexterity' dexterity) = Defense' (armor * dexterity)
 
-calculatePlayerHit :: Int -> Int -> Int -> Int
-calculatePlayerHit damage defense health = health + defense - damage
+calculatePlayerHit :: PlayerDamage -> PlayerDefense -> PlayerHealth -> PlayerHealth
+calculatePlayerHit (Damage damage) (Defense' defense) (Health' health) = Health' (health + defense - damage)
 
 -- The second player hits first player and the new first player is returned
 hitPlayer :: Player -> Player -> Player
```
vrom911, line 657: Nice work on newtypes! :medal_sports: Newtypes are one very important feature in Haskell :+1:

src/Chapter3.hs
```diff
@@ -907,9 +999,25 @@ Implement instances of "Append" for the following types:
...
+instance Append [a] where
+    append :: [a] -> [a] -> [a]
+    append xs ys = xs <> ys
+
+instance (Semigroup a) => Append (Maybe a) where
+    append :: Maybe a -> Maybe a -> Maybe a
+    append Nothing (Just a) = Just a
+    append (Just a) Nothing = Just a
+    append (Just a) (Just b) = Just (a <> b)
+    append _ _ = Nothing
```
vrom911, line 1020: I think this case could be avoided if you generilise first two 🙂 
You can notice that in case of one `Nothing` the result is always the other argument 👌🏼

src/Chapter3.hs
```diff
@@ -971,6 +1079,30 @@ implement the following functions:
...
+             deriving (Ord, Eq, Show, Enum)
+
+isWeekend :: Weekday -> Bool
+isWeekend Saturday = True
+isWeekend Sunday   = True
+isWeekend _        = False
+
+nextDay :: Weekday -> Weekday
+nextDay Sunday = Monday
+nextDay d = succ d
```
vrom911, line 1098: Unfortunately, `succ` is not a safe function as it will fail on runtim e when given the max boundary element (in this case, Sunday). You need to catch this case separately.

src/Chapter3.hs
```diff
@@ -971,6 +1079,30 @@ implement the following functions:
...
+
+daysToParty :: Weekday -> Int
+daysToParty Friday = 0
+daysToParty d
+    | fromEnum d < fromEnum Friday = fromEnum Friday - fromEnum d
+    | otherwise = abs (fromEnum Friday) - fromEnum d + fromEnum Friday
```
vrom911, line 1104: You are using `fromEnum Friday` a lot here, so maybe it could be a good idea to put it into a separate var.

Also, I guess, you can combine these cases together with the usage of `mod` 😉

vrom911 (approved):
Great work on types construction 🚧 👷🏼

PR #474, opened 2021-10-20
Chapter 4

polemius:
### Solutions for Chapter 4

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter4.hs
+++ b/src/Chapter4.hs
@@ -114,23 +114,23 @@ As always, try to guess the output first! And don't forget to insert
 the output in here:
 
 >>> :k Char
-
+Char :: *
 >>> :k Bool
-
+Bool :: *
 >>> :k [Int]
-
+[Int] :: *
 >>> :k []
-
+[] :: * -> *
 >>> :k (->)
-
+(->) :: * -> * -> *
 >>> :k Either
-
+Either :: * -> * -> *
 >>> data Trinity a b c = MkTrinity a b c
 >>> :k Trinity
-
+Trinity :: * -> * -> * -> *
 >>> data IntBox f = MkIntBox (f Int)
 >>> :k IntBox
-
+IntBox :: (* -> *) -> *
 -}
 
 {- |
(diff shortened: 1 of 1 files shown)


src/Chapter4.hs
```diff
@@ -472,10 +478,11 @@ Implement the Applicative instance for our 'Secret' data type from before.
 -}
 instance Applicative (Secret e) where
     pure :: a -> Secret e a
-    pure = error "pure Secret: Not implemented!"
+    pure = Reward
 
     (<*>) :: Secret e (a -> b) -> Secret e a -> Secret e b
-    (<*>) = error "(<*>) Secret: Not implemented!"
+    (<*>) (Trap e) _ = Trap e
+    (<*>) (Reward f) a = Reward f a
```
vrom911, line 485: That's right!

With operators, you can specify the body in the infix form as well:

```suggestion
    Trap e <*> _ = Trap e
    Reward f <*> a = Reward f a
```

This would do the same 👌🏼

src/Chapter4.hs
```diff
@@ -488,7 +495,17 @@ Implement the 'Applicative' instance for our 'List' type.
...
+  pure :: a -> List a
+  pure a = Cons a Empty
 
+  (<*>) :: List (a -> b) -> List a -> List b
+  Empty <*> _ = Empty
+  (Cons f fs) <*> x = concatList (fmap f x) (fs <*> x)
+
+concatList :: List a -> List a -> List a
+concatList Empty x = x
+concatList (Cons a as) x = Cons a (concatList as x)
```
vrom911, line 508: Perfect!

```haskell
concatList xs Empty = xs
```

vrom911 (approved):
:tada: Fantastic work!

You can also ask @chshersh or me any questions and we can discuss any topics you would like to dig into as well  :slightly_smiling_face: Don't be a stranger!

<hr>

If you would like to continue your journey to Haskell, check out our [Haskell Knowledge Map](https://kowainik.github.io/images/Haskell_Knowledge_Map.png) for the list of topics to explore 🗺️

PR #479, opened 2021-10-22
`UPDATE` Chapter1.hs - All tasks completed

sreekesari-vangeepuram:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create a list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -553,8 +554,12 @@ value after "=" where the condition is true.
...
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | (x <= y && x >= z) || (x >= y && x <= z) = x
+    | (y <= x && y >= z) || (y >= x && y <= z) = y
+    | (z <= x && z >= y) || (z >= x && x <= y) = z
+    | otherwise = error "Unkown error encountered!"
```
vrom911, line 562: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

In this case, you can use `otherwise` instead of the third case, as `z` is the only possible output the function can return at this point:

```haskell
    | otherwise = z
```

src/Chapter1.hs
```diff
@@ -631,8 +642,13 @@ Implement a function that returns the sum of the last two digits of a number.
...
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = l1 n + l2 n
+  where
+    l1 :: Int -> Int
+    l1 x = mod (abs x) 10
+    l2 :: Int -> Int
+    l2 x = l1 (div (abs x) 10)
```

```haskell
mod m 10
div m 10
```

```haskell
(x, y) = divMod m 10
```

src/Chapter2.hs
```diff
@@ -336,7 +336,11 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList ll ul list
+    | ll < 0 || ul < 0 = []
+    | ll > ul = []
+    | ll <= ul = drop ll (take (ul + 1) list)
```
vrom911, line 342: You can use `otherwise` here and get rid of the `error` below 🙂

src/Chapter2.hs
```diff
@@ -606,7 +612,8 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate [] = []
+duplicate (x:xs) = [x, x] ++ duplicate xs
```
vrom911, line 616: It is more efficient to construct lists using the cons operator directly without introducing intermediate values.
```suggestion
duplicate (x:xs) = x : x : duplicate xs
```

src/Chapter2.hs
```diff
@@ -621,7 +628,14 @@ Write a function that takes elements of a list only in even positions.
...
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven list = go 0 list []
+  where
+    go :: Int -> [a] -> [a] -> [a]
+    go _ [] es = reverse es
+    go i (x:xs) es
+        | even i = go (i+1) xs (x:es)
```
vrom911, line 637: Nice idea on the helper functions :+1:

src/Chapter2.hs
```diff
@@ -728,7 +742,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate xs = concat [replicate x x | x <- xs]
```
vrom911, line 745: Great idea on list comprehension! 👍🏼 
Alternatively, you can use the standard function `concatMap` which is a combination of `concat` and `map` as the name suggests 🙂

src/Chapter2.hs
```diff
@@ -783,11 +798,11 @@ mastered the skill of eta-reducing.
...
 divideTenBy x = div 10 x
 
--- TODO: type ;)
+listElementsLessThan :: Int -> [Int] -> [Int]
 listElementsLessThan x l = filter (< x) l
 
--- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 805: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -842,7 +857,12 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate _ [] = []
+rotate n xs
+    | n < 0 = []
+    | n == (length xs) = xs
```
vrom911, line 864: This is an interesting check, but maybe instead you can use `mod n len`, so it would be even more efficient!

src/Chapter2.hs
```diff
@@ -858,7 +878,12 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind xs = go xs []
+  where
+    go :: [a] -> [a] -> [a]
+    go [] rxs = rxs
+    go (x:xs) rxs = go xs (x:rxs)
```
vrom911, line 886: Bravo, absolutely correct solution 👏🏼

vrom911 (commented):
Fantastic work on both chapters! 💯 🎉

PR #481, opened 2021-10-23
Tasks for Chapter 1

pparmin:

cc @vrom911 @chshersh

Thank you so much for providing this amazing learning resource! I have tried my best to implement everything, but it seems there are still a number of build errors. Your tests have helped me to find a number of mistakes already, so I am curious what has gone wrong here :) Thank you for the review!

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +491,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Integer -> Integer
+lastDigit n = n `mod` 10
```

src/Chapter1.hs
```diff
@@ -520,11 +522,10 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
-
+closestToZero x y = if abs(x) < abs(y) then x else y
```
vrom911, line 525: `()` are redundant here. I see how it can be confusing to see when to use them and when not to use, especially in the beginning. But it's usually better to remove redundant brackets for cleaner code :broom: 
```suggestion
closestToZero x y = if abs x < abs y then x else y

File: src/Chapter1.hs [hunk 3]
```diff
@@ -554,7 +555,14 @@ value after "=" where the condition is true.
...
 
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z 
+    | x > y && x < z = x
+    | x > z && x < y = x
+    | y > x && y < z = y
+    | y > z && y < x = y
+    | z > x && z < y = z
+    | z > y && z < x = z
```
vrom911, line 565: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

In this case, last two clauses can be replaced by a single:

```suggestion
    | otherwise = z
```

As at this point, this is the only possible result this function can return, so no need to make these checks 🙂

src/Chapter1.hs
```diff
@@ -632,9 +646,12 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
-
-
+sumLast2 :: Int -> Int
+sumLast2 n = last + penultimate  
+  where
+    getLast     = mod n 100
+    last        = mod getLast 10
```
vrom911, line 653: `last` is the name of a standard function. Haskell allows name shadowing, that's why it works. Though, in practice, it's usually better to avoid name clashes with other functions cause it could be confusing 🙂

src/Chapter1.hs
```diff
@@ -632,9 +646,12 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
 
-sumLast2 n = error "sumLast2: Not implemented!"
-
-
+sumLast2 :: Int -> Int
+sumLast2 n = last + penultimate  
+  where
+    getLast     = mod n 100
+    last        = mod getLast 10  
+    penultimate = div getLast 10
```
vrom911, line 654: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod m 10
div m 10
```

```haskell
(x, y) = divMod m 10
```

src/Chapter1.hs
```diff
@@ -653,7 +670,12 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int 
+firstDigit n 
+    | n > 0 && n < 10 = n
```
vrom911, line 675: I guess, `0` can be included into this case as well 👌🏼

src/Chapter1.hs
```diff
@@ -653,7 +670,12 @@ You need to use recursion in this task. Feel free to return to it later, if you
...
+    | n < 0           = firstDigit (n * (-1))
```
vrom911, line 676: It's better to use `abs` for negative numbers 🙂

vrom911 (approved):
Really great job on the Chapter! 🎉 
I left some comments, let me know if you want to discuss something in more detail 💯

chshersh:

PR #484, opened 2021-10-24
complete chapter 1 -L6

justlilith:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -489,8 +489,8 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
```

src/Chapter1.hs
```diff
@@ -520,7 +520,13 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y =
+  if
+    abs (x - 0) < abs (y - 0)
```
vrom911, line 525: No need to subtract 0 here as it won't affect the results 🙂 

```suggestion
    abs x < abs y
```

src/Chapter1.hs
```diff
@@ -553,9 +559,11 @@ value after "=" where the condition is true.
...
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
-
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | (y > x && x > z) || (y < x && x < z) = x
+    | (x > y && y > z) || (x < y && y < z) = y
+    | (y > z && z > x) || (y < z && z < x) = z
```
vrom911, line 566: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

In this case, instead of the last check, you can use `otherwise` as it is the only possible result this function can return at this point:

```suggestion
    | otherwise = z
```

src/Chapter1.hs
```diff
@@ -653,7 +672,15 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit n =
+    if
+      mod n 10 == div n 1
```
vrom911, line 677: `div n 1` is the same as just `n`, so you can use that directly 🙂 

Also, you need to add additional thing in order for this function to work on negative input 🙂

vrom911 (approved):
Fantastic job on the First Chapter 👏🏼 🎉

chshersh:

PR #497, opened 2021-11-02
completed chapter1

udaycruise2903:
### Solutions for Chapter 1

cc @vrom911 @chshersh

Thank you for creating learn4haskell. It really helps with getting started in haskell.

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -520,8 +521,10 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
-
+closestToZero x y 
+  | abs x <= abs y = x
+  | abs x >= abs y = y
+  | otherwise = error""
```
chshersh, line 527: You can use `if-then-else` instead of guards. This will also remove the last redundant `otherwise` case 🧹

src/Chapter1.hs
```diff
@@ -554,7 +557,12 @@ value after "=" where the condition is true.
 Casual reminder about adding top-level type signatures for all functions :)
 -}
 
-mid x y z = error "mid: not implemented!"
+mid :: Ord a => a -> a -> a -> a
+mid x y z 
+    | (x < y && x > z) || (x < z && x > y) = x
+    | (y < x && y > z) || (y < z && y > x) = y
+    | (z < y && z > x) || (z < x && y > z) = z
+    | otherwise = if x == y then x else z
```
chshersh, line 565: You can use `>=` and `<=` instead of `<` and `>`. And with this approach, you'll be able to replace last two guards with a single `otherwise`

src/Chapter1.hs
```diff
@@ -568,7 +576,14 @@ True
...
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+  | c == 'a' || c == 'A' = True 
+  | c == 'e' || c == 'E' = True
+  | c == 'i' || c == 'I' = True
+  | c == 'o' || c == 'O' = True 
+  | c == 'u' || c == 'U' = True
```
chshersh, line 585: Awesome solution! Absolutely correct!  👍🏻 

As `||` alread returns a value of type `Bool`, you don't need to use guards and return the result of operator application directly:

```haskell
isVowel c =
        c == 'a' || c == 'A'
    ||  c == 'e' || c == 'E'
    || ...
```

src/Chapter1.hs
```diff
@@ -631,8 +646,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int 
+sumLast2 n = onesDigit n + tensDigit n
+  where
+    onesDigit p = abs p `mod` 10
+    tensDigit p = (abs p `div` 10) `mod` 10
```
chshersh, line 653: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod (abs p) 10
div (abs p) 10
```

```haskell
```

chshersh (approved):
Nice work on Chapter One! 👏🏻

PR #508, opened 2021-11-27
Chapter 1 complete

thenic95:
### Solutions for Chapter 1

Hi,

I loved working on Chapter 1 it's really cool what you created here.

Thanks a lot.

Can't wait to read your feedback.

Best regards,
Nicolas

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num a => a
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -493,7 +493,7 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 lastDigit :: Int -> Int
-lastDigit n = n `mod` 10
+lastDigit n = abs(n) `mod` 10
  
 {- |
 =⚔️= Task 6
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -489,10 +492,9 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
-
-
+lastDigit :: Int -> Int
+lastDigit n = n `mod` 10
```
chshersh, line 496: Your implementation is almost correct :ok: 

src/Chapter1.hs
```diff
@@ -520,8 +522,9 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
-
+closestToZero x y = if abs (x) == abs(y) then error "values are the same, please choose separate values!" 
+  else if abs (x) > abs (y) then y else x
+-- How could I implemnt another
```
chshersh, line 527: If values are the same, you can return either of them. The task doesn't specify which one is better so you can simplify the implementation 👍🏻

src/Chapter1.hs
```diff
@@ -553,8 +556,14 @@ value after "=" where the condition is true.
...
-
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | x < y && x > z = x
+  | y < x && y > z = y
+  | x < z && x > y = x
+  | z < y && z > x = z
+  | y < z && y > x = y
+  | z < x && z > y = z
```
chshersh, line 566: As we mentioned, the compiler in Haskell is very attentive to the exhaustive pattern-matching. And here it would warn you that `Pattern matching is not exhaustive`, as the guards have quite complicated logic, and the compiler won't be able to prove that it covers all the cases.

Because of that, you will need to replace the last guard with – `| otherwise = z`, to tell the compiler, that your pattern matching is exhaustive 🙂


It looks like you can move all combine some of your guards into a single otherwise and return z in here.

src/Chapter1.hs
```diff
@@ -568,7 +577,15 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c 
+  | c == 'a' = True
+  | c == 'e' = True
```
chshersh, line 583: Absolutely correct!  👍🏻 

As `||` alread returns a value of type `Bool`, you don't need to use guards and return the result of operator application directly:

```haskell
isVowel c =
        c == 'a' || c == 'A'
    ||  c == 'e' || c == 'E'
    || ...
```

src/Chapter1.hs
```diff
@@ -631,8 +648,15 @@ Implement a function that returns the sum of the last two digits of a number.
...
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> (Int)
+sumLast2 n  
+  | n < 1000 =
+    let lastN = (n `mod` 10)
+        previousN 
+          | n < 100 = (n `div` 10)
+          | n < 1000 = ((n `mod` 100) `div` 10)
+    in (lastN + previousN)
+  | otherwise = error "Number is out of range. Choose a number smaller than 1000"
```
chshersh, line 659: Is there any specific reason to throw an error on numbers bigger than 1️⃣ 0️⃣ 0️⃣ 0️⃣ ?
I believe, the function should work for them as well 👍🏻

src/Chapter1.hs
```diff
@@ -652,9 +676,13 @@ Implement a function that returns the first digit of a given number.
...
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit n 
+  | n < 100 = n `div` 10
+  | n < 1000 = n `div` 100
+  | n < 10000 = n `div` 1000
+  | n < 100000 = n `div` 10000
```
chshersh, line 684: I'm afraid, this solution doesn't work for all the numbers 😞 
You need to implement a recursive function that will divide the number by 10 unless only a single digit left

chshersh (approved):
Great work on Chapter 1 👏🏻 

I reviewed only the first chapter as I see that you haven't solved the second chapter entirely and probably are in the process of solving it. Hope you can find my feedback helpful 🙂

src/Chapter1.hs
thenic95: Thank you for the feedback. 
I'm now using the `abs()` function to also evaluate negative numbers 😃

src/Chapter1.hs
thenic95: Okay, changed it 😄

chshersh:

PR #514, opened 2022-02-26
Thank You developers!!

abhideepd:
First of all, thank you for the hard work and thoughtfulness given to create this course, it has really helped me get my heads around Haskell.

It would make my day if you could suggest me some areas of improvements or some suggestions to hone skills in Haskell and Functional Programming 

Regards

### Solutions for Chapter {1}

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -39,7 +39,7 @@ Now, if you are ready, bring it on!
 -}
 
 module Chapter2 where
-
+import Data.List
 {-
 =🛡= Imports
 
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True, False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create a list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -653,7 +662,7 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit n = if (abs n)<10 then (abs n) else firstDigit (div (abs n) 10)
```
vrom911, line 665: Great job 💯 

Here you can use a bit of alignment as well:

```suggestion
firstDigit n = 
    if  abs n < 10
    then abs n
    else firstDigit (div (abs n) 10)
```

Also note, that you are using `abs n` a lot in this function. You can move it into a separate variable in side of the `where` construction, for example 🙂

src/Chapter2.hs
```diff
@@ -336,8 +348,11 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
-
+--subList = error "subList: Not implemented!"
+subList a b x
+  | a<0=[]
+  | b<0=[]
+  | otherwise=drop a (take (b+1) x)
```
vrom911, line 355: You can also check on the left bound to be bigger than the right one 🙂

src/Chapter2.hs
```diff
@@ -501,8 +518,11 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
-
+--isThird42 = error "isThird42: Not implemented!"
+isThird42 x
+    | length x<3=False
+    | (x!!2) == 42=True
```
vrom911, line 524: The `!!` function is not total, meaning it could fail on runtime. Even though you have a check on that, it is still better to avoid using that at all.
Also, the `length` function is costly here, and it won't work on infinite lists ⛓️ 
In this particular case, you could use the explicit pattern matching to solve this task 👌🏼

src/Chapter2.hs
```diff
@@ -606,8 +626,10 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
-
+--duplicate = error "duplicate: Not implemented!"
+duplicate x
+    | null x=[]
+    | otherwise=(head x):(head x):duplicate (tail x)
```
vrom911, line 632: Here it is also better to use pattern matching instead of using unsafe `head` function. (It is also more efficient 🙂 )

src/Chapter2.hs
```diff
@@ -621,7 +643,11 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+--takeEven = error "takeEven: Not implemented!"
+takeEven x
+    | ((length x) == 0) = []
+    | ((length x) == 1) = head x:[]
+    | otherwise=(head x):takeEven (tail (tail x))
```
vrom911, line 650: Same here 🙂

src/Chapter2.hs
```diff
@@ -741,8 +769,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+--contains = error "contains: Not implemented!"
+contains a x= filter (elem a) x
```
vrom911, line 773: Now, after you've mastered eta-reduction, you can apply such a technique to this function as well  🙂 

src/Chapter2.hs
```diff
@@ -781,13 +809,13 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
-divideTenBy x = div 10 x
+divideTenBy = div 10 
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan x= filter (< x) 
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul  = zipWith (*)
```
vrom911, line 818: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -858,8 +888,10 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
-
+--rewind = error "rewind: Not Implemented!"
+rewind x
+    | null x=[]
+    | otherwise=(rewind (tail x))++((head x):[])
```
vrom911, line 894: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Fantastic work 💯 
I have added some comment, hope it would be helpful 🙂

PR #520, opened 2022-04-06
Chapter 1

raishovan:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
-
+42 :: Num p => p
 A pair of boolean and char:
+
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +490,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit x = mod x 10
```

src/Chapter1.hs
```diff
@@ -631,9 +643,15 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
+sumLast2 :: Int -> Int
+sumLast2 n = let last    = mod n 10
+                 seclast = mod (div n 10) 10 -- div n 10 gives the digit stripped off last digit.
```

```haskell
mod m 10
div m 10
```

```haskell
(x, y) = divMod m 10
```

src/Chapter1.hs
```diff
@@ -653,8 +671,8 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit n = if n < 10 then n else firstDigit (div n 10)
```
vrom911, line 675: I see, that for a negative number like -19, this implementation will return the negative number itself instead of the last digit because the first check n < 10 will always succeed for negative numbers.

vrom911 (approved):
Great work! Thanks for taking your time to finish the Chapter 😊

PR #523, opened 2022-04-16
ch1 done

Aadamko1:

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Nump p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -429,6 +429,7 @@ task is to specify the type of this function.
 49
 -}
 
+squareSum :: Num a => a -> a -> a
```

src/Chapter1.hs
```diff
@@ -490,7 +491,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int 
+lastDigit x = x `mod` 10
```

src/Chapter1.hs
```diff
@@ -568,7 +577,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c = let vowel = c `elem` ['a','e','i','o','u'] in vowel
```
vrom911, line 581: Nice one! 👍🏼 

```haskell
```

src/Chapter1.hs
```diff
@@ -632,7 +642,14 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = digitOne + digitTwo
+  where
+    digitOne = n `mod` 10
+    inter = n `div` 10
```
vrom911, line 649: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod m 10
div m 10
```

```haskell
(x, y) = divMod m 10
```

src/Chapter1.hs
```diff
@@ -653,7 +670,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+  | n `div` 10 == 0 = n
+  | otherwise = firstDigit(n `div` 10)
```
vrom911, line 676: In Haskell, function is usually separated from the applied arguments with spaces 👾 

```suggestion
  | otherwise = firstDigit (n `div` 10)
```

vrom911 (approved):
Bravo! Very good job 💯

vrom911:
Closing as we cannot merge the solution into the course.
Feel free to open another PR with solutions and we can continue discussions there 🙂 👍🏼

PR #536, opened 2022-10-06
completed first assignment

kumarpriyanshu2k2:

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True,'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -428,7 +428,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-
+squareSum :: Int->Int->Int
```
vrom911, line 431: Correct! 💯 
Although, in Haskell all type operators and generally operators are separated with spaces like this:

```suggestion
squareSum :: Int -> Int -> Int
```

src/Chapter1.hs
```diff
@@ -448,8 +448,8 @@ Implement the function that takes an integer value and returns the next 'Int'.
   every type ｡.☆.*｡. No need to worry much about "error" here, just replace the
   function body with the proper implementation.
 -}
-next :: Int -> Int
-next x = error "next: not implemented!"
+next :: Integer->Integer
+next x = (x + 1)
```
vrom911, line 452: That's right!
Note that brackets are unnecessary 🙂 

```suggestion
next x = x + 1
```

src/Chapter1.hs
```diff
@@ -489,8 +489,9 @@ Implement a function that returns the last digit of a given number.
   results. Or you can try to guess the function name, search for it and check
   whether it works for you!
 -}
--- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int->Int
+lastDigit n = mod n 10
```

src/Chapter1.hs
```diff
@@ -631,8 +642,12 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int->Int
+sumLast2 n = 
+    let last = mod n 10
+        x = div n 10
```

```haskell
mod m 10
div m 10
```

```haskell
(x, y) = divMod m 10
```

kumarpriyanshu2k2:
Thank you so much for the valuable feedback I have made the necessary changes as per your suggestion

PR #537, opened 2022-10-06
Completed Chapter 1

ngntrgduc:
### Solutions for Chapter 1

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -213,7 +213,7 @@ True :: Bool
 >>> :t 'a'
 'a' :: Char
 >>> :t 42
-42 :: Num a => a
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num a => a
 
 A pair of boolean and char:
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -632,7 +642,11 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = a + mod b 10
+  where 
+    a = abs n `mod` 10
+    b = div (abs n - a) 10
```
vrom911, line 649: No need to subtract `a` when doing `div`, it'll take care of this 👌🏼 

```haskell
mod m 10
div m 10
```

```haskell
(x, y) = divMod m 10
```

src/Chapter1.hs
```diff
@@ -653,7 +667,11 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n 
+  | n < 10 = n
+  | n >= 10 = firstDigit m
+  where m = div (n - mod n 10) 10
```
vrom911, line 674: You can get rid of subtraction here too! 💯

src/Chapter1.hs
```diff
@@ -653,7 +667,11 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n 
+  | n < 10 = n
```
vrom911, line 672: I see, that for a negative number like -19, this implementation will return the negative number itself instead of the last digit because the first check n < 10 will always succeed for negative numbers.

vrom911 (approved):
Great work on the Chapter! 👏🏼

ngntrgduc:
Thank you ❤️. I've made some changes according to your feedback. I'm so excited to do the next chapter :">.

vrom911 (approved):
@ngntrgduc I think it looks great now! 🏅

PR #538, opened 2022-10-07
completed Chapter 2

kumarpriyanshu2k2:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter2.hs
+++ b/src/Chapter2.hs
@@ -136,43 +136,43 @@ functions in GHCi and insert the corresponding resulting output below:
 
 List of booleans:
 >>> :t [True, False]
-
+[True,False] :: [Bool]
 
 String is a list of characters:
 >>> :t "some string"
-
+"some string" :: [Char]
 
 Empty list:
 >>> :t []
-
+[] :: [a]
 
 Append two lists:
 >>> :t (++)
-
+(++) :: [a] -> [a] -> [a]
 
 Prepend an element at the beginning of a list:
 >>> :t (:)
-
+(:) :: a -> [a] -> [a]
 
 Reverse a list:
 >>> :t reverse
-
+reverse :: [a] -> [a]
 
 Take first N elements of a list:
 >>> :t take
-
+take :: Int -> [a] -> [a]
 
 Create a list from N same elements:
 >>> :t replicate
-
+replicate :: Int -> a -> [a]
 
 Split a string by line breaks:
 >>> :t lines
-
+lines :: String -> [String]
 
 Join a list of strings with line breaks:
 >>> :t unlines
-
+unlines :: [String] -> String
 
 -}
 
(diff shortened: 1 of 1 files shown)


src/Chapter2.hs
```diff
@@ -606,7 +613,8 @@ Implement a function that duplicates each element of the list
 
 -}
 duplicate :: [a] -> [a]
-duplicate = error "duplicate: Not implemented!"
+duplicate [] = []
+duplicate (x:xs)= [x] ++ [x] ++ duplicate(xs)
```
vrom911, line 617: It is more efficient to construct lists using the cons operator directly without introducing intermediate values.
```suggestion
duplicate (x:xs) = x : x : duplicate xs
```

src/Chapter2.hs
```diff
@@ -621,7 +629,9 @@ Write a function that takes elements of a list only in even positions.
 >>> takeEven [2, 1, 3, 5, 4]
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [Int] -> [Int]
+takeEven [] =[]
+takeEven(x:xs) = if mod x 2 == 0 then [x] ++ takeEven xs else [] ++ takeEven xs
```
vrom911, line 634: Here you also can avoid `++` operator by replacing it with `:` or no action at all!

```suggestion
takeEven(x:xs) = if mod x 2 == 0 then x : takeEven xs else takeEven xs
```

src/Chapter2.hs
```diff
@@ -741,7 +752,9 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int->[[Int]] -> [[Int]]
+contains _ []=[]
+contains y (x:xs) = if elem y x then [x] ++ contains y xs else []++ contains y xs
```
vrom911, line 757: You can use `filter` function here instead of the pattern matching, but this is nice solution already (you'll just need to replace `++` with `:` operator 🙂 )

src/Chapter2.hs
```diff
@@ -781,13 +794,13 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
-divideTenBy x = div 10 x
+divideTenBy  = div 10 
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan x  = filter (< x) 
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul  = zipWith (*)
```
vrom911, line 803: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -842,7 +855,12 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate :: Int -> [a] -> [a]
+rotate x s = 
+      let y = cycle s
```
vrom911, line 860: Unfortunately, `cycle` fails in runtime on empty lists :recycle: So you need to handle this case separately for this function to work properly.

src/Chapter2.hs
```diff
@@ -858,7 +876,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind(x:xs) = rewind xs ++ [x]
```
vrom911, line 881: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

PR #540, opened 2022-10-10
Completed Chapter 2

ngntrgduc:
### Solutions for Chapter 2

cc @vrom911 @chshersh

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -213,7 +213,7 @@ True :: Bool
 >>> :t 'a'
 'a' :: Char
 >>> :t 42
-42 :: Num a => a
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num a => a
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -653,7 +667,12 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n 
+  | n < 0 = firstDigit (abs n)
+  | n < 10 = n
+  | n >= 10 = firstDigit m
+  where m = div n 10
```
vrom911, line 675: 😍 Amazing!

src/Chapter2.hs
```diff
@@ -336,7 +348,12 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList a b x
+  | a < 0 || b < 0 = []
+  | b < a = []
+  | a == b = take 1 (drop a x)
+  | a < b = take (b - a + 1) (drop a x)
```
vrom911, line 355: This 2 clauses look similar 🕵🏼 
In case of `a == b` the expression `b - a + 1` would be equal to `1` which is exactly like it is done in the previous clause, so you can unite two of this into just one 🙂

src/Chapter2.hs
```diff
@@ -348,8 +365,8 @@ Implement a function that returns only the first half of a given list.
 >>> firstHalf "bca"
 "b"
 -}
--- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+firstHalf :: [a] -> [a]
+firstHalf l = take ((length l) `div` 2) l
```
vrom911, line 369: Great! 💯 
Note that you can omit the brackets here as you are using the infix form of the `div` function:

```suggestion
firstHalf l = take (length l `div` 2) l
```

src/Chapter2.hs
```diff
@@ -741,8 +765,8 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains a [] = []
+contains a (x:y) = if a `elem` x then x : (contains a y) else (contains a y)
```
vrom911, line 769: That's right!
You also can use the `filter` function  😉

src/Chapter2.hs
```diff
@@ -781,13 +805,13 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
-divideTenBy x = div 10 x
+divideTenBy = div 10 
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul xs = zipWith (*) xs
```
vrom911, line 814: You can go even further and eta-reduce `xs` too! 🤩 

```suggestion
pairMul = zipWith (*)
```

src/Chapter2.hs
```diff
@@ -842,7 +866,11 @@ list.
 
 🕯 HINT: Use the 'cycle' function
 -}
-rotate = error "rotate: Not implemented!"
+rotate a l
+  | a < 0 = []
+  | a == 0 = l
+  | l == [] = []
+  | otherwise = take (length l) (drop a (cycle l))
```
vrom911, line 873: Amazing 👍🏼 

If you want to be sneaky, you can also drop not just `a` but `a` mod by the length of the list so it is a bit more efficient in case when `a` is bigger than the length👌🏼

src/Chapter2.hs
```diff
@@ -858,7 +886,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [a] -> [a]
+rewind [] = []
+rewind x = drop ((length x) - 1) x ++  rewind (init x)
```
vrom911, line 891: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Yes, fantastic chapter! 🚆 👏🏼

ngntrgduc:
Thanks for your feedback ❤️. I'm ready for Chapter 3 :">

PR #542, opened 2022-10-14
Completed Chapter 1

yebowhatsay:
### Solutions for Chapter 1

cc @vrom911 @chshersh

Thanks for the amazing tutorial!

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num p => p
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -522,7 +522,10 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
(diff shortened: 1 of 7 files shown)


src/Chapter1.hs
```diff
@@ -490,7 +491,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Integral a => a -> a
+lastDigit n = (mod n 10)
```

src/Chapter1.hs
```diff
@@ -520,7 +522,10 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+--closestToZero x y = min (abs(x)) (abs(y))
+closestToZero x y = if (abs(x)) < (abs(y))
```
vrom911, line 526: `()` are redundant here. I see how it can be confusing to see when to use them and when not to use, especially in the beginning. But it's usually better to remove redundant brackets for cleaner code :broom: 
```haskell
```

src/Chapter1.hs
```diff
@@ -520,7 +522,10 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+--closestToZero x y = min (abs(x)) (abs(y))
```
vrom911, line 525: Almost does a job, but would return the absolute value instead of the initial value of `x` or `y` 🙂

src/Chapter1.hs
```diff
@@ -632,7 +649,12 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = let a = mod n 100
+                 b = mod a 10
+                 c = div a 10
```
vrom911, line 655: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod m 10
div m 10
```

```haskell
(x, y) = divMod m 10
```

src/Chapter1.hs
```diff
@@ -653,7 +675,11 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n 
+            | x < 10 = x
```
vrom911, line 680: I see, that for a negative number like -19, this implementation will return the negative number itself instead of the last digit because the first check n < 10 will always succeed for negative numbers 👌🏼

vrom911 (approved):
Hello there! Wonderful work on the Chapter!! 👏🏼

src/Chapter1.hs
yebowhatsay: Thanks @vrom911 . I think I fixed lastDigit. My solution is to use abs on all the inputs now. Hope this is the idiomatic way to do this.

src/Chapter1.hs
yebowhatsay: Thanks. A bit of a hangover coming from other languages. Fixed.

src/Chapter1.hs
yebowhatsay: Good point. Hadn't thought of that! My bad.

src/Chapter1.hs
yebowhatsay: fixed.

src/Chapter1.hs
yebowhatsay: nice function. added another implementation using divMod.

vrom911:
Looks great now, @yebowhatsay 🎉

yebowhatsay:
Thanks for your help, ***@***.****

yebowhatsay:
Completed Chapter 2.

PR #547, opened 2022-10-17
Solutions for Chapter 1

guntbert:
### Solutions for Chapter 1

cc @vrom911 @chshersh

This was real fun - thx a lot :bowing_man:

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -494,7 +494,7 @@ Implement a function that returns the last digit of a given number.
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
 lastDigit :: Int -> Int
-lastDigit n = mod n 10
+lastDigit n = mod (abs n) 10
 
 
 {- |
@@ -524,7 +524,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = min (abs x) (abs y)
+closestToZero x y = if (abs x) <= (abs y) then x else y
 
 
 {- |
@@ -640,8 +640,8 @@ specifying complex expressions.
 -}
 sumLast2 :: Int -> Int
 sumLast2 n = 
-    let digit1 = mod n 10
-        digit2 = mod (div n 10) 10
+    let digit1 = mod (abs n) 10
+        digit2 = mod (div (abs n) 10) 10
     in digit1 + digit2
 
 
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -568,7 +572,10 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel c
+    | elem c ['a', 'e', 'i', 'o', 'u'] = True
+    | otherwise = False
```
vrom911, line 578: Nice one! 👍🏼 

```haskell
isVowel c = elem c ['a', 'e', 'i', 'o', 'u']
```

src/Chapter1.hs
```diff
@@ -631,8 +638,11 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = 
+    let digit1 = mod (abs n) 10
+        digit2 = mod (div (abs n) 10) 10
```

```haskell
mod m 10
div m 10
```

```haskell
(x, y) = divMod m 10
```

src/Chapter1.hs
```diff
@@ -653,7 +663,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n
+    | abs n < 10 = n
```
vrom911, line 668: In case of `n` being negative, this would return negative result. as you are using `n` here 🙂 
But you are on the right way! 💪🏼

vrom911 (approved):
Amazing job 👏🏼

src/Chapter1.hs
guntbert: Ouch! I ran `make test-chapter1-basic` only, and hence missed the failure - should have caught this :-/

guntbert:
Thank you very much for the elaborate feedback - you are putting a lot of thoughts and work into this :bow: :pray:

src/Chapter1.hs
guntbert: Hmm...
I fail to see how this would be simpler - I need to apply the modulo-division by 10 to the result of the first division whereas `divMod` only applies it to the original value albeit as `div` and as `mod` at the same time.

PR #551, opened 2022-10-25
Completed Chapter1

rskbansal:
### Solutions for Chapter 1

cc @vrom911

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num a => a
 
 A pair of boolean and char:
 >>> :t (True, 'x')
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -632,7 +641,13 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = numLast + num2ndLast
+    where
+        numLast = mod m 10
+        num2ndLast = mod (div m 10) 10
```

```haskell
mod m 10
div m 10
```

```haskell
(x, y) = divMod m 10
```

src/Chapter1.hs
```diff
@@ -653,7 +668,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n = if div m 10 == 0 then m else firstDigit(div m 10)
```
vrom911, line 672: Great Job!
In Haskell all the functions and operators are separated from its arguments with spaces 👾 
So here it would look like this:

```suggestion
firstDigit n = if div m 10 == 0 then m else firstDigit (div m 10)
```

vrom911 (approved):
Amazing!1 Congrats with finishing Chapter 1 👏🏼

rskbansal:
> Amazing!1 Congrats with finishing Chapter 1 👏🏼

Thanks a lot!

PR #554, opened 2022-11-01
Pullrequest from Brazil - Learn Haskell

adrwtr:
### Solutions for Chapter 1

What a wonderful tutorial! Thanks for the service.
Could you rate my code?

cc @vrom911

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num a => a
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+max :: Ord a => a -> a -> a
 
 You might not understand each type at this moment, but don't worry! You've only
 started your Haskell journey. Types will become your friends soon.
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -301,43 +301,43 @@ expressions in GHCi
   functions and operators first. Remember this from the previous task? ;)
 
 >>> 1 + 2
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+3
 
 >>> 10 - 15
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+-5
 
 >>> 10 - (-5)  -- negative constants require ()
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+15
 
 >>> (3 + 5) < 10
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> True && False
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+False
 
 >>> 10 < 20 || 20 < 5
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> 2 ^ 10  -- power
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+1024
 
 >>> not False
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True
 
 >>> div 20 3  -- integral division
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+6
 
 >>> mod 20 3  -- integral division remainder
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+2
 
 >>> max 4 10
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+10
 
 >>> min 5 (max 1 2)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+2
 
 >>> max (min 1 10) (min 5 7)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+5
 
 Because Haskell is a __statically-typed__ language, you see an error each time
 you try to mix values of different types in situations where you are not
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -646,7 +646,14 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = mod10 (div10 n) + mod10 n
+    where
+        div10 :: Int -> Int
+        div10 x = div x 10
+
+        mod10 :: Int -> Int
+        mod10 x = mod x 10
 
 
 {- |
@@ -667,7 +674,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n = if n < 10
+    then n
+    else firstDigit (n `div` 10)
 
 
 {-
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -428,7 +428,7 @@ task is to specify the type of this function.
 >>> squareSum 3 4
 49
 -}
-squareSum :: Int a => a -> a -> a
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -553,8 +554,14 @@ value after "=" where the condition is true.
...
-
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+    | x < y && x < z && z < y = z
+    | x < y && x < z && y < z = y
+    | z < x && z < y && x < y = x
+    | z < x && z < y && y < x = y
+    | y < x && y < z && x < z = x
+    | otherwise = z
```
vrom911, line 564: Awesomely done!

One note: you can remove the first case here as it can be covered by the `otherwise` clause in this case:

```suggestion
    | x < y && x < z && y < z = y
    | z < x && z < y && x < y = x
    | z < x && z < y && y < x = y
    | y < x && y < z && x < z = x
    | otherwise = z
```

src/Chapter1.hs
```diff
@@ -632,7 +646,14 @@ Try to introduce variables in this task (either with let-in or where) to avoid
...
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 n = mod10 (div10 n) + mod10 n
+    where
+        div10 :: Int -> Int
+        div10 x = div x 10
+
+        mod10 :: Int -> Int
+        mod10 x = mod x 10
```
vrom911, line 656: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

```haskell
mod m 10
div m 10
```

```haskell
(x, y) = divMod m 10
```

src/Chapter1.hs
```diff
@@ -653,7 +674,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
+firstDigit :: Int -> Int
+firstDigit n = if n < 10
```
vrom911, line 678: I see, that for a negative number like -19, this implementation will return the negative number itself instead of the last digit because the first check n < 10 will always succeed for negative numbers.

src/Chapter2.hs
```diff
@@ -336,7 +348,9 @@ from it!
 ghci> :l src/Chapter2.hs
 -}
 subList :: Int -> Int -> [a] -> [a]
-subList = error "subList: Not implemented!"
+subList _ _ [] = []
+subList x 0 y = []
+subList x y b = take (y-x+1) $ drop x b
```
vrom911, line 353: Great solution!
Note that in Haskell all operators/functions are separated from its arguments with spaces 👾 

```suggestion
subList x y b = take (y - x + 1) $ drop x b
```

src/Chapter2.hs
```diff
@@ -349,7 +363,10 @@ Implement a function that returns only the first half of a given list.
 "b"
 -}
 -- PUT THE FUNCTION TYPE IN HERE
-firstHalf l = error "firstHalf: Not implemented!"
+
+firstHalf :: [a] -> [a]
+firstHalf [] = []
```
vrom911, line 368: Actually, this case is included in the next line solution too, so can be safely removed ✨

src/Chapter2.hs
```diff
@@ -501,7 +518,9 @@ True
 >>> isThird42 [42, 42, 0, 42]
 False
 -}
-isThird42 = error "isThird42: Not implemented!"
+isThird42 :: (Eq a, Num a) => [a] -> Bool
+isThird42 [] = False
+isThird42 ( _ : _ : x : xs ) = if x == 42 then True else False
```
vrom911, line 523: You don't need to compare `x` with 42 separately, you can pattern match on `42` directly in the pattern.

```suggestion
isThird42 ( _ : _ : 42 : _ ) = True
isThird42 _ = False
```

src/Chapter2.hs
```diff
@@ -621,7 +641,15 @@ Write a function that takes elements of a list only in even positions.
...
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven [] =[]
+takeEven (x:xs) = recursiveEven 0 (x : xs)
+    where
+        recursiveEven :: Int -> [a] -> [a]
+        recursiveEven _ [] = []
+        recursiveEven c (x:xs) = if mod c 2 == 0
+            then x : recursiveEven (c+1) xs
+            else recursiveEven (c+1) xs
```
vrom911, line 652: Nice idea on the helper functions :+1:

src/Chapter2.hs
```diff
@@ -727,8 +755,13 @@ value of the element itself
 
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
+-- smartReplicate :: [Int] -> [Int]
+-- smartReplicate [] = []
+-- smartReplicate (x:xs) = (replicate x x) ++ smartReplicate xs
+-- dont understod how to use map smartReplicate arr = map (\x -> replicate x x)  arr
```
vrom911, line 761: I see that on the final solution below, you eventually used `concatMap` function that is basically both maps and concatenates (flattens) the resulting list. The problem with just `map` function is that it will return you the list of the lists, but what you need here is the flattened list, therefore usage of the `concat` function is needed 🙂 👍🏼

src/Chapter2.hs
```diff
@@ -741,8 +774,12 @@ the list with only those lists that contain a passed element.
 
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
-
+contains :: Int -> [[Int]] -> [Int]
```
vrom911, line 777: Here, in the result, it is expected to have all of the list that contain the given element, so the type of the function will look like:

```suggestion
contains :: Int -> [[Int]] -> [[Int]]
```

src/Chapter2.hs
```diff
@@ -781,13 +818,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: [Int] -> [Int] -> [Int]
+pairMul = zipWith (*)
```
vrom911, line 829: Eta-reduction award :trophy:

src/Chapter2.hs
```diff
@@ -858,7 +899,9 @@ and reverses it.
   function, but in this task, you need to implement it manually. No
   cheating!
 -}
-rewind = error "rewind: Not Implemented!"
+rewind :: [Int] -> [Int]
+rewind [] = []
+rewind (x:xs) = (rewind xs) ++ [x]
```
vrom911, line 904: Your solution is correct! However, it is slow. In lists, it is quite slow to add anything at the end of the list. That is why it is always better to rewrite it with the `:` cons. Remember the explanation with trains? 🚂 🚋 🚋 

vrom911 (approved):
Amazing work on the course 💪🏼 👏🏼

src/Chapter1.hs
```diff
@@ -490,7 +490,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit n = mod n 10
```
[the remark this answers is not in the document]
adrwtr: ok, i did the following test

ghci> lastDigit (-222)
8

I understood the problem and fixed it with abs - thanks

src/Chapter1.hs
adrwtr: ow --- that´s true - thanks

src/Chapter1.hs
adrwtr: That´s true - i put abs on test

src/Chapter2.hs
adrwtr: Hmmmm.. hahahah understood

src/Chapter2.hs
adrwtr: >> Alternatively, you could use pattern matching in takeEven instead. 
>> As you need to take every second element, it is possible to represent with a few pattern matching clauses


I tried to use the pattern but I didn't understand

takeEven :: [a] -> [a]
takeEven [] = []
takeEven (x:y:xs) = [y] ++ takeEven xs

ghci> takeEven [2, 1, 3, 5, 6]
[1,5*** Exception: src\Chapter2.hs:(654,1)-(655,38): Non-exhaustive patterns in function takeEven

adrwtr: I got it!! A clean and readeble solution. Thx for support

src/Chapter2.hs
adrwtr: I understood. Just wanted to make sure it was right to use ConcatMap. Thx

src/Chapter2.hs
adrwtr: wow.. I didn't realize when I built the function. I managed to fix

src/Chapter2.hs
adrwtr: I'm going to recap this chapter to understand better.

src/Chapter1.hs
adrwtr: I broke my head on this one. In the end I got it. I had problems with a partially applied function.

src/Chapter2.hs
adrwtr: Ok, i recap a understand how to use de rotate to do this! Thx

adrwtr:
Hi @vrom911 it's been a long journey to learn haskell with your help.

Chapters 3 and 4 were pretty tough for me.

In chapter 4 I got stuck and had to do research in order to evolve. Despite this, I think that I can evolve a lot in language difficulties. I practically managed to make the "final boss" by myself which shows that your method works.

Thank you very much and if you can rate chapter 3 and 4 I would appreciate it.

PR #556, opened 2023-02-24
Solutions for Chapter 1,2 & 3

WolfSoko:
### Solutions for Chapter 1 and 2

cc @vrom911

### DIFF
--- a/src/Chapter3.hs
+++ b/src/Chapter3.hs
@@ -1149,6 +1149,25 @@ implement the following functions:
 🕯 HINT: to implement this task, derive some standard typeclasses
 -}
 
+data Weekday
+  = Monday
+  | Tuesday
+  | Wednesday
+  | Thursday
+  | Friday
+  | Saturday
+  | Sunday
+  deriving (Show, Read, Eq, Ord, Enum, Bounded)
+
+isWeekend ::  Weekday -> Bool
+isWeekend day = day == Saturday || day == Sunday
+
+nextDay :: Weekday -> Weekday
+nextDay day = if day == maxBound then minBound else succ day
+
+daysToParty :: Weekday -> Int
+daysToParty day = fromEnum Friday - fromEnum day
+
 {-
 =💣= Task 9*
 
@@ -1195,26 +1214,30 @@ Deriving: https://kowainik.github.io/posts/deriving
 Extensions: https://kowainik.github.io/posts/extensions
 -}
 
+printTaskHeader :: String -> IO ()
+printTaskHeader taskName = do
+  putStrLn ""
+  putStrLn "---------------------------"
+  putStrLn $ "---------- Task-" ++ taskName ++ " ---------"
+  putStrLn "---------------------------"
+
 main :: IO ()
 main = do
-  print "---------------------------"
-  print "Task 5"
+  printTaskHeader "5"
   print auenland
   print grownAuenland
   print auenlandWithCastle
   print auenlandWithCastleAndWalls
   print auenlandCannotBuildWalls
   print auenlandCannotBuildWallsEvenWithCastle
-  print "---------------------------"
-  print "Task 6"
+  printTaskHeader "6"
   print $
     Liar
       { dragon = Dragon "Fire Dragon Eremil" $ Just "Fire Spell",
         treasure = Just $ TreasureChest {treasureChestGold = 100, treasureChestLoot = "Dragon Plate"}
       }
 
-  print "---------------------------"
-  print "Task 7"
+  printTaskHeader "7"
 
   print $ append Gold {goldAmount = 10} Gold {goldAmount = 20}
   print $ append [1, 2] [3, 4, 5]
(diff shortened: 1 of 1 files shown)


src/Chapter1.hs
```diff
@@ -520,7 +519,7 @@ branches because it is an expression and it must always return some value.
   satisfying the check will be returned and, therefore, evaluated.
 -}
 closestToZero :: Int -> Int -> Int
-closestToZero x y = error "closestToZero: not implemented!"
+closestToZero x y = if (abs x) < (abs y) then x else y
```
vrom911, line 522: `()` are redundant here. I see how it can be confusing to see when to use them and when not to use, especially in the beginning. But it's usually better to remove redundant brackets for cleaner code :broom: 
```suggestion
closestToZero x y = if abs x < abs y then x else y

File: src/Chapter1.hs [hunk 2]
```diff
@@ -553,8 +552,11 @@ value after "=" where the condition is true.
 
 Casual reminder about adding top-level type signatures for all functions :)
 -}
-
-mid x y z = error "mid: not implemented!"
+mid :: Int -> Int -> Int -> Int
+mid x y z
+  | x <= y && y <= z = y
+  | x >= y && y >= z = y
+  | otherwise = mid y z x
```
vrom911, line 559: That's clever solution to mix them around 💡

src/Chapter1.hs
```diff
@@ -631,9 +645,12 @@ Implement a function that returns the sum of the last two digits of a number.
 Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
-
-sumLast2 n = error "sumLast2: Not implemented!"
-
+sumLast2 :: Int -> Int
+sumLast2 n = (lastDigit n) + secondLastDigit
```
vrom911, line 649: Nice reusage of previous `lastDigit` function! 👏🏼 

I see that you have already a var to represent this bit – `lastD`, you can reuse it here too!

```suggestion
sumLast2 n = lastD + secondLastDigit
```

src/Chapter1.hs
```diff
@@ -652,9 +669,12 @@ Implement a function that returns the first digit of a given number.
...
 -}
-
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit n 
+  | absN < 10  = absN
+  | otherwise = firstDigit (div absN 10)
+  where
+    absN = abs n
```
vrom911, line 677: Very elegant to use `where` clause here 👍🏼

src/Chapter2.hs
```diff
@@ -621,7 +627,13 @@ Write a function that takes elements of a list only in even positions.
...
 [2,3,4]
 -}
-takeEven = error "takeEven: Not implemented!"
+takeEven :: [a] -> [a]
+takeEven l = go True l
+  where
+    go :: Bool -> [a] -> [a]
+    go _ [] = []
+    go True (x:xs) = x : go False xs
+    go False (_:xs) = go True xs
```
vrom911, line 636: Nice idea on the helper functions :+1:

src/Chapter2.hs
```diff
@@ -728,7 +740,7 @@ value of the element itself
 🕯 HINT: Use combination of 'map' and 'replicate'
 -}
 smartReplicate :: [Int] -> [Int]
-smartReplicate l = error "smartReplicate: Not implemented!"
+smartReplicate l = concat (map (\x -> replicate x x) l)
```
vrom911, line 743: Great! 👍🏼 

src/Chapter2.hs
```diff
@@ -741,7 +753,13 @@ the list with only those lists that contain a passed element.
...
 🕯 HINT: Use the 'elem' function to check whether an element belongs to a list
 -}
-contains = error "contains: Not implemented!"
+contains :: Int -> [[Int]] -> [[Int]]
+contains n l = go n l where
+  go :: Int -> [[Int]] -> [[Int]]
+  go _ [] = []
+  go n (x:xs) = if n `elem` x 
+    then x : go n xs 
+    else go n xs
```
vrom911, line 762: Nice `go`! It is actually the standard `filter` function, that you can use instead of the custom `go`, but it is correct!

src/Chapter2.hs
```diff
@@ -781,13 +799,15 @@ Let's now try to eta-reduce some of the functions and ensure that we
...
 
 -- TODO: type ;)
-listElementsLessThan x l = filter (< x) l
+listElementsLessThan :: Int -> [Int] -> [Int]
+listElementsLessThan x = filter (< x)
 
 -- Can you eta-reduce this one???
-pairMul xs ys = zipWith (*) xs ys
+pairMul :: Num a => [a] -> [a] -> [a]
+pairMul = zipWith (*)
```
vrom911, line 810: Eta-reduction award :trophy:

vrom911 (approved):
Congrats with finishing these chapters 🎉

PR #572, opened 2024-01-25
🎉 chapter1

CypressVillage:
### Solutions for Chapter 1

cc @vrom911

### DIFF
--- a/src/Chapter1.hs
+++ b/src/Chapter1.hs
@@ -209,31 +209,31 @@ So, the output in this example means that 'False' has type 'Bool'.
 > Try to guess first and then compare your expectations with GHCi output
 
 >>> :t True
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+True :: Bool
 >>> :t 'a'
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+'a' :: Char
 >>> :t 42
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+42 :: Num a => a
 
 A pair of boolean and char:
 >>> :t (True, 'x')
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(True, 'x') :: (Bool, Char)
 
 Boolean negation:
 >>> :t not
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+not :: Bool -> Bool
 
 Boolean 'and' operator:
 >>> :t (&&)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(&&) :: Bool -> Bool -> Bool
 
 Addition of two numbers:
 >>> :t (+)
-<INSERT THE RESULT INSTEAD OF THE TEXT>
+(+) :: Num a => a -> a -> a
 
 Maximum of two values:
 >>> :t max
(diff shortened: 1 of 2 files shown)


src/Chapter1.hs
```diff
@@ -568,7 +574,8 @@ True
 >>> isVowel 'x'
 False
 -}
-isVowel c = error "isVowel: not implemented!"
+isVowel :: Char -> Bool
+isVowel = (`elem` "aoeiuv")
```
vrom911, line 578: Nice one! 👍🏼 

One note in here, that sometimes, `elem` could be slower than the explicit pattern matching. I remember there were some benchmarks on one particular case, that showed how moving to pattern matching on each case separately drastically decrease time 🐎 But for this solution it is totally fine!

src/Chapter1.hs
```diff
@@ -632,7 +639,12 @@ Try to introduce variables in this task (either with let-in or where) to avoid
 specifying complex expressions.
 -}
 
-sumLast2 n = error "sumLast2: Not implemented!"
+sumLast2 :: Int -> Int
+sumLast2 x =
+  let ax = abs x 
+      last = ax `mod` 10
+      secondlast = ax `mod` 100 `div` 10
```
vrom911, line 646: That is a wonderful solution! 👏🏼  You correctly noticed that it is the `div` and `mod`, cool 😎 

Another approach could be also using this: you can see that you use both:

```haskell
mod m 10
div m 10
```

```haskell
(x, y) = divMod m 10
```

src/Chapter1.hs
```diff
@@ -653,8 +665,10 @@ You need to use recursion in this task. Feel free to return to it later, if you
 aren't ready for this boss yet!
 -}
 
-firstDigit n = error "firstDigit: Not implemented!"
-
+firstDigit :: Int -> Int
+firstDigit x
+  | x `div` 10 == 0 = x
+  | otherwise       = firstDigit (x `div` 10)
```
vrom911, line 671: I see, that for a negative numbers it will be not return the correct answer. Sometimes it is tricky to keep in mind all the edge cases, so no worries, but the idea and recursive solution is awesome 👏🏼

src/Chapter1.hs
```diff
@@ -490,7 +491,8 @@ Implement a function that returns the last digit of a given number.
   whether it works for you!
 -}
 -- DON'T FORGET TO SPECIFY THE TYPE IN HERE
-lastDigit n = error "lastDigit: Not implemented!"
+lastDigit :: Int -> Int
+lastDigit = (`mod` 10) . abs
```
vrom911, line 495: That's pretty!

Nice work on using infix form of calling functions! We don't explain this syntax in our course to make it short. But well done :+1:

vrom911 (approved):
Bravo! Chapter number One is completed, congrats 🎉

vrom911 (approved):

CypressVillage:
Thank you for such a careful and meticulous reply!😀 I have corrected the relevant sections

vrom911:
No worries! It is a very great job! Feel free to get to the next Chapter ❤️

